PR #49, opened 2022-10-10
Support only Python 3.8, 3.9 and 3.10

callumforrester:
Closes #48

PR #51, opened 2022-10-11
Adopt new skeleton

callumforrester:
Adopt new skeleton, no longer using pipenv: https://github.com/DiamondLightSource/python3-pip-skeleton


callumforrester:
Closes #49 by rendering it obsolete

PR #61, opened 2022-11-16
Explicitly tag empty-bodied functions that are supposed to be empty-bodied

callumforrester:
Mypy 0.991 disallows empty function bodies by default, see https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html

While I have mixed feelings about this change I do find empty bodies useful, for example for dummy functions in tests:
```python
def my_test_function() -> int:
    ...

def test_something():
    assert something_to_do_with(my_test_function.__annotations__)
```
So I have added config to disable it. Empty functions are not widely spread and I could probably also remove them all.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,6 @@ console_scripts =
 [mypy]
 # Ignore missing stubs for modules we use
 ignore_missing_imports = True
-allow_empty_bodies = True
 
 [isort]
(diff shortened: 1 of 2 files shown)


callumforrester:
Decision is to keep the empty functions and tag them with `#type: ignore` to be explicit

PR #85, opened 2023-01-27
Revert "Prefix event topics with public (#83)"

callumforrester:
This reverts commit 3ec41a0efdad1a0435eabfb55f5e6ff2d4392905.


callumforrester:
Fixed by #84

PR #90, opened 2023-02-08
Switch from using apischema to pydantic

callumforrester:
* Replace all public dataclasses with `BaseModel`s.
* Add "type validators" to replicate device lookup behaviour. They do not work as well as apischema converters and various workarounds were needed. Will make issues off of this PR.
* Add tests


src/blueapi/plans/plans.py
```diff
@@ -15,17 +12,20 @@
 
 def scan(
     detectors: List[Readable],
-    spec: Spec[Movable],
+    axes_to_move: Mapping[str, Movable],
```
callumforrester, line 15: Workaround documented in https://github.com/DiamondLightSource/blueapi/issues/118

PR #107, opened 2023-03-31
Add support for AMQP over RabbitMQ broker

### LINKED ISSUE #104: AMQP Compatibility

DiamondJoseph:
Closes #104 with support from command line and in helm

- [ ] Test CLi
- [ ] Test on Helm

### DIFF
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -6,10 +6,10 @@ from typing import Optional
 import click
 
 from blueapi import __version__
-from blueapi.config import StompConfig
-from blueapi.messaging import StompMessagingTemplate
+from blueapi.config import StompConfig, AMQPConfig
+from blueapi.messaging import StompMessagingTemplate, AMQPMessagingTemplate
 
-from .amq import AmqClient
+from .amq import MessagingClient
 from .updates import CliEventRenderer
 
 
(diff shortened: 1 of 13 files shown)


src/blueapi/config.py
```diff
@@ -13,6 +13,26 @@ class StompConfig:
...
+    host: str = "localhost"
+    port: int = 5672
+    userid: str = "guest"  # TODO: Config reads from env, so can be injected as Secret?
+    password: str = "guest"  # Can only use guest/guest on loopback address
+    virtual_host: str = "/"
+
+
+@dataclass
+class MessagingConfig:
+    impl: str = "stomp"
```
callumforrester, line 31: To make it a bit more spring-esque, I suggest making implementations independent of eachother. For example:
```suggestion
    using: List[str] = ["stomp"]
```

and then when instantiating:
```python
for impl in config.using:
    if impl == "stomp":
        ...
    elif impl == "amqp":
        ...
```

src/blueapi/config.py
```diff
@@ -13,6 +13,26 @@ class StompConfig:
...
+@dataclass  # TODO: Defer immediately to pika.ConnectionProperties?
+class AMQPConfig:
+    """
+    Config for connecting to the AMQP broker
+    """
+
+    host: str = "localhost"
+    port: int = 5672
+    userid: str = "guest"  # TODO: Config reads from env, so can be injected as Secret?
+    password: str = "guest"  # Can only use guest/guest on loopback address
```
callumforrester, line 25: Do we need the capability to inject this via some sort of secret?

src/blueapi/cli/cli.py
```diff
@@ -53,23 +53,76 @@ def start_worker(config: Optional[Path]):
...
+    if impl == "amqp":
+        if prompt:
+            password = getpass.getpass(prompt=f"AMQP Password for user {user}:")
+        client = MessagingClient(
+            AMQPMessagingTemplate.autoconfigured(
+                AMQPConfig(host, port, userid, password, virtual_host)
+            )
+        )
+    else:
+        client = StompMessagingTemplate.autoconfigured(StompConfig(host, port))
```
callumforrester, line 117: In the first case, `client` is a `MessagingClient` wrapping a `MessagingTemplate` instance. In the second case it is just a `MessagingTemplate` instance on its own. Is that right?

callumforrester (changes requested):
Looking good! Left a few comments, also to do:
* Linting needs fixing (`tox -e pre-commit,mypy` to run locally) 
* Tests are timing out
* Need rabbit as a service in CI

DiamondJoseph:
Superseded by #135

PR #131, opened 2023-04-20
Allow plans with Concrete/non-Protocol parameters 

### LINKED ISSUE #117: Pass concrete device types as plan parameters
Currently blueapi will resolve

```python
def my_plan(det: Readable)
```
but not...

```python
def my_plan(det: XPress3)
```

We were hoping to get away with this, but it's becoming clear we will need to support concrete device types as well so that developers can access the component tree. `XPress3` will have fields like `cam.acquire_time`, that `Readable` doesn't have.

DiamondJoseph:
Closes #117


callumforrester (approved):
Looks good to me, one minor nitpick and you can probably remove the TODOs, as explained in comments.

src/blueapi/core/context.py
```diff
@@ -144,20 +146,41 @@ def device(self, device: Device, name: Optional[str] = None) -> None:
...
+    def get_validators(
+        plan_arg: str, field_type: Type
+    ) -> Iterable[TypeValidatorDefinition]:
+        if isinstance(field_type, _GenericAlias):
+            for arg in get_args(field_type):
+                yield from get_validators(plan_arg, arg)
+        elif is_bluesky_compatible_device_type(field_type):
+            yield get_validator(plan_arg, field_type)
+
+    for k, v in Signature.from_callable(plan).parameters.items():
```
callumforrester, line 185: This will fail if a plan has two arguments of the same type because `utils.type_validator` doesn't allow it. Example test case:
```python
def concrete_plan_multi_arg(
    motor_1: DelegatingNamedMovable,
    motor_2: DelegatingNamedMovable,
) -> MsgGenerator:  # type: ignore
    ...
```
Presumably some sort of caching is needed

src/blueapi/core/context.py
```diff
@@ -144,20 +146,41 @@ def device(self, device: Device, name: Optional[str] = None) -> None:
...
+        return TypeValidatorDefinition(field_type, get_device)
+
+    def get_validators(
+        plan_arg: str, field_type: Type
+    ) -> Iterable[TypeValidatorDefinition]:
+        if isinstance(field_type, _GenericAlias):
+            for arg in get_args(field_type):
+                yield from get_validators(plan_arg, arg)
+        elif is_bluesky_compatible_device_type(field_type):
+            yield get_validator(plan_arg, field_type)
```
callumforrester, line 183: It feels like there should be a way to write a generic version of this in `utils.type_validator`, but I think it's fine to consider that out of scope. Mind making a ticket for it?

src/blueapi/core/context.py
```diff
@@ -144,20 +146,41 @@ def device(self, device: Device, name: Optional[str] = None) -> None:
...
-def device_validators(ctx: BlueskyContext) -> Iterable[TypeValidatorDefinition]:
-    def get_device(name: str) -> Device:
-        device = ctx.find_device(name)
-        if device is None:
-            raise KeyError(f"Could not find a device named {name}")
-        return device
-
-    for proto in BLUESKY_PROTOCOLS:
-        yield TypeValidatorDefinition(proto, get_device)
+def plan_validators(
```
callumforrester, line 157: I think this function may grow and change and possibly be split up and recombined in all sorts of ways as we add more fancy plan parameter processing. Could we rename it to `_plan_validators` to discourage it's use outside this module for now?

callumforrester (changes requested):

PR #132, opened 2023-04-20
Make all API models convert to camelCase

callumforrester:
Previously, with apischema, the API models were serialized with camel case aliasing. For example, converting a Python field called `foo_bar` to a JSON field called `fooBar` and vice versa. This is to comply with the Google JSON style guide. 

https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format 

This PR adds a custom base model with custom config primarily to preserve this change and also to prevent the ingestion of arbitrary JSON alongside a model's known fields, which apischema also did not allow.

### DIFF
--- a/src/blueapi/utils/base_model.py
+++ b/src/blueapi/utils/base_model.py
@@ -1,0 +1,35 @@
+from pydantic import BaseConfig, BaseModel, Extra
+
+
+def _to_camel(string: str) -> str:
+    words = string.split("_")
+    return words[0] + "".join(word.capitalize() for word in words[1:])
+
+
+class BlueapiModelConfig(BaseConfig):
+    """
+    Pydantic config for blueapi API models with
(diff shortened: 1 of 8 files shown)


src/blueapi/utils/base_model.py
```diff
@@ -0,0 +1,35 @@
+from pydantic import BaseConfig, BaseModel, Extra
+
+
+def _to_camel(string: str) -> str:
```
DiamondJoseph, line 4: Some words on why this change is required would be useful to prevent it being undone again at some latter point

callumforrester: Done in docstring of `BlueapiPlanModelConfig`

DiamondJoseph (approved):

Issue #120, opened 2023-04-14
Support commands via REST API


callumforrester:
There are two types of interaction with the worker
1. Command/response, e.g. "tell me what plans you have?" "please run this plan"
2. Continuous updates e.g. data events, task events
We wanted to use one protocol for communication so we decided on message buses. Message buses aren't fantastic in terms of support for type 1. but the normal REST approach doesn't support type 2 at all. 

We felt that the complexity/inelegance of having both approaches outweighed the cost of dealing with command/response in message buses. From discussion, it seems that that is no longer the prevailing opinion. 
The design now:
![image](https://user-images.githubusercontent.com/29771545/232048553-5608f809-b30c-4df7-a7c9-391d2df54d80.png)

callumforrester:
@RAYemelyanova we should use this space for discussion of the API, I am vaguely thinking something along the lines of

GET `/plans`
GET `/plans/{name}`
GET `/devices`
GET `/devices/{name}`
GET `/task`
PUT `/task`

RAYemelyanova:
In terms of the put command, I think we should actually have two routes;

1. just puts the task onto the queue. Immediately returns.
2. blocks and waits for the task to finish. Returns with appropriate error if there is one.

I think I prefer 2 but 1 is probably still useful so, should we have both?

RAYemelyanova:
The other option is I run some validation on a task before I queue it - what do you think of this idea?

PR #135, opened 2023-04-21
Use FastAPI to provide rest endpoints

RAYemelyanova:
resolves https://github.com/DiamondLightSource/blueapi/issues/120

The only thing left to do is change the cli commands.


src/blueapi/rest/handler.py
```diff
@@ -0,0 +1,77 @@
...
+from blueapi.worker.worker import Worker
+
+settings = AppSettings()
+
+
+class Handler:
+    context: BlueskyContext
+    worker: Worker
+    config: ApplicationConfig
+    message_bus: MessagingTemplate
```
DiamondJoseph, line 20: Should we allow message_bus to be null?

src/blueapi/rest/handler.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def stop(self) -> None:
+        self.worker.stop()
+
+
+HANDLER: Optional[Handler] = None
+
+
+def get_handler() -> Handler:
+    """Retrieve the class which handles TODO change this objects."""
```
DiamondJoseph, line 60: Resolve this TODO?

src/blueapi/rest/main.py
```diff
@@ -0,0 +1,51 @@
...
+    try:
+        return PlanModel.from_plan(handler.context.plans[name])
+    except KeyError:
+        raise HTTPException(status_code=404, detail="Item not found")
+
+
+@app.get("/devices", response_model=DeviceResponse)
+async def get_devices(handler: Handler = Depends(get_handler)):
+    return DeviceResponse(
+        devices=[DeviceModel.from_device(i) for i in handler.context.devices.values()]
```
DiamondJoseph, line 29: nit: i -> device to match the handling for get_plans

src/blueapi/rest/main.py
```diff
@@ -0,0 +1,51 @@
...
+        return DeviceModel.from_device(handler.context.devices[name])
+    except KeyError:
+        raise HTTPException(status_code=404, detail="Item not found")
+
+
+@app.put("/task/{name}")
+async def execute_task(
+    name: str,
+    task: RunPlan = Body(
+        ..., example={"name": "count", "params": {"detectors": ["x"]}}
```
DiamondJoseph, line 45: is example a fastapi thing?

src/blueapi/worker/reworker.py
```diff
@@ -64,20 +67,33 @@ def __init__(
...
-    def run_forever(self) -> None:
+    def start(self) -> None:
+        LOGGER.info("Starting worker in own thread")
+        run_worker_in_own_thread(self)
+
+    def stop(self) -> None:
+        LOGGER.info("Stopping worker thread")
+        self._request_exit.set()
+        self._finished.wait()
+        LOGGER.info("Worker finished")
```
DiamondJoseph, line 86: Should clear the status of these events after stopping, in case we later call start() again (or implement a restart)

DiamondJoseph (changes requested):

src/blueapi/cli/cli.py
```diff
@@ -22,29 +19,22 @@ def main(ctx) -> None:
         print("Please invoke subcommand!")
 
 
-@main.command(name="worker")
-@click.option("-c", "--config", type=Path, help="Path to configuration YAML file")
```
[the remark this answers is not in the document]
RAYemelyanova: From a fast API perspective, passing things into the application via command line arguments is difficult to do.
It's easier to have some settings that are automagically detected based on environment variables, using pydantic's BaseSettings. I think if you set `APP_CONFIG_FILE` to something it'll work in the same way. How likely are people to want to change the configuration regularly enough where it's annoying to type `'export APP_CONFIG_FILE=...'`

src/blueapi/cli/amq.py
```diff
@@ -2,7 +2,7 @@
 from typing import Any, Callable, Mapping, Optional, TypeVar
 
 from blueapi.messaging import MessageContext, MessagingTemplate
-from blueapi.service.model import (
+from blueapi.rest.model import (
```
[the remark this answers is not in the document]
RAYemelyanova: Ok, so do you think it makes more sense to rename the 'rest' module to 'service'? While I'm at it I might rename `blueapi.rest.main` to `blueapi.service.rest` or something. I don't like the name main, just didn't know what else to call it...

src/blueapi/cli/cli.py
```diff
@@ -22,29 +19,22 @@ def main(ctx) -> None:
...
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        universal_newlines=True,
+        # capture_output=True,
+    )
 
 
 @main.group()
-@click.option(
```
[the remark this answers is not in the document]
RAYemelyanova: I suppose you could - the CLI commands on this file were all things that have now been replaced by the rest api. There wasn't anything like 'get event documents' or some other function like that.

This is why I've removed the config, as now all the functionality this file used to have (i.e. getting plans, devices and running tasks) is handled by the fastAPI server instead.

src/blueapi/cli/cli.py
```diff
@@ -54,53 +44,43 @@ def start_worker(config: Optional[Path]):
...
-@click.pass_context
-def get_plans(ctx) -> None:
-    client: AmqClient = ctx.obj["client"]
-    plans = client.get_plans()
-    print("PLANS")
-    for plan in plans.plans:
-        print("\t" + plan.name)
+@check_connection
+def get_plans() -> None:
+    resp = requests.get("http://localhost:8000/plans")
```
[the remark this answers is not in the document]
RAYemelyanova: Yep, good point; I'll add a setting in config.py that sets the url, so that it polls whichever one is set there.

src/blueapi/rest/handler.py
RAYemelyanova: This isn't a dataclass, so my understand is it's never really null - when it's initialised it's set to something. It needs to be set in __init__ because it needs configuration...

Thinking about it, now that I've made the config file path an environment variable that is ingested by `pydantic.BaseSettings`, it might be a good idea to simplify the `__init__` function here and remove the config_path entirely, instead taking it from the settings. But that's if you like what I've done with the settings - we could, potentially, revert back to config being passed via the CLI. I'd probably need to do some research into how to pass args to fastAPI but I think it should be possible.

src/blueapi/cli/cli.py
```diff
@@ -54,53 +44,43 @@ def start_worker(config: Optional[Path]):
...
+    print(f"Response Returned With {resp.status_code}: ")
+    pprint(resp.json())
```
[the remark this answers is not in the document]
RAYemelyanova: my understanding was that the cli is just for user interaction. Actual logs written to greylog for example will be produced by the `bluesky run` cli command, or running `uvicorn src.blueapi.rest.main:app --host $HOST --port $PORT`

src/blueapi/cli/cli.py
```diff
@@ -54,53 +44,43 @@ def start_worker(config: Optional[Path]):
...
-    ctx.obj["client"] = client
-    client.app.connect()
+
+
+def check_connection(func):
+    def wrapper(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except ConnectionError:
+            print("failed to establish connection. Are you sure the app is running?")
```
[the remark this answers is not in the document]
RAYemelyanova: I'll meet you in the middle and make it, 'Failed to establish connection'. I want it to be a little bit more descriptive. Also the only way this error happens is if someone tries to do,

`blueapi controller devices`

before starting the restapi. So, if this is going to be running in a docker container, it'll be running the fastAPI server. It won't be making CLI commands so there's no need to try to minimise how big the error message is.

src/blueapi/rest/handler.py
```diff
@@ -0,0 +1,77 @@
...
+    global HANDLER
+    config_path = (
+        Path(settings.app_config_path) if settings.app_config_path is not None else None
+    )
+    HANDLER = Handler(config_path)
+    HANDLER.start()
+
+
+def teardown_handler() -> None:
+    return HANDLER.stop() if HANDLER else None
```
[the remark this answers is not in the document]
RAYemelyanova: Not sure I understand you here; HANDLER.stop() is a blocking call that returns None. But it's possible someone tries to use this function before setting HANDLER globally in the first place.

src/blueapi/rest/main.py
RAYemelyanova: yep, it means in the docs it'll populate it with an example.

RAYemelyanova:
Just so you know I don't have write access so can't resolve the merge conflicts. Happy for someone else to do it or give me write access (or both?)

src/blueapi/cli/cli.py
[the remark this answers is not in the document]
callumforrester: The CLI does listen to events in certain cases, see `AmqClient`

src/blueapi/rest/handler.py
callumforrester: I think Joseph means should we permit the existence of a handler with no message bus? You currently force it to exist. I think the answer is "yes, but possibly out of scope for this change".

src/blueapi/rest/handler.py
```diff
@@ -0,0 +1,78 @@
...
+        self.worker.progress_events.subscribe(
+            lambda event, corr_id: self.message_bus.send(
+                "public.worker.event.progress", event, None, corr_id
+            )
+        )
+
+        self.message_bus.connect()
+
+    def stop(self) -> None:
+        self.worker.stop()
```
callumforrester, line 53:
```suggestion
        self.worker.stop()
        self.message_bus.disconnect()
```

src/blueapi/rest/handler.py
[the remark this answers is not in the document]
callumforrester: `HANDLER.stop()` doesn't return anything so this could be 


```suggestion
    if HANDLER is not None:
        HANDLER.stop()
```

callumforrester (changes requested):
Looking good, needs a bit of back and forth but definitely getting there.

src/blueapi/rest/handler.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+HANDLER: Optional[Handler] = None
```
callumforrester, line 56: Just a thought, one alternative to a global variable is to use an `lru_cache`: https://cosmiccoding.com.au/tutorials/simple_singletons/
Probably out of scope but could be worth making a ticket, what do you think?

src/blueapi/rest/handler.py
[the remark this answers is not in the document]
RAYemelyanova: yeah, this is more readable. I was trying to be fancy and do things in less lines.

src/blueapi/rest/handler.py
RAYemelyanova: Hmm, okay. Perhaps something to do for a future PR?

src/blueapi/cli/cli.py
```diff
@@ -22,29 +19,22 @@ def main(ctx) -> None:
...
 
-@main.command(name="worker")
-@click.option("-c", "--config", type=Path, help="Path to configuration YAML file")
-def start_worker(config: Optional[Path]):
-    from blueapi.service import start
-
-    start(config)
+@main.command(name="run")
+def start_application():
+    subprocess.run(
```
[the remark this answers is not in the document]
RAYemelyanova: Good idea. Looking into this now...

src/blueapi/cli/cli.py
```diff
@@ -53,54 +50,41 @@ def start_worker(config: Optional[Path]):
...
+
+
+def check_connection(func):
+    def wrapper(*args, **kwargs):
+        try:
+            func(*args, **kwargs)
+        except ConnectionError:
+            print("Failed to establish connection.")
+
+    return wrapper
```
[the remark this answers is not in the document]
RAYemelyanova: I think a `ConnectionError` will only happen if you try to do, `bluesky controller devices` and there's no server running (i.e. you haven't already done `bluesky run`). This is why I've written this wrapper, I wanted it to be really clear to the user that the server isn't running.

I'm happy to take it out, but I don't really know who is going to use the CLI in this way. If it's other engineers, sure. If it's people downloading this and playing with it for their use case, I'd rather leave it in and make the error message more descriptive like 'make sure you've typed in `bluesky run` in another terminal'.

src/blueapi/worker/reworker.py
```diff
@@ -87,11 +105,13 @@ def _cycle_with_error_handling(self) -> None:
 
     def _cycle(self) -> None:
         try:
-            LOGGER.info("Awaiting task")
-            next_task: ActiveTask = self._task_queue.get()
+            LOGGER.debug("Awaiting task")
+            next_task: ActiveTask = self._task_queue.get(timeout=1.0)
```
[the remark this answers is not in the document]
RAYemelyanova: Looks like an extension of what you've helped me with r.e. stopping the worker here in this PR. Happy to help make it not a draft.

src/blueapi/rest/handler.py
RAYemelyanova: Good idea; I don't think it's out of scope at all, It's part of the whole 'handler' business. I'd prefer to avoid global variables. Thanks for looking into this.

RAYemelyanova: The only thing I'm not sure of, is if this cache can be accessed by multiple sources at the same time. I want 3/4 simultaneous API calls to get the handler with no problem, which with a global variable will happen but I'm not sure about the cache. I'm assuming this is fine but I can't explain to myself why it's fine, and I haven't found good documentation explaining this case yet...

RAYemelyanova: ^ It is thread safe so hopefully that's a good indication it's fine. I might write some tests for this.

RAYemelyanova:
I think I've changed stuff enough now, in response to comments. Please have a look.

I'm aware that https://github.com/DiamondLightSource/blueapi/pull/145 is related to this PR, i.e. it should probably get merged first and then the lines changed for this PR will be smaller. Happy to work on this with you @callumforrester

src/blueapi/cli/cli.py
```diff
@@ -22,29 +22,15 @@ def main(ctx) -> None:
         print("Please invoke subcommand!")
 
 
-@main.command(name="worker")
-@click.option("-c", "--config", type=Path, help="Path to configuration YAML file")
-def start_worker(config: Optional[Path]):
-    from blueapi.service import start
+@click.version_option(version=__version__)
+@main.command(name="run")
```
callumforrester, line 26: Was a reason for changing the name of this command by the way?

src/blueapi/cli/cli.py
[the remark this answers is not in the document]
callumforrester: It's definitely designed to be an engineering interface, so errors with lots of detail are fine

src/blueapi/worker/reworker.py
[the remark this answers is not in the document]
callumforrester: Sounds good, pair programming session?

src/blueapi/rest/handler.py
callumforrester: Yeah exactly, maybe make a new issue about it

src/blueapi/rest/handler.py
callumforrester: I *think* the multiple access case is fine but good to double check. There is still a single object stored in the cache that is accessed by all requests. However, one thing that has just occurred to me is that the handler won't be created until the first REST call. Not sure we want this, we ideally want it to happen on startup so we know quickly if there's a problem. Maybe we can call the `lru_cache`ed function as a startup task. That will create a handler and put it in the cache.

src/blueapi/cli/cli.py
RAYemelyanova: yeah, I preferred it. `bluesky worker` made sense when you were actually spinning up a service that started the worker. But here, this command is starting a fastAPI application, which indirectly (through startup tasks) starts up a worker.

RAYemelyanova: yeah, I don't know what you prefer @callumforrester I'm happy to keep it 'worker' it just seems out of style with other cli tools?

.devcontainer/Dockerfile
```diff
@@ -34,4 +34,4 @@ ENV PATH=/venv/bin:$PATH
 
 # change this entrypoint if it is not the same as the repo
 ENTRYPOINT ["blueapi"]
-CMD ["worker"]
+CMD ["run"]
```
callumforrester, line 37: Could we put this back to `worker`? More descriptive and saves having to change the docs

.vscode/launch.json
```diff
@@ -4,6 +4,18 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+        {
+            "name": "Python: FastAPI",
```
callumforrester, line 8: Could replace the `Worker Service` with this, and also rename it to `Worker Service`

src/blueapi/cli/cli.py
callumforrester: So the problem is that the command `run` goes into the global namespace, so in your terminal you can type `ls`, `rm`, `cat` or now `run`. It definitely doesn't belong there, it needs to be called something that indicates it has anything to do with Bluesky for a start.

src/blueapi/cli/cli.py
```diff
@@ -19,24 +20,30 @@
...
 def main(ctx, config: Optional[Path]) -> None:
     # if no command is supplied, run with the options passed
+
     config_loader = ConfigLoader(ApplicationConfig)
     if config is not None:
         config_loader.use_values_from_yaml(config)
 
     ctx.ensure_object(dict)
-    ctx.obj["config"] = config_loader.load()
+    ctx.obj["config_loader"] = config_loader
```
callumforrester, line 29: Feel like this was better the old way. Now you have to call `load()` in two places instead of one

src/blueapi/config.py
```diff
@@ -35,6 +35,11 @@ class LoggingConfig(BlueapiBaseModel):
     level: LogLevel = "INFO"
 
 
+class FastApiConfig(BlueapiBaseModel):
```
callumforrester, line 38: Think this could have a more generic name, nothing ties it to FastAPI, maybe `RestConfig`?

pyproject.toml
```diff
@@ -22,6 +22,9 @@ dependencies = [
     "scanspec",
     "PyYAML",
     "click",
+    "fastapi",
+    "uvicorn",
+    "httpx",
```
callumforrester, line 27: What's this for?

callumforrester (changes requested):
Getting close, main thing is we can't have a top-level command called `run`. I also agree that `bluesky worker` was the wrong choice which is why I renamed it to `blueapi worker`, but happy to discuss further if you have objections to that.

src/blueapi/cli/cli.py
RAYemelyanova: Yes, the reason I did this, is because pydantic basemodels are not hashable. This meant I couldn't call `get_handler` with the ApplicationConfig because it has a lru_cache which by default puts it into a dictionary... so I initially went around this by just passing around the ConfigLoader instead.

However... I have since got rid of the lru_cache and decided the easiest way to deal with this is via global variables. In which case I don't need it. I will try passing around just the ApplicationConfig because I agree it's a lot neater.

src/blueapi/config.py
RAYemelyanova: good point, will change.

pyproject.toml
RAYemelyanova: for the fastAPI application to run it needs these three dependencies.

src/blueapi/cli/cli.py
RAYemelyanova: edit: yep, have changed it back and it works perfectly fine.

src/blueapi/cli/cli.py
```diff
@@ -49,39 +54,51 @@ def controller(ctx) -> None:
...
+@click.pass_obj
+def run_plan(obj: dict, name: str, parameters: str) -> None:
+    config: ApplicationConfig = obj["config"]
+
+    resp = requests.put(
+        f"http://{config.api.host}:{config.api.port}/task/{name}",
+        json=json.loads(parameters),
     )
+    print(f"Response returned with {resp.status_code}: ")
+    pprint(resp.json())
```
tpoliaw, line 104: This response/json is currently always None

pyproject.toml
tpoliaw, line 27: Should this be `requests` instead of `httpx`? It doesn't look like anything uses httpx but requests is used

src/blueapi/service/handler.py
```diff
@@ -0,0 +1,73 @@
...
+    context: BlueskyContext
+    worker: Worker
+    config: ApplicationConfig
+    message_bus: MessagingTemplate
+
+    def __init__(self, config: Optional[ApplicationConfig] = None) -> None:
+        self.context = BlueskyContext()
+        self.config = config if config is not None else ApplicationConfig()
+
+        logging.basicConfig(level=self.config.logging.level)
```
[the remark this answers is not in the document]
RAYemelyanova: I will leave this for a future PR.

src/blueapi/service/handler.py
```diff
@@ -0,0 +1,75 @@
...
+
+        self.context.with_startup_script(self.config.env.startup_script)
+
+        self.worker = RunEngineWorker(self.context)
+        self.message_bus = StompMessagingTemplate.autoconfigured(self.config.stomp)
```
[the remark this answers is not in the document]
RAYemelyanova: Good idea; will write something in for that.

src/blueapi/cli/cli.py
```diff
@@ -49,39 +54,51 @@ def controller(ctx) -> None:
...
+def get_devices(obj: dict) -> None:
+    config: ApplicationConfig = obj["config"]
 
+    resp = requests.get(f"http://{config.api.host}:{config.api.port}/devices")
+    print(f"Response returned with {resp.status_code}: ")
+    pprint(resp.json())
 
-@controller.command(name="run")
+
+@controller.command(name="serve")
```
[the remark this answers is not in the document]
RAYemelyanova: Yep... good spot. I think this has changed now.

src/blueapi/cli/cli.py
```diff
@@ -49,39 +54,51 @@ def controller(ctx) -> None:
...
-        renderer.on_progress_event,
-        timeout=120.0,
+@check_connection
+@click.pass_obj
+def run_plan(obj: dict, name: str, parameters: str) -> None:
+    config: ApplicationConfig = obj["config"]
+
+    resp = requests.put(
+        f"http://{config.api.host}:{config.api.port}/task/{name}",
+        json=json.loads(parameters),
```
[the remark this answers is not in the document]
RAYemelyanova: Have fixed the differences between running plans; now it should work as it used to, i.e. the first way you specified in this comment.

pyproject.toml
RAYemelyanova: Here's what happens if you do that, looking at CI test jobs:

```
ImportError while loading conftest '/home/runner/work/blueapi/blueapi/tests/conftest.py'.
tests/conftest.py:4: in <module>
    from fastapi.testclient import TestClient
/opt/hostedtoolcache/Python/3.9.[16](https://github.com/DiamondLightSource/blueapi/actions/runs/4927983607/jobs/8805745964?pr=135#step:6:17)/x64/lib/python3.9/site-packages/fastapi/testclient.py:1: in <module>
    from starlette.testclient import TestClient as TestClient  # noqa
/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/starlette/testclient.py:16: in <module>
    import httpx
E   ModuleNotFoundError: No module named 'httpx'
```

RAYemelyanova: have now fixed this; turns out I just had to change `fastapi` to `fastapi[all]` in the pyproject.toml.

RAYemelyanova:
just need to write some tests for the handler to increase code coverage

src/blueapi/cli/cli.py
```diff
@@ -30,58 +35,74 @@ def main(ctx, config: Optional[Path]) -> None:
...
+def start_application(obj: dict):
+    start(obj["config"])
 
-    config: ApplicationConfig = obj["config"]
-    start(config)
+
+@main.command(name="worker", deprecated=True)
+@click.pass_obj
+def deprecated_start_application(obj: dict):
+    print("Please use run command instead.\n")
```
tpoliaw, line 47: should this now be serve command as well?

src/blueapi/service/handler.py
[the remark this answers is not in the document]
callumforrester: I agree but think we should sort this out in a separate PR to avoid scope creep.

src/blueapi/cli/fake_cli.py
tpoliaw, line 1: What's this module for?

src/blueapi/service/main.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@app.put("/task/{name}")
+async def execute_task(
+    name: str,
+    task: Mapping[str, Any] = Body(..., example={"detectors": ["x"]}),
+    handler: Handler = Depends(get_handler),
+):
+    handler.worker.submit_task(name, RunPlan(name=name, params=task))
+
```
callumforrester, line 54: Should return a `TaskResponse`

src/blueapi/service/main.py
```diff
@@ -0,0 +1,61 @@
...
+@app.get("/device/{name}", response_model=DeviceModel)
+async def get_device_by_name(name: str, handler: Handler = Depends(get_handler)):
+    try:
+        return DeviceModel.from_device(handler.context.devices[name])
+    except KeyError:
+        raise HTTPException(status_code=404, detail="Item not found")
+
+
+@app.put("/task/{name}")
+async def execute_task(
```
callumforrester, line 48:
```suggestion
async def submit_task(
```

src/blueapi/service/main.py
```diff
@@ -0,0 +1,61 @@
...
+from blueapi.worker import RunPlan
+
+from .handler import Handler, get_handler, setup_handler, teardown_handler
+from .model import DeviceModel, DeviceResponse, PlanModel, PlanResponse
+
+app = FastAPI(docs_url="/docs", on_shutdown=[teardown_handler])
+
+
+@app.get("/plans", response_model=PlanResponse)
+async def get_plans(handler: Handler = Depends(get_handler)):
```
callumforrester, line 15: Docstrings on routes

src/blueapi/service/main.py
```diff
@@ -38,19 +41,22 @@ async def get_devices(handler: Handler = Depends(get_handler)):
...
-async def execute_task(
+@app.put("/task/{name}", response_model=TaskResponse)
+async def submit_task(
     name: str,
     task: Mapping[str, Any] = Body(..., example={"detectors": ["x"]}),
     handler: Handler = Depends(get_handler),
 ):
+    """Submit a task onto the worker queue."""
     handler.worker.submit_task(name, RunPlan(name=name, params=task))
+    return TaskResponse(task_name=f"Task {name} submitted")
```
callumforrester, line 59: This needs to literally be the same name, it is meant to be computer-readable, not human-readable.
```suggestion
    return TaskResponse(task_name=name)
```

RAYemelyanova:
This is ready to be merged, but here are a few more ideas for future PRs:
1. the `service/handler.py::Handler` should gracefully exit if a stomp connection cannot be established,
2. Need to deal with correlation ID's
3. Maybe don't use a global `HANDLER`, but for this some restructuring might be necessary.

callumforrester (approved):

DiamondJoseph (approved):

Issue #113, opened 2023-04-05
Add catalog-info.yaml


DiamondJoseph:
APIs for: running plans, subscribing, emitting documents (TODO: Should Documents be defined centrally?), sample information (post #112), ...

keithralphs:
Completed in #137

PR #137, opened 2023-04-21
add catalog-info for backstage

RAYemelyanova:
related to https://github.com/DiamondLightSource/blueapi/issues/113


RAYemelyanova:
This should not be merged until https://github.com/DiamondLightSource/blueapi/pull/135 is merged.

keithralphs:
AsyncAPI yaml file has now been added at docs/user/reference/asyncapi.yaml, so please add a reference to this to the cataloginfo file. Also please put your openAPI file in the same folder and update the reference to it accordingly.

DiamondJoseph:
Be nice to have [this validator for the catalog-info](https://github.com/RoadieHQ/backstage-entity-validator), but I'm happy to add it once this is over the line

keithralphs:
Also be good to add   

```
annotations:
    github.com/project-slug: DiamondLightSouce/blueapi
```

to the component>metadata section so that the CI and other stuff is visble in backstage

catalog-info.yaml
```diff
@@ -0,0 +1,43 @@
...
+    github.com/project-slug: DiamondLightSouce/blueapi
+spec:
+  type: service
+  lifecycle: production
+  owner: user:vid18871 # TODO: owner: DAQ-Core
+  #  system: Athena  # TODO: Define Athena system: presumably same location as DAQ-Core/DAQ?
+  providesApis:
+    - message-topics
+    - restapi
+    - blueskydocument-to-scanmessage
```
DiamondJoseph, line 17: I think the ugliness of this (since it's not actually emitting ScanMessages, and ScanMessages are defined elsewhere, and this whole provided API is defined elsewhere) is reason enough to say: now is the time to do #167  and fold it into this change.

src/blueapi/service/openapi.py
DiamondJoseph, line 1: Can this be defined in a workflow instead of in the source code? Is there already a workflow to handle autogeneration of OpenAPI Schemas?

DiamondJoseph (changes requested):

catalog-info.yaml
keithralphs: I would prefer to address that in a separate change that also includes updating the AsyncAPI definition to reflect that some of it's functionality is now repaced by the REST API.  That will allow us to close this ticket and its associate Jira one and schedule the cleanup activity for next sprint.

src/blueapi/service/openapi.py
keithralphs: That would mean we couldn't update the OpenAPI definition without submitting the change to GitHub, which sounds wrong to me; I think it should exist in the repo before GitHub is involved to facilitate testing of the API.

DiamondJoseph (approved):

docs/user/reference/openapi.json
```diff
@@ -0,0 +1,328 @@
+{
```
callumforrester, line 1: I've made an issue to deal with the fact that this in a different markup langauge to the AyncAPI spec. Not merge-blocking, can be dealt with later: https://github.com/DiamondLightSource/blueapi/issues/181

docs/user/reference/openapi.rst
```diff
@@ -0,0 +1,5 @@
+OpenAPI Specification
```
callumforrester, line 1: You can actually get sphinx to render the spec in a swagger-like way with a plugin, see example in scanspec: https://github.com/dls-controls/scanspec/blob/master/docs/reference/rest_api.rst

Also, I think we should rethink calling the docs "openapi.rst" and "asyncapi.rst", as that only defines the type, not the purpose, of the API they are documenting.

callumforrester (changes requested):

docs/user/reference/openapi.rst
RAYemelyanova: Yes, I agree. I do think however that renaming the docs is maybe a separate issue that can tie into #181?

RAYemelyanova: Gonna have a look at that plugin though.

callumforrester (approved):

PR #145, opened 2023-04-24
Make worker asynchronous and stoppable

callumforrester:
Changes:
* Change worker API from `run_forever` to `stop` and `start`
* Make `start` run worker in its own thread

### DIFF
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -50,6 +50,7 @@ class RunEngineWorker(Worker[Task]):
     _worker_events: EventPublisher[WorkerEvent]
     _progress_events: EventPublisher[ProgressEvent]
     _data_events: EventPublisher[DataEvent]
+    _started: Event
     _stopping: Event
     _stopped: Event
 
@@ -68,6 +69,7 @@ class RunEngineWorker(Worker[Task]):
         self._data_events = EventPublisher()
         self._status_lock = RLock()
         self._status_snapshot = {}
+        self._started = Event()
         self._stopping = Event()
         self._stopped = Event()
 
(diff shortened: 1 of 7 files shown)


callumforrester:
@RAYemelyanova I think this this is ready now

tests/worker/test_reworker.py
```diff
@@ -0,0 +1,127 @@
...
+    worker = RunEngineWorker(context)
+    yield worker
+    worker.stop()
+
+
+def test_stop_doesnt_hang(worker: Worker) -> None:
+    worker.start()
+
+
+def test_stop_is_idempotent(worker: Worker) -> None:
```
RAYemelyanova, line 37: nit-pick: I feel like if I have to google what idempotent means, and then still don't understand why this test name makes sense, maybe it's a bad test name?

Idempotent; denoting an element of a set which is unchanged in value when multiplied or otherwise operated on by itself.

RAYemelyanova: aha - ok, I get it now. You don't call worker.start(). So you're testing that worker.stop() does nothing if the worker isn't started.

So perhaps, instead, 
```python
test_stop_is_idempontent_if_worker_not_started(worker: Worker) -> None:
```

src/blueapi/worker/reworker.py
```diff
@@ -64,20 +69,45 @@ def __init__(
...
+            raise Exception("Worker is already running")
+        run_worker_in_own_thread(self)
+
+    def stop(self) -> None:
+        LOGGER.info("Attempting to stop worker")
+
+        # If the worker has not yet started there is nothing to do.
+        if self._started.is_set():
+            self._task_queue.put(KillSignal())
+            self._stopped.wait(timeout=30.0)
```
RAYemelyanova, line 92: should we make this timeout some global var like BLUESKY_TIMEOUT that is set elsewhere?

RAYemelyanova (approved):
looks good, I've got a couple of comments about small things.

PR #147, opened 2023-04-24
Make BlueskyContext accept all objects with a name property as devices

callumforrester:
Previously, BlueskyContext only accepted objects that conformed to the `Readable` or `Flyable` protocols  as devices. This is inherited from the days when all devices conformed to these protocols or their inheritors. Now the protocols no longer have an inheritance heirachy and are mixed and matched through multiple inheritance. The only sensible criteria is whether an object has a `name` property. This opens the door to abuse, storing objects that are definitely not devices, but that may be the price to pay for the required flexibility.

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -6,7 +6,6 @@ from types import ModuleType
 from typing import Dict, Iterable, List, Optional, Union
 
 from bluesky import RunEngine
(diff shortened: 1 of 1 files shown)


callumforrester:
> 

Agreed, made an issue: https://github.com/DiamondLightSource/blueapi/issues/148

PR #150, opened 2023-04-25
Support type validator models from partials as well as functions

callumforrester:
* Type validators can create models from function arguments
* Also support partials


callumforrester:
Obsolete thanks to #154, but maybe still of interest @DiamondJoseph

PR #152, opened 2023-04-25
Create documentation page about events 

callumforrester:
* Add page about events to developer documentation
* Explain why we need to add new events on top of the event model
* Explain the use of correlation IDs


callumforrester:
Still to do: make API reference and point to that

docs/developer/explanations/events.rst
```diff
@@ -0,0 +1,67 @@
+Events Emitted by the Worker
+============================
+
+Blueapi adds new events on top of the `bluesky event model`_.
+
+Reasons
+-------
+
+Since the ``RunEngine`` is traditionally used by a human in front of an IPython terminal, it
+sometimes assumes intuitive behavoir. The worker replaces the human and so must fill in the
```
DiamondJoseph, line 10: behavoir -> behaviour

DiamondJoseph (changes requested):
nit, otherwise LGTM

DiamondJoseph:
Presumably going to need an update for emitting state events on same topic?

keithralphs:
> Presumably going to need an update for emitting state events on same topic?

Do that as part of addressing #176

PR #154, opened 2023-04-27
Deserialise plan method arguments by replacing types instead of building custom validators

### LINKED ISSUE #118: Support scanspec deserialization in plans with pydantic
With apischema we could have `Spec[Movable]` and custom converters to deserialize a scanspec while simultaneously looking up devices. This has proved hard with pydantic and has required several workarounds. See plans.py in https://github.com/DiamondLightSource/blueapi/pull/90/files

### LINKED ISSUE #119: Move plan parameter deserialization is broken
The move plan cannot process it's arguments following the move to pydantic. It gives the following error:
```shell
ERROR:blueapi.worker.reworker:1 validation error for ParsingModel[group]
__root__ -> moves
  'int' object is not subscriptable (type=type_error)
Traceback (most recent call last):
  File "/home/vid18871/projects/blueapi/src/blueapi/worker/reworker.py", line 94, in _cycle
    self._current.task.do_task(self._ctx)
  File "/home/vid18871/projects/blueapi/src/blueapi/worker/task.py", line 45, in do_task
    """
  File "/home/vid18871/projects/blueapi/src/blueapi/worker/task.py", line 66, in _lookup_params
    sanitized_params = lookup_params(ctx, plan, self.params)
  File "/home/vid18871/projects/blueapi/venv/lib/python3.8/site-packages/pydantic/tools.py", line 38, in parse_obj_as
    return model_type(__root__=obj).__root__
  File "/home/vid18871/projects/blueapi/venv/lib/python3.8/site-packages/pydantic/main.py", line 341, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for ParsingModel[group]
__root__ -> moves
  'int' object is not subscriptable (type=type_error)
```

tpoliaw:
Proof of concept for more flexible type checking/conversion of device types. 

Should in theory fix #119

Not tried passing `Spec[Movable]` yet but if `Spec` is defined as something like

```python
class Spec(Generic(TypeVar('T')):
    axis: T
```
it should also fix #118


callumforrester:
Makes sense and much more concise, good job! We also need to make sure we don't lose the functionality introduced by #131

tpoliaw:
> We also need to make sure we don't lose the functionality introduced by #131

Should now handle concrete types as well as protocols.

src/blueapi/core/context.py
```diff
@@ -151,13 +159,88 @@ def device(self, device: Device, name: Optional[str] = None) -> None:
...
+
+    Args:
+        typ: The type that is required - potentially referencing Bluesky protocols
+        ctx: The context to be used to look up references to devices.
+
+    Returns:
+        A Type that can be deserialised by Pydantic
+    """
+    if typ is Parameter.empty:
+        return Any  # type: ignore
```
tpoliaw, line 236: I don't know why it doesn't accept returning `Any` here. Is `Any` not a type?

tests/core/test_context.py
```diff
@@ -167,3 +170,27 @@ def test_add_non_plan(empty_context: BlueskyContext) -> None:
...
+    assert empty_context._reference(Movable) is empty_context._reference(Movable)
+    assert empty_context._reference(Movable) is not empty_context._reference(Readable)
+
+
+def test_reference_type_conversion(empty_context: BlueskyContext) -> None:
+    movable_ref: Type = empty_context._reference(Movable)
+    assert context._convert_type(Movable, empty_context) == movable_ref
+    assert (
+        context._convert_type(dict[Movable, list[tuple[int, Movable]]], empty_context)
+        == dict[movable_ref, list[tuple[int, movable_ref]]]  # type: ignore
```
tpoliaw, line 195: Again, not sure about types here. It says `movable_ref` can't be used as a type.

docs/developer/explanations/type_validators.rst
```diff
@@ -52,55 +52,48 @@ to be devices and look them up against the context. For example with the request
...
-        d: List[List[Readable]], 
+        a: Readable,
+        b: List[Readable],
+        c: Dict[str, Readable],
+        d: List[List[Readable]],
         e: List[Dict[str, Set[Readable]]]) -> Plan:
         ...
 
 
 Implementation Details
```
DiamondJoseph, line 96: I think the referencing is explained sufficiently above, so can lose the heading.

DiamondJoseph:
I think enforcing typing makes our life easier long term, as plans are eventually going under code review then sitting in repositories for years while support staff change until they're updated and makes potential auto-generation for GUIs etc. possible down the line. It seems a very small requirement that is easily enforced and has potential benefit.

src/blueapi/core/context.py
tpoliaw: No longer relevant as check removed

tpoliaw:
Checks to enforce at least an `Any` annotation are back in

docs/developer/explanations/type_validators.rst
```diff
@@ -52,55 +52,45 @@ to be devices and look them up against the context. For example with the request
...
+
+These new intermediate types are used only in the deserialisation process. The
+object returned from validator method is not checked by pydantic so it can be
+the actual instance and the plan never sees the runtime generated reference
+type, only the type it was expecting.
+
+NB. This uses the fact that the new types generated at runtime have access to
+the context that required them via their closure. This circumvents the usual
+problem of pydantic validation not being able to access external state when
+validating or deserialising.
```
callumforrester, line 69: Sphinx has a proper formatting mechanism for this:
```suggestion
.. note:: This uses the fact that the new types generated at runtime have access to the context that required them via their closure. This circumvents the usual problem of pydantic validation not being able to access external state when validating or deserialising.
```

callumforrester (approved):
Noted a few things but nothing merge-blocking. Thanks for this!

tests/core/test_context.py
tpoliaw: Maybe, but it's being returned by a function that is annotated as returning `Type`

DiamondJoseph (approved):

Issue #157, opened 2023-05-02
Throw an error if asked to run a plan while a plan is already in progress


callumforrester:
Behaviour consistent with artemis. @RAYemelyanova as discussed.

PR #158, opened 2023-05-02
make worker queue have a size of 1

RAYemelyanova:
made queue have one thing in it. Raising error if attempt to submit while worker is running

### DIFF
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -15,7 +15,6 @@ from blueapi.core import (
     EventStream,
     WatchableStatus,
 )
-from blueapi.worker.worker_busy_error import WorkerBusyError
 
 from .event import (
     ProgressEvent,
@@ -28,6 +27,7 @@ from .event import (
 from .multithread import run_worker_in_own_thread
 from .task import ActiveTask, Task
 from .worker import Worker
+from .worker_busy_error import WorkerBusyError
 
 LOGGER = logging.getLogger(__name__)
 
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -71,7 +71,7 @@ class RunEngineWorker(Worker[Task]):
         self._state = WorkerState.from_bluesky_state(ctx.run_engine.state)
         self._errors = []
         self._warnings = []
-        self._task_queue = Queue()
+        self._task_queue = Queue(maxsize=1)
         self._current = None
         self._worker_events = EventPublisher()
         self._progress_events = EventPublisher()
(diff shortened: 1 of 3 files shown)


RAYemelyanova:
I would write tests for this, but we need https://github.com/DiamondLightSource/blueapi/pull/145 to be merged before we can do that. So I think this should be a WIP until that is done.

RAYemelyanova:
solves https://github.com/DiamondLightSource/blueapi/issues/157

src/blueapi/worker/reworker.py
```diff
@@ -68,6 +68,10 @@ def __init__(
     def submit_task(self, name: str, task: Task) -> None:
         active_task = ActiveTask(name, task)
         LOGGER.info(f"Submitting: {active_task}")
+        if self._task_queue.qsize() != 0:
+            LOGGER.error("Cannot submit task while another is running")
+            raise Exception("Cannot submit task while another is running")
```
callumforrester, line 73: Would be good if we could use a more specific type of exception. It's a shame there's no `BusyError` or similar. Feel free to make one up if you can't find something close enough.

src/blueapi/worker/reworker.py
```diff
@@ -68,6 +68,10 @@ def __init__(
     def submit_task(self, name: str, task: Task) -> None:
         active_task = ActiveTask(name, task)
         LOGGER.info(f"Submitting: {active_task}")
+        if self._task_queue.qsize() != 0:
```
callumforrester, line 71: `qsize()` is approximate and unreliable, per the docs https://docs.python.org/3/library/queue.html#queue.Queue.qsize

The thread-safe solution is something like this
```python
from queue import Queue
from queue import Full
q = Queue(maxsize=1)
try:
        q.put_nowait(1)
        q.put_nowait(2)
    except Full:
        raise Exception("Whatever my exception is")
```

RAYemelyanova: thanks, I was wondering about this... will change now.

src/blueapi/worker/reworker.py
```diff
@@ -15,6 +15,7 @@
     EventStream,
     WatchableStatus,
 )
+from blueapi.worker.worker_busy_error import WorkerBusyError
```
callumforrester, line 18: I would suggest a relative import since `WorkerBusyError` is never actually leaving the `worker` package

```suggestion
from .worker_busy_error import WorkerBusyError
```

tests/worker/test_reworker.py
```diff
@@ -125,3 +126,12 @@ def on_event(event: E, event_id: Optional[str]) -> None:
     sub = stream.subscribe(on_event)
     future.add_done_callback(lambda _: stream.unsubscribe(sub))
     return future
+
+
+def test_worker_only_accepts_one_task_on_queue(worker: Worker, timeout: float = 5.0):
```
callumforrester, line 131: Timeout is unused

callumforrester (approved):

PR #160, opened 2023-05-02
Extract message definitions in schema to allow referencing from other APIs

DiamondJoseph:


DiamondJoseph:
Have made a pull request with the updated AsyncAPI parser to try and get the github action enabled, but will disable for now.

keithralphs (approved):
Looks good now

.github/workflows/api.yaml
```diff
@@ -0,0 +1,17 @@
+#name: API CI
```
callumforrester, line 1: Would prefer this was deleted and put in an issue

callumforrester (changes requested):
Would prefer CI job was deleted and put in an issue

callumforrester (approved):

Issue #161, opened 2023-05-02
Use random UUID for status_name in StatusView


joeshannon:
Currently the uuid assigned for `StatusView` uses `uuid1`. This is not a fully random uuid which makes it a bit difficult to determine different statuses when looking at logs.

PR #162, opened 2023-05-02
Use random UUID for StatusView names

joeshannon:
This generates names which are more clearly distinct when looking through logs.


joeshannon:
Fixes #161

DiamondJoseph (commented):
uuid1 is more guaranteed to be unique, but uuid4 is much more obvious when two uuids aren't the same when reading the logs. I think the odds of collision over the lifetime of the service is so low that this is worth using as the standard generator but I leave the final decision to @callumforrester

callumforrester (approved):
Happy to do so, I'm also not married to uuids, we can use an atomic counter or something if that helps.

DiamondJoseph (approved):

PR #163, opened 2023-05-03
Rename CLI to blueapi and deprecate old one

callumforrester:
From discussion with @joeshannon, it's probably best if the old CLI command `bluesky` goes away to be replaced with `blueapi` to avoid confusion.


DiamondJoseph (approved):
LGTM

Do we have a deprecation cycle? Would be nice to formalise something early

callumforrester:
Hmm, good question. We don't have a formal release cycle yet. Could be time-based?

callumforrester:
Have decided for this particular case it will be 1 sprint: https://github.com/DiamondLightSource/blueapi/issues/164 
We should think about a more general approach.

PR #165, opened 2023-05-03
Document Command Line Interface

callumforrester:
Add docs to CLI:
* How-to page on basic usage
* Reference page with full options etc.
* Descriptions of commands that appear when the user types `--help`


callumforrester:
@abbiemery Do we think this lives in user docs or should be moved to dev docs?

abbiemery (commented):
Looks good to me, but I feel like it should follow the instructions on how to run the worker and active mq. I think both sets of instructions should probably be in user. The users here are mostly developers and will be setting up the development environment to play with, however they won't be actually developing blueapi, so I think that could be a sensible place to draw the line. I feel like it also most appropriately represents the beamline scientist position.

callumforrester:
I too was worried about where the line sits between user and developer and had a chat with @coretl a few days ago.
I agree with where you've drawn the line. The instructions on how to run the worker and AMQ in dev guide are specifically for people who are developing blueapi. I think we actually cover both if I just add something about ActiveMQ in the CLI docs here, what do you think?

abbiemery:
Are the instructions in dev specifically for developing blueapi? They seam generic to me for the use case of other daq teams trying out and setting up athena elements to play with; run local messaging, run local bluesky worker.  I know the goal is to have end users to use a server, so they only need to know how to submit plans and it'll be in k8. In which case have they even got as far as finding the repo?

You are right there is a lot of overlap here so there is no one clear answer. I would potentially at least add something in the user tutorial docs which says: If you haven't got a beamline set-up already, and you want a local instance to play with go to dev page.

docs/user/how-to/run-cli.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. code:: shell
+
+    # Run the sleep plan
+    blueapi controller run sleep -p '{"time": 5.0}'
+
+    # Run the count plan
+    blueapi controller run count -p '{"detectors": ["current_det", "image_det"]}'
+
+The command will block until the plan is finished and will forward error/status messages
```
tpoliaw, line 77: The run command no longer blocks or forwards status messages

docs/user/how-to/run-cli.rst
```diff
@@ -0,0 +1,80 @@
...
+-------------------
+
+The worker can tell you which plans and devices are available via:
+
+.. code:: shell
+
+    blueapi controller plans
+    blueapi controller devices
+
+By default, the CLI will talk to the worker via a message broker on ``tcp://localhost:61613``,
```
tpoliaw, line 57: Doesn't it default to using rest/http now?

tpoliaw (dismissed):

docs/user/how-to/run-cli.rst
callumforrester: It does both, rest for commands, message bus for events

docs/user/how-to/run-cli.rst
callumforrester: Hmm, it should, will restore that functionality...

docs/user/how-to/run-cli.rst
tpoliaw: I don't think the controller section of the CLI uses aMQ at all.

docs/user/how-to/run-cli.rst
```diff
@@ -0,0 +1,80 @@
+Control the Worker via the CLI
+==============================
+
+The worker comes with a minimal CLI client for basic control. It should be noted that this is 
+a test/development/debugging tool and not meant for production!
```
RAYemelyanova, line 5: We use the CLI for the dockerfile. In future will the server be spun up any differently?

docs/user/how-to/run-cli.rst
```diff
@@ -0,0 +1,80 @@
...
+
+
+Start the Worker
+----------------
+
+To start the worker
+
+.. code:: shell
+
+    blueapi serve
```
RAYemelyanova, line 44: might be worth mentioning about the config. I've just realised that, unfortunately the following does not work:

```
blueapi --config some/file.yaml
blueapi serve
```
i.e. the `blueapi serve` command will not use the configuration from the first command. Is this a problem at all? You would need to do;

```
blueapi --config some/path.yaml serve
```

For config to be ingested to the server.

docs/user/how-to/run-cli.rst
RAYemelyanova: ^ in a future PR I will integrate aMQ for running plans. But not for getting devices or plans.

docs/user/how-to/run-cli.rst
RAYemelyanova: In future the CLI run command will be blocking.

RAYemelyanova (commented):
Just some general comments/discussion points.

docs/user/how-to/run-cli.rst
callumforrester: Good point, I forgot that this page refers only to the `blueapi controller` CLI. Have moved the stuff about ActiveMQ and the `serve` command to a new page called "Quckstart Guide".

docs/user/how-to/run-cli.rst
callumforrester: Done in quickstart guide

tpoliaw: That's the behaviour I'd expect. I think it would be odd if the config persisted from one call to the next.

callumforrester: Agreed

docs/user/how-to/run-cli.rst
callumforrester: Now done

docs/user/how-to/run-cli.rst
callumforrester: Now done

PR #169, opened 2023-05-03
Simplify config

RAYemelyanova:
This PR aims to make config a little bit neater, so there is one single source of truth for configuration.


RAYemelyanova:
Wanted to do this before merging https://github.com/DiamondLightSource/blueapi/pull/135 (which has now become a bit messy so I was just going to do a new branch...)

RAYemelyanova:
Everything seems fine, except @callumforrester it seems like the entrypoints you've added for `blueapi` as opposed to `bluesky` don't work.

I tried doing:
```
blueapi worker
```
and got 'command not found'... not sure why. There seems to be an import error with cli._deprecated.py as well, i.e. calling it directly as an entrypoint doesn't seem to work.

src/blueapi/cli/cli.py
```diff
@@ -21,54 +19,36 @@ def main(ctx) -> None:
...
+        if DEFAULT_YAML_PATH.exists()
+        else config_loader.load()
+    )
+
 
 @main.command(name="worker")
-@click.option("-c", "--config", type=Path, help="Path to configuration YAML file")
-def start_worker(config: Optional[Path]):
+@click.pass_obj
+def start_worker(obj: dict) -> None:
```
DiamondJoseph, line 34: Should be able to tighten up to dict[str, BaseModel]?

src/blueapi/cli/cli.py
```diff
@@ -21,54 +19,36 @@ def main(ctx) -> None:
     if ctx.invoked_subcommand is None:
         print("Please invoke subcommand!")
 
+    config_loader = ConfigLoader(ApplicationConfig)
+
+    ctx.ensure_object(dict)
+    ctx.obj["config"] = (
+        config_loader.load_from_yaml(DEFAULT_YAML_PATH)
```
DiamondJoseph, line 26: So only allowing loading from DEFAULT_YAML_PATH, and assuming that file is changed (by mounting configMap or by editing), rather than passing -c path/to/config.yaml?

I understand your reasoning but will highlight for Callum so you can discuss with him.

src/blueapi/cli/cli.py
```diff
@@ -21,54 +19,36 @@ def main(ctx) -> None:
...
+        if DEFAULT_YAML_PATH.exists()
+        else config_loader.load()
```
DiamondJoseph, line 28: is config_load.load() with empty dict a valid configuration?

src/blueapi/service/app.py
```diff
@@ -29,7 +27,7 @@ class Service:
     def __init__(self, config: ApplicationConfig) -> None:
         self._config = config
         self._ctx = BlueskyContext()
-        self._ctx.with_startup_script(self._config.env.startup_script)
+        self._ctx.with_startup_script(config.env.startup_script)
```
DiamondJoseph, line 30: Can revert this line as not making any functional change and preserves the git blame

src/blueapi/config.py
```diff
@@ -36,3 +40,68 @@ class ApplicationConfig(BlueapiBaseModel):
...
+        dataclass.
+
+        Returns:
+            C: Dataclass instance holding config
+        """
+
+        try:
+            return parse_obj_as(self._schema, self._values)
+        except ValidationError:
+            raise InvalidConfigError(
```
DiamondJoseph, line 104: I would allow the ValidationError to propagate, it's already indicative of what the error is, and allows you to get rid of InvalidConfigError

tests/config_yaml_files/config.yaml
DiamondJoseph, line 1: Probably more obvious what this directory is if it's `example` or `testfiles` or similar

DiamondJoseph (changes requested):
Concept looks about right, just some changes to strip it down a bit

src/blueapi/cli/cli.py
RAYemelyanova: Sorry; this has since changed.
Turns out you can't modify context between child click commands. Here, 'worker' and 'controller' are child commands of 'main'. So if I pass in config to 'worker', and modify the object/context within worker, it has no difference to the context 'controller' receives. 

For now, my solution is to make 'main' take config instead. This way, config options actually propagate to both children.

In future with the REST api stuff this may not be needed, as everything the controller does now will be done via rest api. In that case we can migrate 'config' back to worker.

DiamondJoseph: Would that ever happen? The workers is the RunEngine worker with access to the devices, the Controller is the client that sends requests to the Worker remotely. Both shouldn't be called on the same invocation

src/blueapi/cli/cli.py
RAYemelyanova: yep; the schema given into `ConfigLoader` must be of type `BaseSettings`. These can have defaults or default factories. In the case of the ApplicationContext, every field has a default or default factory.

src/blueapi/config.py
RAYemelyanova: I think the InvalidConfigError is a bit more descriptive, it states the expected schema for example and says that the config file is wrong. The Validation Error is also useful, and still included in the traceback when this is raised.

src/blueapi/cli/cli.py
RAYemelyanova: It's more that you'd expect the configuration would be shared between both.

The invocation isn't shared, it's just if we want a non-global config that is passed around the application it needs to happen at the parent level, at 'main'. Happy to discuss more.

src/blueapi/config.py
DiamondJoseph: I thought that would require `from`?

src/blueapi_config.yaml
```diff
@@ -0,0 +1,7 @@
+env:
```
callumforrester, line 1: Rather than `src`, should this go in `config`?

src/blueapi/cli/cli.py
```diff
@@ -15,60 +14,49 @@
...
 @click.pass_context
-def main(ctx) -> None:
+def main(ctx, config: Path) -> None:
     # if no command is supplied, run with the options passed
+    config_loader = ConfigLoader(ApplicationConfig)
+
+    ctx.ensure_object(dict)
+    ctx.obj["config"] = (
+        config_loader.load_from_yaml(config)
+        if DEFAULT_YAML_PATH.exists()
```
callumforrester, line 32: I am unsure if this will survive packaging, i.e. when you pip install blueapi, I think this may no longer be true.

src/blueapi/config.py
RAYemelyanova: Yeah, it seems not to need it. I can write a test for this maybe...

src/blueapi/cli/cli.py
RAYemelyanova: why not? Is it an issue with the global variable `DEFAULT_YAML_PATH`?

src/blueapi_config.yaml
RAYemelyanova: Initially I would have preferred it to go in the top level directory, i.e. even before src/ ... But perhaps for pypi purposes that isn't a good idea? I'm not really sure.

I don't know how you're expecting blueapi to be used - i.e. thought it would just be in a container.

docs/developer/how-to/configure-app.rst
```diff
@@ -0,0 +1,19 @@
+Configure the application
```
callumforrester, line 1: Could this go in the user docs?

tests/test_config.py
```diff
@@ -86,19 +87,27 @@ def test_inject_values_into_defaultless_schema() -> None:
...
+    loader.use_values_from_yaml(override_config_yaml)
+
     assert loader.load() == ConfigWithDefaults(foo=7)
+
+
+def test_error_thrown_if_schema_does_not_match_yaml(nested_config_yaml: Path) -> None:
+    loader = ConfigLoader(Config)
+    loader.use_values_from_yaml(nested_config_yaml)
+    with pytest.raises(InvalidConfigError):
+        loader.load()
```
DiamondJoseph, line 113: Having a test that constructs a ConfigLoader without any values, and shows it is equal to the a ConfigLoader with DEFAULT_YAML would be a convenient way of knowing that a. that file always validates and b. that we're exposing most of the config in it as an example to be built upon.

RAYemelyanova: Ok, all done now.

callumforrester (approved):

DiamondJoseph (approved):

PR #170, opened 2023-05-04
Remove servers section as it is optional

keithralphs:
As discussed, remove the server section as it is not appropriate for BlueAPI which only has internal interfaces.


DiamondJoseph (approved):
LGTM

PR #171, opened 2023-05-04
121 dodal integration

### LINKED ISSUE #121: Dodal integration
Should be possible to make devices from a dodal beamline module.
https://github.com/DiamondLightSource/blueapi/issues/120

abbiemery:
Fixes #121 .

Needs to have [this](https://github.com/DiamondLightSource/dodal/pull/44) dodal pr to be merged to work, otherwise the devices that use other devices in the example start up will not work.

It also fails on the 3.8 tests, so maybe we remove the support for that first.


callumforrester:
The CI should be fixed if you rebase on `main`

src/blueapi/config.py
```diff
@@ -18,7 +23,16 @@ class EnvironmentConfig(BaseModel):
     Config for the RunEngine environment
     """
 
-    startup_script: Union[Path, str] = "blueapi.startup.example"
+    sources: list[Source] = [
```
callumforrester, line 26: Shouldn't this be in `EnvironmentConfig`?

callumforrester (changes requested):
Looking good! Only a few small items.

src/blueapi/config.py
callumforrester: Nevermind, I can't read!

src/blueapi/startup/example.py
```diff
@@ -1,38 +1,74 @@
...
-    name="theta", delay=0.2, events_per_move=12, egu="degrees"
-)
-x_err = BrokenSynAxis(name="x_err", timeout=1.0)
-sample_pressure = SynAxisWithMotionEvents(
-    name="sample_pressure", delay=30.0, events_per_move=128, egu="MPa", value=0.101
-)
-sample_temperature = SynSignal(
-    func=lambda: ((x.position + y.position + z.position) / 1000.0) + 20.0,
+
+def x(name="x") -> SynAxisWithMotionEvents:
```
tpoliaw, line 6: Is the `name` field always required and is it always going to be the same as the function name?

src/blueapi/config.py
```diff
@@ -1,9 +1,14 @@
 from pathlib import Path
-from typing import Union
+from typing import TypedDict, Union
 
 from pydantic import BaseModel, Field
 
 
+class Source(TypedDict):
```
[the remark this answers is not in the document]
abbiemery: I planned to do this but forgot, oops.

src/blueapi/core/context.py
```diff
@@ -7,6 +7,7 @@
 
 from bluesky import RunEngine
 from bluesky.protocols import Flyable, Readable
+from dodal.utils import make_all_devices
```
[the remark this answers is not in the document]
abbiemery: I did add it but accidentially removed it in my last commit, well spotted. 

Also this was discussed briefly at the time and the code was already there for us to use so @callumforrester suggested we use it. We then rewrote all of that code anyway so it became slightly moot. I would be in favour of having our own util for it, however it then is the same code and we risk incompatibility with dodal if they change stuff up on their end and our `make_all_devices` doesn't keep up with those.

Happy to do it either way and will listen to any compelling opinion.

src/blueapi/startup/example.py
abbiemery: No it is not required as an argument, as long as the device return has one. I just copied the dodal format. It also doesn't matter what the function is called. The function name is used to determine if others depend on it and instantiate the devices in order, but the resulting dictionary is `device.name: device`.

abbiemery:
@callumforrester I made the changes you requested. I also added a better test case for the context which will cause the CI to fail untill the `make_all_devices` changes are implemented. Either by dodal or if we choose to migrate the method to the blueapi utils.

src/blueapi/startup/example.py
callumforrester: It's not always going to be the same as the function name: https://nsls-ii.github.io/ophyd/architecture.html#names

src/blueapi/core/context.py
```diff
@@ -63,9 +64,10 @@ def find_device(self, addr: Union[str, List[str]]) -> Optional[Device]:
         else:
             return find_component(self.devices, addr)
 
-    def with_startup_script(self, path: Union[Path, str]) -> None:
-        mod = import_module(str(path))
-        self.with_module(mod)
+    def with_config(self, config: EnvironmentConfig) -> None:
+        for source in config.sources:
+            mod = import_module(str(source.module))
+            self.with_module(mod)
```
callumforrester, line 70: We want to be doing a different action based on the value of `source.type`, so something like:
```python
if source.type == "plans":
    self.with_plan_module(mod)
elif source.type == "dodal":
    self.with_device_module(mod)
else:
    ...
```

src/blueapi/core/context.py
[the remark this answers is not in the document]
abbiemery: I spoke to mx and agree we should leave the util in dodal.

abbiemery:
@callumforrester Done the last things you pointed out and rebased. The tests will still fail until the dodal updates are merged. I'm just waiting on a review for that.

src/blueapi/config.py
```diff
@@ -9,7 +10,18 @@
 LogLevel = Literal["NOTSET", "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
 
 
-class StompConfig(BlueapiBaseModel):
+class SourceKind(str, Enum):
+    planFunctions = "planFunctions"
+    deviceFunctions = "deviceFunctions"
+    dodal = "dodal"
```
callumforrester, line 16: Very minor, but enums count as constants under PEP 8 so names should in SCREAMING_SNAKE_CASE

```suggestion
    PLAN_FUNCTIONS = "planFunctions"
    DEVICE_FUNCTIONS = "deviceFunctions"
    DODAL = "dodal"
```

callumforrester (approved):
Provisionally approving pending dodal PR merge. Only minor comments now. Some docstrings on new classes/methods would also be good if you have time.

PR #173, opened 2023-05-05
Stop models converting to camelCase

callumforrester:
Previously with #132, all config, API parameter and plan model parameter names were converted to camelCase when loaded externally (e.g. from JSON). So 

```python
class TaskResponse(BlueapiBaseModel):
    """
    Acknowledgement that a task has started, includes its ID
    """

    task_name: str = Field(description="Unique identifier for the task")
```
Would accept the following via the API:
```json
{"taskName": "my task"}
```

and a plan like

```python
def my_plan(foo_bar: str) -> MsgGenerator:
    ...
```
Would accept:
```json
{"name": "my_plan", "params": {"fooBar": "qwerty"}}
```

This PR removes all aliasing and we accept that there are snake_case models

### DIFF
--- a/docs/developer/explanations/decisions/0003-api-case.rst
+++ b/docs/developer/explanations/decisions/0003-api-case.rst
@@ -1,0 +1,26 @@
+2. API Model Case
+=================
+
+Date: 2023-05-23
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+Considering whether keys in JSON blobs from the API should be in snake_case or camelCase.
+This includes plan parameters which may be user-defined.
+
+Decision
+--------
+
+The priority is not to confuse users, so we will not alias any field names defined in Python.
+
+Consequences
+------------
+
+Most code will be written with pep8 enforcers which means most field names will be snake_case.
+Some user defined ones may differ.
--- a/src/blueapi/utils/base_model.py
+++ b/src/blueapi/utils/base_model.py
@@ -17,7 +17,7 @@ class BlueapiModelConfig(BaseConfig):
     allow_population_by_field_name = True
 
 
-class BlueapiPlanModelConfig(BlueapiModelConfig):
+class BlueapiPlanModelConfig(BaseConfig):
     """
     Pydantic config for plan parameters.
     Includes arbitrary type config so that devices
@@ -27,6 +27,7 @@ class BlueapiPlanModelConfig(BlueapiModelConfig):
     from the context.
     """
 
+    extra = Extra.forbid
     arbitrary_types_allowed = True
     validate_all = True
 
--- a/src/blueapi/utils/base_model.py
+++ b/src/blueapi/utils/base_model.py
@@ -1,18 +1,12 @@
 from pydantic import BaseConfig, BaseModel, Extra
 
 
-def _to_camel(string: str) -> str:
-    words = string.split("_")
-    return words[0] + "".join(word.capitalize() for word in words[1:])
-
-
 class BlueapiModelConfig(BaseConfig):
     """
     Pydantic config for blueapi API models with
     common config.
     """
 
-    alias_generator = _to_camel
     extra = Extra.forbid
     allow_population_by_field_name = True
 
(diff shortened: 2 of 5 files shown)


tpoliaw:
I would be in favour of this. Having the request format be able to take multiple forms just add confusion. eg how would it handle requests for 
```python
def demo(foo_bar: int = 42, fooBar: str = "42") -> MsgGenerator:
    pass
```
It's a slightly contrived example but I don't see the benefit to allowing variations in general.

callumforrester:
If no one strongly objects I'm going to merge this, will make life easier for artemis purposes

tests/core/test_context.py
```diff
@@ -329,3 +329,15 @@ def test_nested_str_default(
     assert parse_obj_as(model, {}).m == [sim_motor]  # type: ignore
     empty_context.device(alt_motor)
     assert parse_obj_as(model, {"m": [ALT_MOTOR_NAME]}).m == [alt_motor]  # type: ignore
+
+
+def test_plan_models_not_auto_camelcased(empty_context: BlueskyContext) -> None:
```
tpoliaw, line 334: This test doesn't really test anything to do with camelCasing. It passes without this change as the casing stuff creates aliases.

src/blueapi/utils/base_model.py
```diff
@@ -17,7 +17,7 @@ class BlueapiModelConfig(BaseConfig):
     allow_population_by_field_name = True
 
 
-class BlueapiPlanModelConfig(BlueapiModelConfig):
+class BlueapiPlanModelConfig(BaseConfig):
```
tpoliaw, line 20: Is this class still needed? Could BlueapiModelConfig be used everywhere instead

tpoliaw (commented):
Should non-plan models still allow camelCase aliasing? Seems weird if it works in some places but not others.

tests/core/test_context.py
callumforrester: Good spot, have amended the test and checked it fails without the change.

src/blueapi/utils/base_model.py
callumforrester: True if we decide non-plan models shouldn't have camelcase either, as you suggest. I'm on the fence about it.

The reason behind the aliasing in the first place was because we're going to have multiple services in multiple languages so I think there should be _a_ standard among the various APIs. CamelCase seems to be a more common standard for REST etc. from some brief googling I did at the time. It also helps interoperation with Java services and GDA.

We can just say blueapi's API is all snake_case and be done with it, just expect other applications to deal with it, and accept a certain level of divergence. I guess we are already accepting that by making this change...

I'm not sure, any opinions?

tpoliaw: I'd err on the side of having the API be strict and having the clients adhere to it via whatever mechanisms the de/serialisation libraries offer. If camelCase is more common in rest APIs I wouldn't be as against forcing everything to use camelCase instead, but it'd make the python side more awkward.

[Only tangentially relevant but an interesting read](https://www.ietf.org/archive/id/draft-iab-protocol-maintenance-05.html)

callumforrester: The lack of strictness is definitely annoying, and wasn't a problem we had with apischema. I haven't found a way to overcome it with pydantic yet. Maybe it is just worth having an all snake_case API then...
Would also be interested in @joeshannon's thoughts.

tpoliaw: How did apischema handle it?

callumforrester: You can straight up convert rather than alias

tpoliaw: I think pydantic lets you do that if you set `allow_population_by_field_name` to `False` and then set the alias.

callumforrester: I'll take a look at that then, cheers. Although still open to discussion of whether we should be doing any camelCasing at all, may drag us together offline.

keithralphs: Hive mind says remove all aliasing

callumforrester: Done

docs/developer/explanations/decisions/0003-api-case.rst
```diff
@@ -0,0 +1,26 @@
+2. API Model Case
```
keithralphs, line 1: This 2 should be a 3; 2 is No queues

keithralphs (approved):

src/blueapi/utils/base_model.py
tpoliaw: Back to the original question, do we still need different models for plans vs non-plans?

PR #174, opened 2023-05-05
Remove public fields of BlueskyContext and replace with methods

callumforrester:
It seems more sustainable to use private fields and access them via methods. The device and plan dictionaries and run engine are now hidden behind methods.

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -49,10 +49,7 @@ class BlueskyContext:
     _reference_cache: Dict[Type, Type]
 
     def __init__(self, run_engine: Optional[RunEngine] = None) -> None:
-        if run_engine is None:
-            run_engine = RunEngine(context_managers=[])
-
-        self._run_engine = run_engine
+        self._run_engine = run_engine or RunEngine(context_managers=[])
         self._devices = {}
         self._plans = {}
         self._plan_functions = {}
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -1,5 +1,4 @@
 import logging
-from dataclasses import field
 from importlib import import_module
 from inspect import Parameter, signature
 from pathlib import Path
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -62,21 +62,73 @@ class BlueskyContext:
 
     @property
     def run_engine(self) -> RunEngine:
+        """
+        RunEngine capable of running the plans in the context against the devices in the context.
+
+        Returns:
+            RunEngine: A Bluesky RunEngine
+        """
+
         return self._run_engine
 
     def find_plan_function(self, name: str) -> Optional[PlanGenerator]:
+        """
+        Return a plan function matching the given name
+
+        Args:
+            name: The name of the plan
+
+        Returns:
+            Optional[PlanGenerator]: The plan function, None if there is
+                no plan matching the name in the context.
+        """
         return self._plan_functions.get(name)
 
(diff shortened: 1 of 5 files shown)


src/blueapi/core/context.py
```diff
@@ -37,21 +38,44 @@
...
+    _devices: Dict[str, Device]
+    _plan_functions: Dict[str, PlanGenerator]
 
     _reference_cache: Dict[Type, Type] = field(default_factory=dict)
 
+    def __init__(self, run_engine: Optional[RunEngine] = None) -> None:
+        if run_engine is None:
+            run_engine = RunEngine(context_managers=[])
+
+        self._run_engine = run_engine
```
tpoliaw, line 55: `self._run_engine = run_engine or RunEngine(context_managers=[])`?

src/blueapi/core/context.py
```diff
@@ -37,21 +38,44 @@
...
-    )
-    plans: Dict[str, Plan] = field(default_factory=dict)
-    devices: Dict[str, Device] = field(default_factory=dict)
-    plan_functions: Dict[str, PlanGenerator] = field(default_factory=dict)
+    _run_engine: RunEngine
+    _plans: Dict[str, Plan]
+    _devices: Dict[str, Device]
+    _plan_functions: Dict[str, PlanGenerator]
 
     _reference_cache: Dict[Type, Type] = field(default_factory=dict)
```
tpoliaw: Shouldn't this field be removed as well?

src/blueapi/core/context.py
```diff
@@ -37,21 +38,44 @@
...
+    def run_engine(self) -> RunEngine:
+        return self._run_engine
+
+    def find_plan_function(self, name: str) -> Optional[PlanGenerator]:
+        return self._plan_functions.get(name)
+
+    def find_plan_metadata(self, name: str) -> Optional[Plan]:
+        return self._plans.get(name)
+
+    def all_devices(self) -> Iterable[Device]:
```
tpoliaw, line 100: worth making these two properties as well?

src/blueapi/worker/task.py
```diff
@@ -41,15 +41,18 @@ class RunPlan(Task):
     def do_task(self, ctx: BlueskyContext) -> None:
         LOGGER.info(f"Asked to run plan {self.name} with {self.params}")
 
-        plan = ctx.plans[self.name]
-        func = ctx.plan_functions[self.name]
-        sanitized_params = _lookup_params(ctx, plan, self.params)
-        plan_generator = func(**sanitized_params.dict())
+        metadata = ctx.find_plan_metadata(self.name)
```
tpoliaw, line 44: If these two functions are only ever called together and both have to be present, it'd be clearer to have a single `find_plan(name) -> Optional[(Plan, PlanFunc)]`. Probably suggests `plans` and `plan_functions` should be merged into a single field in the context as well. Maybe a dict of `name -> (Plan, func)`?

tpoliaw (commented):
There are a few docs changes here not related to replacing fields with methods. Is it picky to suggest these are committed/merged as separate commits when the PR is merged?

src/blueapi/core/context.py
callumforrester: Why? It's in use and already private

tpoliaw: I meant the `= field(default_factory=dict)` part, but it looks like you changed it between me writing and submitting the comment

callumforrester:
Very out of date, we may revisit what to do about `BlueskyContext` in future, it has become quite large and bloated... https://github.com/DiamondLightSource/blueapi/issues/616

PR #175, opened 2023-05-11
Allow injection of Devices into plan arguments from the BlueskyContext

### LINKED ISSUE #116: Enable injection of device from context into plans
Some plans will need to access the available plans and devices in the worker while running. Example design:

```python
@blueapi.pass_context
def my_plan(ctx: BlueskyContext):
    det = ctx.devices["xpress3"]
    yield from scan([det], ...)
```

DiamondJoseph:
closes #116


DiamondJoseph:
Probably wants some documentation too

src/blueapi/core/context.py
```diff
@@ -181,6 +217,11 @@ def __get_validators__(cls):
 
                 @classmethod
                 def valid(cls, value):
+                    if target is PlanGenerator:
+                        val = self.plan_functions[value]
+                        if val is target:
+                            return val
+                        raise ValueError(f"value is not {target}")
```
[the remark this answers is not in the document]
DiamondJoseph: Ticket says "injection of plan/device", this (or the most recent version of it) is required to support plans being injected

tests/core/test_context.py
```diff
@@ -214,3 +223,17 @@ def demo(named: Named) -> None:
 
     spec = empty_context._type_spec_for_function(demo)
     assert spec == {"named": (hasname_ref, None)}
+
+
+def test_str_default(empty_context: BlueskyContext, sim_motor: SynAxis):
+    movable_ref = empty_context._reference(Movable)
+
+    spec = empty_context._type_spec_for_function(has_default_str)
+    assert spec == {"m": (movable_ref, SIM_MOTOR_NAME)}
```
tpoliaw, line 232: These are going to fail if/when #185 gets merged

src/blueapi/startup/example.py
```diff
@@ -36,3 +40,35 @@
     Imax=1,
     labels={"detectors"},
 )
+
+T = TypeVar("T")
+
+
+def ref(t: T) -> T:
```
tpoliaw, line 47: Is this ever used?

tests/core/test_context.py
callumforrester: Since #185 is now approved, shall we merge it and resolve the conflicts here?

DiamondJoseph: :+1:  from my side

tpoliaw: Now merged

src/blueapi/startup/example_plans.py
```diff
@@ -0,0 +1,30 @@
+from typing import List
+
+from bluesky.protocols import Movable, Readable
+
+from blueapi.plans import count, move
+
+from ..core import MsgGenerator, inject
```
callumforrester, line 7: If it's in a different package it's better to use the canonical path

```suggestion
from blueapi.core import MsgGenerator, inject
```

src/blueapi/core/__init__.py
```diff
@@ -25,6 +26,7 @@
     "EventStream",
     "DataEvent",
     "WatchableStatus",
+    "inject",
```
callumforrester, line 29: Should we call it `inject_device`, to be clearer?

callumforrester (approved):
Looks good, only minor nitpicks

callumforrester (approved):

src/blueapi/core/device_lookup.py
```diff
@@ -48,3 +48,7 @@ def find_component(obj: Any, addr: List[str]) -> Optional[D]:
             f"Found {component} in {obj} while searching for {addr} "
             "but it is not a device"
         )
+
+
+def inject(x):
```
tpoliaw, line 53: Worth commenting that types are intentionally omitted to bypass type checking? This function doesn't look like it does much otherwise. I think `x` could be typed as `str` as well.

tests/core/test_context.py
```diff
@@ -274,3 +293,39 @@ def demo(named: Named) -> None:
...
+def test_str_default(
+    empty_context: BlueskyContext, sim_motor: SynAxis, alt_motor: SynAxis
+):
+    movable_ref = empty_context._reference(Movable)
+    empty_context.device(sim_motor)
+    empty_context.plan(has_default_reference)
+
+    spec = empty_context._type_spec_for_function(has_default_reference)
+    assert spec["m"][0] is movable_ref
+    assert spec["m"][1].default_factory() is SIM_MOTOR_NAME
```
tpoliaw, line 307: `==` instead of `is`? Does the instance matter?

tpoliaw (approved):

tests/core/test_context.py
DiamondJoseph: Done

src/blueapi/core/device_lookup.py
DiamondJoseph: Have added some doc string and if the linting fails I can revert the name: str

PR #179, opened 2023-05-11
Validate

keithralphs:
Adds Ci to validate AsyncAPI def

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -1,0 +1,19 @@
+# Validates the AsyncAPI definition
+
+name: asyncapi_validate
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  validate:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


DiamondJoseph (approved):

.github/workflows/validate.yml
```diff
@@ -0,0 +1,19 @@
+# Validates the AsyncAPI definition
+
+name: asyncapi_validate
```
callumforrester, line 3: Could we change the case to match all the other CI jobs?

image

```suggestion
name: AsyncAPI Validate
```

docs/user/reference/asyncapi.yaml
```diff
@@ -2,7 +2,7 @@ asyncapi: "2.6.0"
 id: https://github.com/DiamondLightSource/blueapi
 info:
   title: Bluesky Worker Service
-  version: <main>
+  version: 0.0.1
```
callumforrester, line 5: This change is already in main, may need a rebase?

callumforrester (changes requested):

callumforrester (approved):

PR #185, opened 2023-05-12
Fix handling of default parameters in plans

tpoliaw:
Passing `None` as the default for a field sets the default to None and
is different to there being no default.

If a default value is given, it is now returned via a default_factory.
This is required to prevent pydantic errors when the default is an ophyd
device.

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -20,6 +20,7 @@ from typing import (
 
 from bluesky import RunEngine
 from pydantic import create_model
+from pydantic.fields import FieldInfo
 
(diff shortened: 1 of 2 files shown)


DiamondJoseph (approved):

callumforrester (changes requested):
Makes sense, not sure why CI is unhappy. Could we also have a test for this use case?

tpoliaw:
> not sure why CI is unhappy

Because `lambda x: x` != `lambda x: x`

callumforrester (approved):

Issue #177, opened 2023-05-11
Add correlation ID handling to rest endpoints.


rosesyrett:
At the moment the rest api does not handle correlation ID's, but it needs to.
It seems like the client of the API should generate a correlation ID, and call the rest API endpoints with it. The rest endpoints should do things with it (e.g. submitting a task) and return it back as confirmation that stuff has been done.

callumforrester:
@keithralphs does this relate to your opentelemetry work?

keithralphs:
Yes, this is one of my ideas for this using the baggage to pass the id , however it may be possible to accomplish automatically via context propagation if all clients are OT instrumented when the make the call as the default is for the combintaion of the current Span plus the trace id to form the correlation id. In this scheme the Trace id would identify the client and the span id , the call it made.

callumforrester:
Fair enough, if that works we can close this PR I guess. Do you have that in a draft anywhere?

keithralphs:
Not as such. I have the server side example at https://github.com/DiamondLightSource/blueapi/pull/355, but it;'s not really properly addressing correlation id throughout at the moment.

callumforrester:
Now done by @keithralphs tracing changes

PR #186, opened 2023-05-12
fix json parameterization bug

RAYemelyanova:
fixed a bug in the json parametarization for running plans, so plans can actually be run


DiamondJoseph (approved):

RAYemelyanova:
Going to follow this up with a couple of PRs; one to deal with the fact the tests fail if the server is running, and the other dealing with correlation IDs https://github.com/DiamondLightSource/blueapi/issues/177 which would involve changing the cli command so it doesn't automatically return on submission and can return error codes from the worker.

PR #188, opened 2023-05-12
Make more dependencies optional

callumforrester:
Some use cases involve using blueapi as a library. In these cases, users may not want to depend on I/O libraries such as fastapi and stomp.
This change separates those into optional dependencies so that the full suite of functionality can only be achieved via a pip install blueapi[all].


callumforrester:
Will break the container

Issue #187, opened 2023-05-12
tests fail if the server is running


RAYemelyanova:
if you spin up the blueapi server and try to run the tests they will fail. This shouldn't be the case.

PR #189, opened 2023-05-12
Made tests work with server running

RAYemelyanova:
This was a simple bug; one of the tests in test_cli.py assumed the server isn't running, which means if it is, it fails.

I've mocked out the api call this makes now. So now the tests work with the server running.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -48,7 +48,7 @@ def test_main_with_nonexistent_config_file():
 
 
 @patch("requests.get")
-def test_controller_plans(mock_requests: Mock):
+def test_connection_error_caught_by_wrapper_func(mock_requests: Mock):
     mock_requests.side_effect = ConnectionError()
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -39,7 +40,9 @@ def test_main_with_nonexistent_config_file():
     type(result.exception) == FileNotFoundError
 
 
-def test_controller_plans():
+@patch("requests.get")
+def test_controller_plans(mock_requests: Mock):
```
tpoliaw, line 44: should this test be renamed? It's more testing the failure case than the controller plans method/endpoint

RAYemelyanova: Sure, good idea. Have renamed this now.

DiamondJoseph (approved):

RAYemelyanova:
solves https://github.com/DiamondLightSource/blueapi/issues/187

Issue #181, opened 2023-05-12
Unify markup language for API reference


callumforrester:
AsyncAPI spec is currently in YAML, OpenAPI one is in JSON. They should all be in YAML because readability.

PR #190, opened 2023-05-12
Unifying markup languages for api schemas

RAYemelyanova:
renamed some docs files, added more detail about apis. Made openapi.py output yaml instead of json

resolves https://github.com/DiamondLightSource/blueapi/issues/181

Also aims to be more descriptive in the docs about the purposes of the different api schemas we have


docs/user/reference/openapi.yaml
```diff
@@ -0,0 +1,217 @@
...
+          title: Error Type
+          type: string
+      required:
+      - loc
+      - msg
+      - type
+      title: ValidationError
+      type: object
+info:
+  title: FastAPI
```
keithralphs, line 105: This should have a better Title;  I would suggest just 'BlueAPI' as the fact that it's and OpenAPI spec tells us it's REST already

docs/user/reference/rest-spec.rst
```diff
@@ -0,0 +1,8 @@
+Rest Specification
```
keithralphs, line 1: Change Rest to REST as it's an acronym

keithralphs (changes requested):

docs/user/reference/openapi.yaml
RAYemelyanova: Have updated.

Issue #176, opened 2023-05-11
Combine all events emitted by bluesky into one topic


DiamondJoseph:
Bluesky worker currently emits on `public.worker.event.data` the Bluesky Event Model Documents, on `public.worker.event.progress` events describing the status of an ongoing plan and on `public.worker.event` events describing the state of the run engine and the plan it is currently running. 

In order to guarantee that the order of these is preserved, they should be on the same topic: it should also be the case that the `public.worker.event` events for the start of a plan being run should arrive before any StartDocuments from Runs within its plan, and the event for the end of a plan being run should arrive after any StopDocuments from runs within its plan.

PR #195, opened 2023-05-15
fix topic publishing

rosesyrett:
copied code from before fastAPI change in service/app.py to re-establish the same subscriptions on the worker


RAYemelyanova:
solves https://github.com/DiamondLightSource/blueapi/issues/196

I feel like we need a test case for this. But maybe this is overkill considering active-mq topics are going to change anyways due to https://github.com/DiamondLightSource/blueapi/issues/176

joeshannon (approved):
Thanks, this fixes it!

PR #197, opened 2023-05-15
Correlation id

rosesyrett:
Attempts to deal with correlation id via CLI and rest api.

The concept is, whatever is calling the rest API will already have some correlation id for the plan to be run. 
This PR should allow the client to call the server via the CLI to:
1. put a plan onto the queue (with some correlation id). This will return immediately.
2. Check for the status of a plan with some correlation id.

I'm not yet sure the best way of restfully accomplishing these tasks. So far, I've just done everything through the CLI run_plan function, but this handles the active mq subscriptions directly which doesn't seem smart. It seems better to have some sort of 'cache' of bluesky events that can be accessed by the rest API to get the latest status.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
             "request": "launch",
             "module": "uvicorn",
             "args": [
-                "src.blueapi.rest.main:app",
+                "src.blueapi.service.main:app",
                 "--reload"
             ],
             "jinja": true,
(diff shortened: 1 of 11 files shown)


callumforrester:
Worker transactions now in #202

callumforrester:
@keithralphs does this PR contain anything useful to you in your opentelemetery endeavours?

callumforrester:
This is now very out of date, we can decide on a new way forward based on #586

PR #202, opened 2023-05-15
Add transaction mode to worker

callumforrester:
Draft PR with a suggested transaction mode for the worker, transaction mode enables creating of new tasks before they are submitted.

To consider:
* Is the transaction API right?
* Should we allow multiple pending transactions or only one at a time? Currently we lock the transaction mode if one is in progress.
* Can the implementation be simplified? (probably yes)
* Should we still allow non-transaction plan running? (probably not, as it provides a way around the aforementioned transaction lockout)

### DIFF
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -52,7 +52,7 @@ class RunEngineWorker(Worker[Task]):
     _state: WorkerState
     _errors: List[str]
     _warnings: List[str]
-    _task_queue: Queue  # type: ignore
+    _task_channel: Queue  # type: ignore
     _current: Optional[TrackableTask]
     _status_lock: RLock
     _status_snapshot: Dict[str, StatusView]
@@ -77,7 +77,7 @@ class RunEngineWorker(Worker[Task]):
         self._state = WorkerState.from_bluesky_state(ctx.run_engine.state)
         self._errors = []
         self._warnings = []
-        self._task_queue = Queue(maxsize=1)
+        self._task_channel = Queue(maxsize=1)
         self._current = None
         self._worker_events = EventPublisher()
         self._progress_events = EventPublisher()
@@ -115,7 +115,7 @@ class RunEngineWorker(Worker[Task]):
     def _submit_trackable_task(self, trackable_task: TrackableTask) -> None:
         LOGGER.info(f"Submitting: {trackable_task}")
         try:
-            self._task_queue.put_nowait(trackable_task)
+            self._task_channel.put_nowait(trackable_task)
         except Full:
             LOGGER.error("Cannot submit task while another is running")
             raise WorkerBusyError("Cannot submit task while another is running")
@@ -130,7 +130,7 @@ class RunEngineWorker(Worker[Task]):
 
         # If the worker has not yet started there is nothing to do.
         if self._started.is_set():
-            self._task_queue.put(KillSignal())
+            self._task_channel.put(KillSignal())
             self._stopped.wait(timeout=self._stop_timeout)
             # Event timeouts do not actually raise errors
             if not self._stopped.is_set():
@@ -164,7 +164,7 @@ class RunEngineWorker(Worker[Task]):
     def _cycle(self) -> None:
         try:
             LOGGER.info("Awaiting task")
-            next_task: Union[TrackableTask, KillSignal] = self._task_queue.get()
+            next_task: Union[TrackableTask, KillSignal] = self._task_channel.get()
             if isinstance(next_task, TrackableTask):
                 LOGGER.info(f"Got new task: {next_task}")
                 self._current = next_task  # Informing mypy that the task is not None
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -47,6 +47,9 @@ class RunEngineWorker(Worker[Task]):
     _ctx: BlueskyContext
     _stop_timeout: float
 
+    _transaction_lock: RLock
+    _pending_transaction: Optional[ActiveTask]
+
     _state: WorkerState
     _errors: List[str]
     _warnings: List[str]
@@ -70,6 +73,9 @@ class RunEngineWorker(Worker[Task]):
         self._ctx = ctx
         self._stop_timeout = stop_timeout
 
+        self._transaction_lock = RLock()
+        self._pending_transaction = None
+
         self._state = WorkerState.from_bluesky_state(ctx.run_engine.state)
         self._errors = []
         self._warnings = []
@@ -85,8 +91,49 @@ class RunEngineWorker(Worker[Task]):
         self._stopping = Event()
         self._stopped = Event()
 
-    def submit_task(self, name: str, task: Task) -> None:
-        active_task = ActiveTask(name, task)
+    def begin_transaction(self, task: Task) -> str:
+        task_id: str = str(uuid.uuid4())
+        with self._transaction_lock:
+            if self._pending_transaction is not None:
+                raise WorkerBusyError("There is already a transaction in progress")
+            self._pending_transaction = ActiveTask(task_id, task)
+        return task_id
+
+    def clear_transaction(self) -> str:
+        with self._transaction_lock:
+            if self._pending_transaction is None:
+                raise Exception("No transaction to clear")
+
+            task_id = self._pending_transaction.task_id
+            self._pending_transaction = None
+            return task_id
+
+    def commit_transaction(self, task_id: str) -> None:
+        with self._transaction_lock:
+            if self._pending_transaction is None:
+                raise Exception("No transaction to commit")
+
+            pending_id = self._pending_transaction.task_id
+            if pending_id == task_id:
+                self._submit_active_task(self._pending_transaction)
+            else:
+                raise KeyError(
+                    "Not committing the transaction requested, asked to commit"
+                    f"{task_id} when {pending_id} is in progress"
+                )
+
+    def get_pending(self) -> Optional[Task]:
+        with self._transaction_lock:
+            if self._pending_transaction is None:
+                return None
+            else:
+                return self._pending_transaction.task
+
+    def submit_task(self, task_id: str, task: Task) -> None:
+        active_task = ActiveTask(task_id, task)
+        self._submit_active_task(active_task)
+
+    def _submit_active_task(self, active_task: ActiveTask) -> None:
         LOGGER.info(f"Submitting: {active_task}")
         try:
             self._task_queue.put_nowait(active_task)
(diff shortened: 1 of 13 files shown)


tpoliaw:
* Is the transaction API right?
The general approach of, "submit task -> get ID", "Start task by ID" is ok. I think the method names should reflect the domain rather than the pattern they're following. `submit_task(t: Task) -> TaskID:` , `run_task(id: TaskID) -> Status`, `drop_task(id: TaskID) -> Status`, `pending_tasks() -> list[TaskID]`. Should possibly be some way of querying what a pending task is as well to get the original `Task` back.
It  would be good if submitting tasks also validated them so users/clients get feedback that the task they're running is valid or not. A task being accepted but invalid doesn't make much sense.
* Should we allow multiple pending transactions or only one at a time?
Multiple probably makes most sense, otherwise you could be blocked by a pending but ignored request. If each submitted task has a timestamp, there could be some expiration that prevents the pending tasks piling up.
* Can the implementation be simplified?
Always but it's a good starting point to build on. It might end up needing to be refactored anyway to allow handling of the expiration/validation to be added.
* Should we still allow non-transaction plan running?
No

callumforrester:
@tpoliaw API design makes sense, have implemented. The validation on submission sounds very sensible and has also been identified as a requirement for Artemis. I think that should go in a separate PR to avoid bloating this one.

More stream-of-conciousness stuff: Generally agree with everything else although I'm a bit concerned about the expiry time solution, since it may be indeterministic and hard to test. Also, is it even blueapi's job to manage the state of everyone's potential tasks? One thought that did occur to me would be to have a stack of tasks, which would be useful for things like pausing/resuming but I haven't thought it out properly yet. Any thoughts welcome.

tpoliaw:
For the expiration side, would an `OrderedDict[id, (task, timestamp)]` work? Each task that's submitted would be added to the dict. If `run_task` is called for a task that is older than x seconds, it fails. If there are more than y tasks in the queue when a new one is added, remove any expired ones. As the dict is ordered, this would be a single check in the usual case. It would be deterministic but the x/y parameters would be kept as an implementation detail and only the timestamp is actually part of the API. It could be returned with the taskID?

If only one task is running at once, a stack probably isn't needed although the current task would need to be available somewhere.

src/blueapi/worker/reworker.py
```diff
@@ -83,8 +87,26 @@ def __init__(
...
+    def begin_task(self, task_id: str) -> None:
+        task = self._pending_tasks.get(task_id)
+        if task is not None:
+            self._submit_active_task(task)
+        else:
+            raise KeyError(f"No pending task with ID {task_id}")
+
+    def submit_task(self, task: Task) -> str:
+        task_id: str = str(uuid.uuid4())
+        active_task = ActiveTask(task_id, task)
```
tpoliaw, line 105: `ActiveTask` should probably be called something else if it's now being used for non-active tasks

src/blueapi/worker/reworker.py
```diff
@@ -193,11 +215,11 @@ def _report_status(
         warnings = self._warnings
         if self._current is not None:
             task_status = TaskStatus(
-                task_name=self._current.name,
+                task_name=self._current.task_id,
```
tpoliaw, line 218: Is there a way to avoid having `task_name` and `task_id` as two names for the same thing?

src/blueapi/worker/reworker.py
```diff
@@ -193,11 +215,11 @@ def _report_status(
...
                 task_complete=self._current.is_complete,
                 task_failed=self._current.is_error or bool(errors),
             )
-            correlation_id = self._current.name
+            correlation_id = self._current.task_id
```
tpoliaw, line 222: Wasn't the correlationId going to be different to the taskId? Doesn't it have to be provided by the client?

src/blueapi/worker/worker.py
```diff
@@ -15,13 +15,38 @@ class Worker(ABC, Generic[T]):
     """
 
     @abstractmethod
-    def submit_task(self, __name: str, __task: T) -> None:
+    def get_pending_tasks(self) -> List[T]:
+        """_summary_
```
tpoliaw, line 19: \_comment\_

src/blueapi/worker/worker.py
```diff
@@ -15,13 +15,38 @@ class Worker(ABC, Generic[T]):
...
+        """
+
+    @abstractmethod
+    def submit_task(self, task: T) -> str:
         """
         Submit a task to be run
 
         Args:
-            __name (str): A unique name to identify this task
+            __name (str): name of the plan to be run
```
tpoliaw, line 47: These are no longer the args to this method

tests/worker/test_reworker.py
```diff
@@ -17,85 +19,158 @@
...
+    worker.clear_task(task_id)
+    assert worker.get_pending_tasks() == []
+
+
+def test_clear_nonexistant_task(worker: Worker) -> None:
+    with pytest.raises(KeyError):
+        worker.clear_task("foo")
+
+
+@pytest.mark.parametrize("num_runs", [2, 3, 4])
```
tpoliaw, line 109: What is gained from running this multiple numbers of tasks? Won't they all fail on the second one?

src/blueapi/worker/reworker.py
```diff
@@ -83,8 +87,26 @@ def __init__(
         self._stopping = Event()
         self._stopped = Event()
 
-    def submit_task(self, name: str, task: Task) -> None:
-        active_task = ActiveTask(name, task)
+    def clear_task(self, task_id: str) -> None:
+        del self._pending_tasks[task_id]
```
tpoliaw, line 91: Should this raise an error if the task doesn't exist? If it matters that the task doesn't exist, it could return a true/false value.

src/blueapi/worker/reworker.py
```diff
@@ -83,8 +87,26 @@ def __init__(
...
+
+    def get_pending_tasks(self) -> List[Task]:
```
tpoliaw, line 93: Should this returns the tasks or the IDs? What is this method intended to be used for?

src/blueapi/worker/reworker.py
callumforrester: I think it does matter since this will eventually be propagated up to the REST API and we want a DELETE operation to fail if the resource doesn't exist. Whether that should be an exception or return value, I have no strong feelings in this case. Have changed to return True or False.

src/blueapi/worker/reworker.py
callumforrester: Intended use is the REST API so it should return the tasks

src/blueapi/worker/reworker.py
callumforrester: Yes, was being conservative because that would change the external API, but we currently  have no users so it's fine...

src/blueapi/worker/reworker.py
callumforrester: Yes, but we're not doing anything to do with the correlation ID in this change, so preserving the old behaviour (where it is the same) for now.

tests/worker/test_reworker.py
callumforrester: This was a holdover from when I was experimenting with stacks, good spot, have changed.

callumforrester (commented):
Cheers, all good spots. I'm a bit more convinced by the expiry time argument now, but want to do it as a separate PR, possibly along with the REST stuff, to avoid bloating this one any further. 
I'm also considering deferring it since we can reasonably expect our own client implementation in GDA to do its own bookkeeping. Maybe we should wait until we have other clients that don't?

src/blueapi/worker/reworker.py
```diff
@@ -70,6 +72,8 @@ def __init__(
         self._ctx = ctx
         self._stop_timeout = stop_timeout
 
+        self._pending_tasks = {}
```
DiamondJoseph, line 75: Are we going back on Blueapi not having more than 1 task queued?

src/blueapi/worker/task.py
```diff
@@ -68,8 +68,13 @@ def _lookup_params(
 
 
 @dataclass
-class ActiveTask:
-    name: str
+class TrackableTask:
+    """
+    An internal representation for use by the worker, holding a
+    task with it's ID and current status
```
DiamondJoseph, line 74: If we're saying that a Task's status is either complete (or not) and in error (or not), that agrees with my thoughts on Pause/Resume being operations on the RunEngine and not of the Task.

src/blueapi/worker/reworker.py
callumforrester: Realised we would also need the task ID for the rest API so I promoted `TrackableTask` (formerly `ActiveTask`) to the API level

src/blueapi/worker/reworker.py
callumforrester: Done

src/blueapi/worker/reworker.py
callumforrester: @keithralphs agreed but still don't want to bloat this change any further. Linking to #177 for reference so this gets done

src/blueapi/worker/worker.py
```diff
@@ -1,27 +1,73 @@
...
 
+class TrackableTask(BlueapiBaseModel, Generic[T]):
+    """
+    A representation of a task that the worker recognizes
+    """
+
+    task_id: str
+    task: T
+    is_complete: bool = False
+    is_error: bool = False
```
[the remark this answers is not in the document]
callumforrester: I moved it here because it has become part of the `Worker` API, rather than just for internal use by `RunEngineWorker`. So one of the API methods on `Worker` returns a `TrackableTask`, which means things outside need to know about `TrackableTask`. Ideally, because it's an interface, everything in `worker.py` should not know about anything else, other things should just know about it.

tests/worker/test_reworker.py
```diff
@@ -7,106 +8,236 @@
...
+) -> None:
+    assert begin_task_and_wait_until_complete(worker, task_id) == expected_events
 
 
-def submit_task_and_wait_until_complete(
-    worker: Worker, task: Task, timeout: float = 5.0
+def begin_task_and_wait_until_complete(
+    worker: Worker,
+    task_id: str,
+    timeout: float = 5.0,
```
DiamondJoseph, line 227: We're going to continue adding more running of plans into tests. Some of those plans/tests will involve waiting, sleeping or checking time to know that a wait or a sleep has been done. In order to prevent this from inflating the time it takes to run our tests, we should use this early opportunity to make tests that mock out time and sleep to have simple examples to point to and to build off of. This also means that our tests will be more robust against external slowness of compute eating up fractions of our timeouts.

[Something like this](https://stackoverflow.com/questions/57747998/how-to-make-unit-test-with-asyncio-sleep-contained-code)

DiamondJoseph (changes requested):

tests/worker/test_reworker.py
RAYemelyanova: I feel this is a scope creep issue; I agree it is a good idea to do this, we should make a separate PR.

keithralphs: I agree: creep bad!

DiamondJoseph (approved):

Issue #111, opened 2023-04-05
Old statuses are kept unnecessarily


callumforrester:
When producing `TaskEvent`s. The worker includes statuses for completed operations e.g. moving a motor between two previous points in a scan. These are not needed and significantly bloat the size of the event, causing performance issues in client software.

callumforrester:
@keithralphs The bookkeeping is done, apparently wrong, in this class: https://github.com/DiamondLightSource/blueapi/blob/main/src/blueapi/worker/reworker.py#L194

joeshannon:
Fixed with #203

PR #203, opened 2023-05-15
Track completed statuses (#111)

joeshannon:
For some devices there may be an additional status update broadcast after the on_complete callback. This causes the status to be added back to the dictionary where it will never be removed.

Track which statuses have been completed and check this set when updating to ensure that a completed status is never added back to the snapshot after on_complete is called.

### DIFF
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass
 from functools import partial
 from queue import Full, Queue
 from threading import Event, RLock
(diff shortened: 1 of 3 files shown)


joeshannon:
The code coverage check does highlight that there aren't currently any tests for this.

callumforrester (approved):

DiamondJoseph:
What is a status_name that we are tracking, the uuid of the status?

DiamondJoseph:
> rename to something like `status_uuid`?

Sounds good to me.

DiamondJoseph (approved):

PR #204, opened 2023-05-16
Update adsim to functional format

### LINKED ISSUE #201: Change ADSim and simmotor startup scrips to the new dodal format
I accidentally overlooked changing both of these start up scripts in my dodal PR. They need to be changed to match the new format.

abbiemery:
Fixes #201


callumforrester (approved):
Could also switch to using the adsim module in dodal now?

abbiemery:
> Could also switch to using the adsim module in dodal now?

So just remove this script entirely and just change the yaml to point to doal?

abbiemery:
@callumforrester I've now removed it and just changed the yaml to use dodal.

PR #207, opened 2023-05-16
Make run plan parameters an argument

tpoliaw:
Instead of an option that is required in every call. If no parameters are passed,
default to empty dict.


joeshannon (approved):
It looks like a sensible change to me.
Do we want to check if others agree?

callumforrester (approved):

PR #209, opened 2023-05-17
Remove deprecated command

keithralphs:


callumforrester (approved):
Looks good from my point of view, would also like @RAYemelyanova to have a quick look.

RAYemelyanova (approved):
looks good, thanks

PR #210, opened 2023-05-17
Standardise names and remove publish elements

keithralphs:
Update Async API spec to remove so it only describes events all emitted on public.bluesky.event and update catalog info to rename component and API titles according tp scheme agreed with Garry.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -26,7 +26,7 @@ spec:
   lifecycle: production
   owner: user:vid18871
   definition:
-    $text: ./docs/user/reference/openapi.yaml
+    $text: ./docs/user/reference/openapi.json
 ---
 apiVersion: backstage.io/v1alpha1
 kind: API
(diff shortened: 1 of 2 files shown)


catalog-info.yaml
```diff
@@ -12,29 +12,28 @@ spec:
...
+  name: blueapiControl
+  title: Athena BlueAPI Control
   description: REST API for getting plans/devices from the worker (and running tasks)
 spec:
   type: openapi
   lifecycle: production
   owner: user:vid18871
   definition:
-    $text: ./docs/user/reference/openapi.json
+    $text: ./docs/user/reference/openapi.yaml
```
DiamondJoseph, line 29: openapi is still .json until #190 goes in, and this change should be in there

DiamondJoseph (changes requested):
Just a tiny nitpick

DiamondJoseph (approved):

Issue #198, opened 2023-05-15
Ensure worker/context tests to same degree as artemis 


callumforrester:
Look at tests in artemis for `BlueskyRunner` class and ensure we test the analogue functionality

DiamondJoseph:
https://github.com/DiamondLightSource/python-artemis/blob/main/src/artemis/system_tests/test_main_system.py


test_start_gives_success: // Return the status of a plan after running it blocking #199
test_getting_status_return_idle: // Get the status of the worker # Done
test_getting_status_after_start_sent_returns_busy: // Get the status of the worker which updates when RE running # Done
test_putting_bad_plan_fails: // Plan not found returns error # Done
test_plan_with_no_params_fails: // Plan run with missing or bad params returns error # Done
test_plan_with_no_run_fails: // Plan with no run method fails = Plans must be PlanGenerator # Done
test_sending_start_twice_fails: // Cannot run plan while plan is running # Done
test_given_started_when_stopped_then_success_and_idle_status: // Can start and stop worker # Not required
test_given_started_when_stopped_and_started_again_then_runs: // Can restart worker # Not required


test_given_started_when_RE_stops_on_its_own_with_error_then_error_reported: // #199
test_when_started_n_returnstatus_interrupted_bc_RE_aborted_thn_error_reptd: // #199
test_given_started_when_RE_stops_on_its_own_happily_then_no_error_reported: // Runs plans to completion then state is idle # Done


test_start_with_json_file_gives_success: // Test config # Done


test_cli_args_parse: // Test CLI parsing. Ignore? #256

test_when_plan_started_then_callbacks_created: // Allow registering callbacks # Currently: Leaving in Artemis' remit #257

test_when_blueskyrunner_initiated_then_plans_are_setup_and_devices_connected: // Behaviour changed, see https://github.com/DiamondLightSource/dodal/pull/62, #219
test_when_blueskyrunner_initiated_and_skip_flag_is_set_then_plans_are_setup_and_devices_are_not_connected: # See above
test_when_blueskyrunner_initiated_and_skip_flag_is_set_then_setup_called_upon_start: // Allow skipping setup for individual plans. Plan setups behaviour removed. # See above
test_when_blueskyrunner_initiated_and_skip_flag_is_not_set_then_all_plans_setup: // Test setup of plans is run. Behaviour removed, change to set up all devices on start in progress


test_log_on_invalid_json_params(caplog, test_env: ClientAndRunEngine): // Assert logs # Not required

PR #219, opened 2023-05-19
Consume information about failed device instantiation and expose information to REST API

DiamondJoseph:
Requires https://github.com/DiamondLightSource/dodal/pull/62 to be merged.

Part of #198, as allows start up when not all devices are connected (desired behaviour replacing plan setups)


stan-dot:
superceded by #515 , petition to close @callumforrester

callumforrester:
How does #515 cover this?

callumforrester:
@DiamondJoseph I think we should close this since its very out of date. If we want to expose this via the API we should make another issue for that?

Issue #217, opened 2023-05-19
Generate openapi schema and commit it if the API is changed by a commit


DiamondJoseph:

RAYemelyanova:
Upon further discussion have decided this should be dropped

PR #221, opened 2023-05-22
CI job to push openapi changes back to repo

RAYemelyanova:
changed gh actions to autogenerate openapi schema and push it to the repo

solves https://github.com/DiamondLightSource/blueapi/issues/217

### DIFF
--- a/.github/workflows/openapi.yml
+++ b/.github/workflows/openapi.yml
@@ -3,7 +3,10 @@ name: OpenAPI
 on:
   push:
   pull_request:
-  
+
+permissions:
+  contents: write
+
 env:
   # The target python version, which must match the Dockerfile version
   CONTAINER_PYTHON: "3.11"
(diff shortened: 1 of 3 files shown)


RAYemelyanova:
Currently I've added this step to the openapi.yml workflow:

```
- name: Publish schema to repo
  run: |
    git config --global user.name 'Github CI'
    git config --global user.email 'github-ci@users.noreply.github.com'
    git commit -am "Updated openapi schema"
    git push
```

this looks like the following in github:
Screenshot from 2023-05-22 09-43-41


However, as you can see above, this makes an extra commit every time the openapi yaml changes. Maybe we don't want this... in that case, I can change the above to:

```
- name: Publish schema to repo
  run: |
    git config --global user.name 'Github CI'
    git config --global user.email 'github-ci@users.noreply.github.com'
    git add .
    git commit --amend --no-edit
    git push
```

Except obviously this is re-writing history. This means for each commit we do, it will be amended by the CI. I don't know if we want this, or if it's cleaner to have the CI handle this...

I can also maybe change this step so it only commits if it notices a difference between files, at the moment every time we push a commit to main this job will run and that can be quite spammy in commit history logging...

PR #224, opened 2023-05-22
Combine events

### LINKED ISSUE #176: Combine all events emitted by bluesky into one topic

abbiemery:
Fixes #176 

I would really appreciate someone looking over the test in `test_reworker` for the ordering of `WorkerEvents` and `DataEvents` and the ordering of the types of `DataEvents`. I got quite caught up in the type system here. It appears to be beyond the reach of pythons typing to accurately type this, as backed up by @WareJosephB , however I could be wrong. I tried to explain the issue in a docstring of the offending function but again, could easily have tripped up my explanation here.


src/blueapi/cli/amq.py
```diff
@@ -51,12 +51,13 @@ def on_progress_event_wrapper(
                 on_progress_event(event)
 
         self.app.subscribe(
-            self.app.destinations.topic("public.worker.event"), on_event_wrapper
+            self.app.destinations.topic("public.worker.events"), on_event_wrapper
```
keithralphs, line 54: We say: How come it's changed to events from event - the convention for topics/queues is singular (we believe)

abbiemery: I wasn't aware of the convention, I'll change it back no problem

joeshannon (approved):
@abbiemery are you happy with the changes Peter made, and for this to be merged now?

PR #225, opened 2023-05-23
Detach WatchableStatus from Status and Remove LogLevel Literal

callumforrester:
Pending https://github.com/bluesky/bluesky/pull/1585

Required because of an issue with the backwards-compatible way in which Bluesky imports protocols.

`LogLevel` literal was removed and replaced with a string because pydantic has issues with `Literal` because it is actually a function, not a class.


DiamondJoseph:
For reference https://github.com/pydantic/pydantic/issues/5821

DiamondJoseph (approved):

DiamondJoseph:
Do we want an issue to revert this once the above is fixed?

callumforrester:
In favour of #226

PR #226, opened 2023-05-23
Pin typing extensions module to <4.6

callumforrester:
4.6 is causing various issues with protocols and pydantic, see https://github.com/pydantic/pydantic/issues/5821


DiamondJoseph (approved):

stan-dot:
https://pypi.org/project/typing-extensions/#history

now (March 2024) the latest version of `typing_extensions` is 4.10. 


it should be checked whether the exact issue can be reproduced.

connected to #273

PR #228, opened 2023-05-23
Mock sleep

rosesyrett:
This is a follow on from a discussion with @DiamondJoseph on https://github.com/DiamondLightSource/blueapi/pull/202. We sat down and got asyncio.sleep to be mocked in the RunEngine so that reworker tests don't have to wait for anything.


RAYemelyanova:
I will rebase this once the transaction worker has been merged... and change the history

callumforrester:
Very out of date, hopefully can be dealt with in follow-ons from #611

PR #229, opened 2023-05-24
Allow pausing and resuming of the worker

DiamondJoseph:


src/blueapi/service/main.py
```diff
@@ -87,6 +89,55 @@ def get_state(handler: Handler = Depends(get_handler)) -> WorkerState:
...
+                If defer is set to true, will wait until the next checkpoint to pause.
+        Returns: 202 "Accepted" if the worker is running, else 400 "Bad Request"
+    - RESUME: Attempts to resume a paused worker
+        Returns: 202 "Accepted" if the worker is paused, else 400 "Bad Request"
+    Rejected values:
+    - All others.
+        Returns: 400 "Bad Request"
+    """
+    current_state = handler.worker.state
+    if new_state in _ALLOWED_TRANSITIONS and current_state in _ALLOWED_TRANSITIONS[new_state]:
```
tpoliaw, line 124: Reversing `_ALLOWED_TRANSITIONS` to be `current -> allowed next state` would let this be reduced to `new_state in ALLOWED[current]`

src/blueapi/service/main.py
```diff
@@ -87,6 +89,55 @@ def get_state(handler: Handler = Depends(get_handler)) -> WorkerState:
...
+        response.status_code = status.HTTP_202_ACCEPTED
+        if new_state == WorkerState.SHUTDOWN:
+            def shutdown():
+                parent = psutil.Process(os.getpid())
+                for child in parent.children(recursive=True):
+                    child.kill()
+                parent.kill()
```
tpoliaw, line 131: Is it possible to have a slightly less brutal way of shutting down as well? Maybe one that checks nothing is running or waits until the current plan is complete?

tpoliaw (commented):
It doesn't feel very RESTy to use query parameters for the payload of a PUT request. Should it be a `{"status": "pause", "defer": true}` body instead?

src/blueapi/service/main.py
DiamondJoseph: After speaking with Callum we've removed shutting down as an option: should be handled from command line or K8s.

src/blueapi/service/main.py
DiamondJoseph: Will flip them over. Currently is only map of PAUSED: RUNNING and RUNNING:PAUSED, so it's as simple as changing the conditional.

src/blueapi/service/main.py
```diff
@@ -177,7 +177,7 @@ def set_state(
         elif new_state == WorkerState.RUNNING:
             handler.worker.resume()
     else:
-        response.status_code = status.HTTP_400_BAD_REQUEST
+        response.status_code = status.HTTP_304_NOT_MODIFIED
```
keithralphs: I'm not convinced this should change from 404. 304 is support to indicate to a client that it can continue using its cached version of a resource because the server doesn't have an updated one, i.e. it doesn't indicate that the client has requested an illegal state transition (because it's not an error code). Hence I think it should stay as BAD_REQUEST, as that's what it is.

tests/service/test_rest_api.py
```diff
@@ -133,7 +133,7 @@ def test_running_while_idle_denied(
 
     assert client.get("/worker/state").text == f'"{WorkerState.IDLE.name}"'
     response = client.put(f"/worker/state", json={"new_state": WorkerState.RUNNING.name})
-    assert response.status_code is status.HTTP_400_BAD_REQUEST
+    assert response.status_code is status.HTTP_304_NOT_MODIFIED
```
keithralphs: see above

tests/service/test_rest_api.py
```diff
@@ -146,7 +146,7 @@ def test_pausing_while_idle_denied(
 
     assert client.get("/worker/state").text == f'"{WorkerState.IDLE.name}"'
     response = client.put(f"/worker/state", json={"new_state": WorkerState.PAUSED.name})
-    assert response.status_code is status.HTTP_400_BAD_REQUEST
+    assert response.status_code is status.HTTP_304_NOT_MODIFIED
```
keithralphs: see above

keithralphs (changes requested):

src/blueapi/service/main.py
```diff
@@ -140,6 +142,50 @@ def get_state(handler: Handler = Depends(get_handler)) -> WorkerState:
...
+        new_state may be PAUSED to pause.
+        defer may be True to defer the pause until the new checkpoint.
+    All other values of new_state will result in 400 "Bad Request"
+    """
+    current_state = handler.worker.state
+    new_state = state_change_request.new_state
+    if (
+        current_state in _ALLOWED_TRANSITIONS
+        and new_state in _ALLOWED_TRANSITIONS[current_state]
+    ):
```
joeshannon, line 180: If this is false should there be some handling to return the 400?

joeshannon: Ah does the response default to 400 due to 
```
status_code=status.HTTP_400_BAD_REQUEST
```
?

DiamondJoseph: Yes, the default comes from the decorator and we only need to set it on the response (and do not need to return the response) if it's not the default.

src/blueapi/cli/rest.py
```diff
@@ -40,6 +55,19 @@ def get_device(self, name: str) -> DeviceModel:
...
+        self,
+        state: Literal[WorkerState.RUNNING, WorkerState.PAUSED],
+        defer: Optional[bool] = False,
+    ):
+        return self._request_and_deserialize(
+            "/worker/state",
+            target_type=WorkerState,
+            method="PUT",
+            data={"new_state": state, "defer": defer},
+            raise_if=_allow_exceptions([400], _is_exception),
```
RAYemelyanova, line 68: In line 101 this function is the default anyways. I think it reads a bit neater to exclude it, i.e. 
`raise_if=_allow_exceptions([400])`

keithralphs (approved):

src/blueapi/cli/rest.py
RAYemelyanova: Also; why are we allowing 400? Why aren't we raising on this exception, given in service/main.py it's the principal exception that can be raised from this endpoint?

src/blueapi/cli/rest.py
```diff
@@ -19,6 +19,21 @@
...
+
+def _allow_exceptions(
+    allowed: Iterable[int], wrapped: Callable[[requests.Response], bool]
+) -> Callable[[requests.Response], bool]:
+    def _pass_exceptions(response: requests.Response) -> bool:
+        if response.status_code in allowed:
+            return False
+        return wrapped(response)
+
+    return _pass_exceptions
```
RAYemelyanova, line 34: nice

RAYemelyanova (approved):
just a few questions/comments

src/blueapi/cli/rest.py
RAYemelyanova: also; doesn't look like this is being tested. Can we test this? In which case maybe we don't want it to be `_allow_exceptions` but just `allow_exceptions`

RAYemelyanova (approved):
ty

PR #230, opened 2023-05-24
Test cli

DiamondJoseph:
Part of #198 

For discussion: What should putting a task with invalid parameters, or putting a task that does not exist do?

Dependent on #202

### DIFF
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -2,6 +2,8 @@ from contextlib import asynccontextmanager
 from typing import Dict, Set
 
 from fastapi import Body, Depends, FastAPI, HTTPException, Request, Response, status
+from pydantic import ValidationError
+from starlette.responses import JSONResponse
 
 from blueapi.config import ApplicationConfig
 from blueapi.worker import RunPlan, TrackableTask, WorkerState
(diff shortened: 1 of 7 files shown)


src/blueapi/worker/reworker.py
```diff
@@ -114,6 +114,8 @@ def begin_task(self, task_id: str) -> None:
 
     def submit_task(self, task: Task) -> str:
         task_id: str = str(uuid.uuid4())
+        if isinstance(task, RunPlan):
+            self._ctx.plans[task.name].model.validate(task.params)
```
callumforrester, line 118: From discussion: Call `_lookup_params` and cache result until plan starts rather than validating twice.

src/blueapi/service/main.py
```diff
@@ -80,7 +81,14 @@ def get_device_by_name(name: str, handler: Handler = Depends(get_handler)):
         raise HTTPException(status_code=404, detail="Item not found")
 
 
-@app.post("/tasks", response_model=TaskResponse, status_code=201)
+@app.post(
+    "/tasks",
+    responses={
+        status.HTTP_404_NOT_FOUND: {"detail": "Task not found"},
+        status.HTTP_400_BAD_REQUEST: {"detail": "Params invalid"},
```
tpoliaw, line 88: There's also `422 Unprocessable Content` that might fit here

keithralphs (approved):

PR #231, opened 2023-05-24
Allow splitting of creating and starting tasks in API

callumforrester:


docs/user/reference/openapi.yaml
```diff
@@ -88,14 +88,39 @@ components:
...
+      - taskId
       title: TaskResponse
       type: object
+    TrackableTask:
+      additionalProperties: false
+      description: A representation of a task that the worker recognizes
+      properties:
+        isComplete:
+          default: false
+          title: Iscomplete
```
DiamondJoseph, line 105: The autogenerated titles are a bit naff
Should be possible to make them a bit nicer on the Pydantic model
Field(title="The description of the item")

docs/user/reference/openapi.yaml
```diff
@@ -240,6 +275,31 @@ paths:
...
+          title: Task Id
+          type: string
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TrackableTask'
+          description: Successful Response
+        '422':
```
DiamondJoseph, line 296: After talking with Rose, we actually return 422 in many places, so it's OK to return this.
But we do also want the 404 to be called out here, especially as it's raised from the HTTPException in the method it's strange it hasn't caught it. Possibly in the decorater.responses={status.HTTP_404_NOT_FOUND: ["item not found"]}?

src/blueapi/service/main.py
```diff
@@ -77,10 +84,38 @@ def submit_task(
...
-    handler.worker.begin_task(task_id)
     return TaskResponse(task_id=task_id)
 
 
+@app.put("/worker/task", response_model=WorkerTask)
+def update_task(
+    task: WorkerTask,
+    handler: Handler = Depends(get_handler),
+) -> WorkerTask:
+    if task.task_id is not None:
```
keithralphs, line 95: I think this should also check here that there isn't already a task running on the worker and return an error status (probably 409 Conflict) with an appropriate message if there is.

src/blueapi/worker/reworker.py
```diff
@@ -99,6 +99,12 @@ def clear_task(self, task_id: str) -> bool:
...
         return list(self._pending_tasks.values())
 
+    def get_pending_task(self, task_id: str) -> Optional[TrackableTask[Task]]:
+        return self._pending_tasks.get(task_id)
+
+    def get_active_task(self) -> Optional[TrackableTask[Task]]:
+        return self._current
+
     def begin_task(self, task_id: str) -> None:
         task = self._pending_tasks.get(task_id)
```
keithralphs, line 109: When does a task get removed from pending_tasks; should it be now?

tests/service/test_rest_api.py
```diff
@@ -74,6 +76,15 @@ class MyDevice:
...
+def test_create_task(handler: Handler, client: TestClient) -> None:
+    response = client.post("/tasks", json=_TASK.dict())
+    task_id = response.json()["taskId"]
+
+    pending = handler.worker.get_pending_task(task_id)
+    assert pending is not None
+    assert pending.task == _TASK
+
+
 def test_put_plan_submits_task(handler: Handler, client: TestClient) -> None:
```
keithralphs, line 88: Shouldn't we have a test_put_worker_plan_begins_task test now instead of this?

src/blueapi/worker/reworker.py
DiamondJoseph: Tasks are currently only removed from pending_tasks when clear_task is called on it, meaning pending_tasks contain tasks that are not is_pending, and various states of is_complete/is_error.
I think we need pending_tasks, active_task, finished_tasks eventually.

DiamondJoseph (changes requested):

src/blueapi/worker/reworker.py
callumforrester: https://github.com/DiamondLightSource/blueapi/issues/233

src/blueapi/service/main.py
```diff
@@ -43,7 +43,11 @@ def get_plans(handler: Handler = Depends(get_handler)):
     )
 
 
-@app.get("/plans/{name}", response_model=PlanModel)
+@app.get(
+    "/plans/{name}",
+    response_model=PlanModel,
+    responses={status.HTTP_404_NOT_FOUND: {"item": "not found"}},
```
DiamondJoseph, line 49: I think the default if the method throws a Key of Value error is "detail: "item not found", if we want to have it the same for neatness sake?

DiamondJoseph (approved):

Issue #235, opened 2023-05-25
Test that OpenAPI Schema is unchanged


joeshannon:
There is currently an action to generate the schema but this does not validate that it matches the schema present in the repository at
`docs/user/reference/openapi.yaml`.

It is important that these match as client libraries are generated from this file.

The check could probably be replaced with a Python test.

stan-dot:
additionally a versioning field might be added to the schema to show breaking changes https://stackoverflow.com/questions/54691517/service-versioning-in-openapi

stan-dot:
this job is key
https://github.com/DiamondLightSource/blueapi/actions/runs/8927707817/job/24521677000

callumforrester:
@joeshannon What do you need that https://github.com/DiamondLightSource/blueapi/blob/main/tests/service/test_openapi.py doesn't do?

joeshannon:
This is an old ticket that forgot to be closed I think.

joeshannon:
Fixed by #238

stan-dot:
https://github.com/DiamondLightSource/blueapi/pull/452#issuecomment-2093180637

I'm not sure @joeshannon

PR #236, opened 2023-05-25
Update OpenAPI Schema to match current state

joeshannon:
This was missed as part of a recent PR. In the future a check will be added to ensure that the committed file matches the current generated file.


joeshannon:
Created #235 and #237 to aim to prevent this happening in future.

PR #238, opened 2023-05-25
Write test to check openapi schema has not changed

callumforrester:
Changes
* Check schema has not changed in tests
* Write CLI command to regenerate schema
* Delete openapi Github action (now covered by the tests)


RAYemelyanova (approved):
Initially I wasn't sure I liked generate schema being a CLI command... I'm okay with it considering blueapi will be an internal service, and any schema that is generated is only used for the docs anyways so even if someone checks out the repo, makes changes to the API and generates the schema, unless they push it the actual docs wont change.

However... the tests do not actually generate the openapi schema. There are two tests:
- test_generate_schema just checks that, if the app has a slightly different title, or description, or whatever, the generated schema returned from the get_openapi function of fastapi.openapi.utils reflects those changes. 
- test_schema_updated seems to just check that the generated schema from the current openapi.yaml file in docs matches the one in the current application.

Could the 2nd test also generate this schema if they're different? Or is that considered bad test-practise?

callumforrester:
> Could the 2nd test also generate this schema if they're different? Or is that considered bad test-practise?

If we did that the test would always pass, which would mean it is not a useful test. The idea of this is: if someone has changed the API and not regenerated the schema, the CI should fail. This alerts them to the fact that they have changed the API in an "are you sure" way.

PR #243, opened 2023-05-25
Blocking cli command to run plan

RAYemelyanova:
The cli command to run a plan is now blocking.

### DIFF
--- a/src/blueapi/cli/amq.py
+++ b/src/blueapi/cli/amq.py
@@ -1,11 +1,9 @@
 import threading
 import time
-from typing import Callable, Optional, TypeVar
+from typing import Callable, Optional
 
 from blueapi.messaging import MessageContext, MessagingTemplate
-from blueapi.worker import ProgressEvent, WorkerEvent
-
-T = TypeVar("T")
+from blueapi.worker import WorkerEvent
 
 
 class BlueskyRemoteError(Exception):
@@ -31,7 +29,6 @@ class AmqClient:
         self,
         task_id: str,
         on_event: Optional[Callable[[WorkerEvent], None]] = None,
-        on_progress_event: Optional[Callable[[ProgressEvent], None]] = None,
     ) -> None:
         """Run callbacks on events/progress events with a given correlation id."""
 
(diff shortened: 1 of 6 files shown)


RAYemelyanova:
also Note: I've changed logging so that it is configured from the CLI not from the handler. Can take this out into its own PR if needed/desired

src/blueapi/cli/amq.py
```diff
@@ -21,60 +13,49 @@ def __init__(self, message: str) -> None:
...
+    def wait_for_complete(self, timeout: Optional[float] = None) -> None:
+        begin_time = time.time()
+        timedout = False
+        while not self.complete.is_set():
+            current_time = time.time()
+            if (timeout is not None) and ((current_time - begin_time) > timeout):
+                timedout = True
+                break
+
+        if timedout:
```
callumforrester, line 56: I believe the following will work, but test it:
```suggestion
        self.complete.wait(timeout=timeout)
        
        if not self.complete.is_set:
```

src/blueapi/cli/cli.py
```diff
@@ -87,16 +94,47 @@ def get_devices(obj: dict) -> None:
 @controller.command(name="run")
 @click.argument("name", type=str)
 @click.argument("parameters", type=str, required=False)
+@click.option(
+    "-t",
+    "--timeout",
+    type=float,
+    help="Timeout for the plan in seconds. None hangs forever",
```
callumforrester, line 101: I think we should word this in a more user friendly way given that hanging forever is a legitimate use case, we cannot predict how long some plans will take.
Maybe this...
```suggestion
    help="If supplied, terminates the client if the plan fails to finish within this time limit",
```

src/blueapi/messaging/context.py
```diff
@@ -10,4 +10,4 @@ class MessageContext:
 
     destination: str
     reply_destination: Optional[str]
-    correlation_id: Optional[str]
+    task_id: Optional[str]
```
callumforrester, line 13: This actually refers to header on the ActiveMQ message, which doesn't have anything to do with tasks etc. so it should be correlation_id and we shouldn't use it for now.

callumforrester (changes requested):

PR #247, opened 2023-05-26
Add github project slug to API catalogue entities

garryod:


catalog-info.yaml
```diff
@@ -22,6 +22,8 @@ metadata:
   name: blueapiControl
   title: Athena BlueAPI Control
   description: REST API for getting plans/devices from the worker (and running tasks)
+  annotations:
+    github.com/project-slug: DiamondLightSouce/blueapi
```
RAYemelyanova, line 26: looks good but I'm unsure of it's purpose. Could I ask what this does?

RAYemelyanova (approved):

catalog-info.yaml
garryod: It's used by a few of the plugins, the first one that pops to mind the GitHub actions one which will allow users to view the CI history of the project from the portal

keithralphs: Haven't we already got this on line 8, except it seems to have been misspelled as DiamondLightSouce? Why do we need it in the API section too, don't just need to correct the spelling on line 8?

keithralphs (changes requested):

catalog-info.yaml
garryod: Aha, good catch, I had completely missed that typo :sweat_smile: For needing it in multiple sections - as far as the portal is concerned these entities are entirely independent, hence if we want any of the goodies that come with linking to the repo (e.g. the aforementioned CI history) on the APIs then we need to include it in their definitions too

keithralphs: Ok fair enough then

keithralphs (approved):

Issue #199, opened 2023-05-15
Queryable status of worker and currently running/last running plan


callumforrester:
There should be REST endpoints and methods on the worker class that give access to:
* The current `WorkerState`
* The status of the last plan to be run/currently running plan

callumforrester:
Sounds good.

I think 1 is a long-term task, I would like to make a mechanism that allows us to encode progress information inside a plan. For 2 we also have the same problem, so if you have a nicer idea it might be worth a look. For 3 I think there may be a third concern as "how well did artemis work?" is different to "how well did my plan work?".

keithralphs:
Based on Dom's comments and to be compliant with API guidelines, the endpoint for this should probably be of the form:

GET <baseurl>/tasks/<identifier>/status

where <idenifier> could be 'current', 'last' or the task id generated and returned by the service when the task was created. This would also allow for requesting a longer history of task statuses if that were deemed useful/supported.

keithralphs:
Further to my comment above, in the latest version of the API submitted in #231, tasks which are running or have been run on the worker are going to have URLs of the form {baseurl}/worker/tasks/{identifier} so we probably want the URL form for this to be:

GET {baseurl}/worker/tasks/{identifier}/status

with {identifier} being as specified before

callumforrester:
Closing this as I think blueapi now supports the desired functionality.

PR #252, opened 2023-05-26
Allow the currently running task to be Stopped or Aborted

DiamondJoseph:
Part of #198 
Additionally exposes clearing pending tasks to the REST API, also a requirement for #198

### DIFF
--- a/docs/user/reference/openapi.yaml
+++ b/docs/user/reference/openapi.yaml
@@ -357,7 +357,7 @@ paths:
       summary: Get State
     put:
       description: "Request that the worker is put into a particular state.\nReturns\
-        \ the state of the worker at the end of the call.\n- **The following transitions\
+        \ the state of the worker at the end of the call.\n\n- **The following transitions\
         \ are allowed and return 202: Accepted**\n- If the worker is **PAUSED**, new_state\
         \ may be **RUNNING** to resume.\n- If the worker is **RUNNING**, new_state\
         \ may be **PAUSED** to pause:\n    - If defer is False (default): pauses and\
--- a/docs/user/reference/openapi.yaml
+++ b/docs/user/reference/openapi.yaml
@@ -366,7 +366,7 @@ paths:
         \ task will instead be Aborted**\n- If the worker is **RUNNING/PAUSED**, new_state\
         \ may be **STOPPING** to stop.\n    Stop marks any currently open Runs in\
         \ the Task as a success and ends the task.\n- If the worker is **RUNNING/PAUSED**,\
-        \ new_state may be **STOPPING** to abort.\n    Abort marks any currently open\
+        \ new_state may be **ABORTING** to abort.\n    Abort marks any currently open\
         \ Runs in the Task as a Failure and ends the task.\n    - If reason is set,\
         \ the reason will be passed as the reason for the Run failure.\n- **All other\
         \ transitions return 400: Bad Request**"
--- a/docs/user/reference/openapi.yaml
+++ b/docs/user/reference/openapi.yaml
@@ -95,6 +95,11 @@ components:
           type: boolean
         new_state:
           $ref: '#/components/schemas/WorkerState'
+        reason:
+          default: ''
+          description: The reason for the current run to be aborted
+          title: Reason
+          type: string
       required:
       - new_state
       title: StateChangeRequest
@@ -292,6 +297,29 @@ paths:
           description: Validation Error
       summary: Submit Task
   /tasks/{task_id}:
+    delete:
+      operationId: delete_submitted_task_tasks__task_id__delete
+      parameters:
+      - in: path
+        name: task_id
+        required: true
+        schema:
+          title: Task Id
+          type: string
+      responses:
+        '200':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TaskResponse'
+          description: Successful Response
+        '422':
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/HTTPValidationError'
+          description: Validation Error
+      summary: Delete Submitted Task
     get:
       description: Retrieve a task
       operationId: get_task_tasks__task_id__get
@@ -337,7 +365,15 @@ paths:
 
         If the worker is RUNNING, new_state may be PAUSED to pause and
 
-        defer may be True to defer the pause until the new checkpoint.
+        defer may be True to defer the pause until the new checkpoint;
+
+        STOPPING to stop execution of the current task, marking the
+
+        current run as a Success; ABORTING to stop execution of the
+
+        current task, marking the current run as Aborted, reason may be
+
+        a string reason for why the run is to be aborted.
 
         All other values of new_state will result in 400 "Bad Request"'
       operationId: set_state_worker_state_put
--- a/docs/user/reference/openapi.yaml
+++ b/docs/user/reference/openapi.yaml
@@ -96,7 +96,6 @@ components:
         new_state:
           $ref: '#/components/schemas/WorkerState'
         reason:
-          default: ''
           description: The reason for the current run to be aborted
           title: Reason
           type: string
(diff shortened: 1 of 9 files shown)


src/blueapi/worker/reworker.py
```diff
@@ -89,12 +89,24 @@ def __init__(
...
-            return False
+    def clear_task(self, task_id: str) -> str:
+        task = self._pending_tasks.pop(task_id)
+        return task.task_id
+
+    def cancel_active_task(
+        self,
+        cancel_type: WorkerState,
+        reason: Optional[str] = None,
+    ) -> str:
```
callumforrester, line 100: I'm not in love with being able to pass a state that has nothing to do with stopping into the cancel method. Could have
```suggestion
    def cancel_active_task(
        self,
        failure: bool = False,
        reason: Optional[str] = None,
    ) -> str:
```

tests/worker/test_reworker.py
```diff
@@ -162,7 +162,8 @@ def test_clear_task(worker: Worker) -> None:
 
 
 def test_clear_nonexistant_task(worker: Worker) -> None:
-    assert not worker.clear_task("foo")
+    with pytest.raises(KeyError):
+        worker.clear_task("foo")
```
callumforrester, line 166: @tpoliaw may have views on this since I had this way originally and he preferred to return a bool

src/blueapi/worker/reworker.py
DiamondJoseph: mypy wasn't a fan of Literal[WorkerState.STOPPING, WorkerState.ABORTING], even after if new_state in {a, s}:...
This doesn't seem too bad. failure == new_state is WorkerState.ABORTING to have default = stop I guess. Will do now.

tests/worker/test_reworker.py
DiamondJoseph: This way is slightly neater to make use of the exceptionHandler was my rationale

tpoliaw: My main reservation about it was when there were discussions of pending tasks could expire and be automatically removed. It didn't make sense for clearing a task that had already been removed to be an error. I'm not strongly against either approach although would err on the side of calling clear multiple times being a no-op instead of an error.

src/blueapi/cli/cli.py
```diff
@@ -202,6 +202,32 @@ def resume(obj: dict) -> None:
     pprint(client.set_state(WorkerState.RUNNING))
 
 
+@controller.command(name="abort")
+@check_connection
+@click.argument("reason", type=str, required=False, default="")
+@click.pass_obj
+def abort(obj: dict, reason: Optional[str] = "") -> None:
```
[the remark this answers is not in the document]
callumforrester: Also the default is set twice, on line 207 and 209

src/blueapi/worker/reworker.py
```diff
@@ -89,12 +89,24 @@ def __init__(
...
+        if self._current is None:
+            raise KeyError("Worker has no active task")
```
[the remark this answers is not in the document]
callumforrester: There is already `WorkerBusyError`, could just expand that module

tests/worker/test_reworker.py
DiamondJoseph: Per https://stackoverflow.com/questions/6439416/status-code-when-deleting-a-resource-using-http-delete-for-the-second-time I'm recommending sticking with a 404 when deleting a resource that doesn't exist.

src/blueapi/worker/reworker.py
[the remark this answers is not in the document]
DiamondJoseph: It's actually an unreachable state I've realised: we only allow calling Stop/Abort if the run engine is currently Running, which means that _current cannot be None. I've put a Bluesky TransitionError, which is what the Run Engine would raise if somehow we did call it while its state isn't Running, but only to make mypy happy.

docs/user/reference/openapi.yaml
```diff
@@ -337,7 +365,15 @@ paths:
 
         If the worker is RUNNING, new_state may be PAUSED to pause and
```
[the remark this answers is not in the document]
DiamondJoseph: Reworded the doc string to use markdown to hopefully be a bit clearer as to what it's saying, and resolved the double spacing:
image

src/blueapi/service/model.py
```diff
@@ -112,3 +112,7 @@ class StateChangeRequest(BlueapiBaseModel):
         description="Should worker defer Pausing until the next checkpoint",
         default=False,
     )
+    reason: Optional[str] = Field(
```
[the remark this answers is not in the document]
DiamondJoseph: Current implementation is stopping/aborting the running task is by putting aborting/stopping to /worker/state and so it belongs here.

If we had Task State that supported Stopped/Aborted (~ #199 if we formalise the schema of task states, or put Stopped/Aborted/Reason in errors?), could be to /worker/task/state, or just updating /worker/task with a new state and it could be on WorkerTask. I invite thoughts on whether that's preferable:
I wanted cancelling a task to be DELETE /worker/task, because we make a task run by PUT /worker/task and it seems analogous: but DELETE is discouraged from having a body and the library we use for tests only allows it in a workaround, and we require Abort reason at least.

We already Pause/Resume a task by pausing/resuming the RunEngine, by requesting a change of the state of the Worker, which updates its state whenever the state of the RunEngine changes. And it seems again neat to Stop/Abort a task by calling stop/abort on the RunEngine, by requesting a change of the state of the Worker...

I think either pause/resume/stop/abort all belong on /worker/task/state, and we need to merge #199 and refactor #229 and this, or pause/resume/stop/abort belong on /worker/state and this is for now the correct place.

docs/user/reference/openapi.yaml
```diff
@@ -357,25 +356,20 @@ paths:
...
-        STOPPING to stop execution of the current task, marking the
-
-        current run as a Success; ABORTING to stop execution of the
-
-        current task, marking the current run as Aborted, reason may be
-
-        a string reason for why the run is to be aborted.
-
-        All other values of new_state will result in 400 "Bad Request"'
+      description: "Request that the worker is put into a particular state.\nReturns\
```
tpoliaw, line 359: Doesn't one of the many ways of writing strings in yaml let you avoid having to insert literal `\n`s for line breaks?

src/blueapi/worker/reworker.py
```diff
@@ -95,17 +96,17 @@ def clear_task(self, task_id: str) -> str:
...
         if self._current is None:
-            raise KeyError("Worker has no active task")
-        if self._current.is_complete:
-            raise TypeError("Task already complete")
-        if cancel_type is WorkerState.STOPPING:
-            self._ctx.run_engine.stop()
-        if cancel_type is WorkerState.ABORTING:
+            # Persuades mypy that self._current is not None
+            # We only allow this method to be called if a Plan is active
+            raise TransitionError
```
tpoliaw, line 105: Should this be an instance rather than the class?

src/blueapi/service/model.py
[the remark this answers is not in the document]
keithralphs: I think they should all be on /worker/state as they are all to do with the state of the Run engine (i.e. worker) in which case the task element of the URL is redundant in evaluating that, as the there is only ever 0 or 1 running task. Also that would then lead to appropriate error conditions if no task was attached to the RE. In this model the /worker/task endpoint is for a different purpose - to see if there is one loaded into the worker and what it is using GET, or to attach one to the worker using PUT.

docs/user/reference/openapi.yaml
DiamondJoseph: Closing as we discussed, it's just the autogenerated schema

callumforrester (approved):

src/blueapi/worker/reworker.py
DiamondJoseph: raise XError is equivalent to raise XError()

I've added a message to the exception just to make it distinct from the TransitionError in the RunEngine that it prevents us reaching, in case we ever call not from the REST client that immediately catches it.

PR #254, opened 2023-05-31
Refactor validation

callumforrester:
This is a refactor with no changes in behaviour. It was motivated by the fact that `RunEngineWorker` accessed the private function `_lookup_params` in `task.py`. 
* Deprecate `RunPlan` and move its functionality into `Task`
* Add method to `Task` that can optionally be called to cache the parameters, call it when run if not
* Remove instance check on task in worker as it is no longer needed


callumforrester:
This PR includes some WIP changes to fix concurrency issues in the worker, will extract those to a separate PR when complete

PR #263, opened 2023-06-15
Parametrize unit tests by stomp server and adjust CI

callumforrester:
Better alternative to having two sets of CI jobs, pytest should now create a suite of tests against `localhost:<port>` for each port we pass it


.github/enabled_plugins
```diff
@@ -0,0 +1 @@
+[rabbitmq_management,rabbitmq_stomp]
```
DiamondJoseph, line 1:
```suggestion
[rabbitmq_stomp]

```

We don't need the management console enabled during tests: I'm almost certain we can't get to view it anyway.

DiamondJoseph (changes requested):

DiamondJoseph (approved):

PR #266, opened 2023-06-28
Update example beamline config for dodal module structure

DiamondJoseph:


callumforrester:
As discussion, would be good to pin to a tag on dodal

PR #274, opened 2023-07-05
Update catalog-info.yaml to show data acquisition group as owner

keithralphs:


callumforrester:
This seems to have been done: https://github.com/DiamondLightSource/blueapi/blob/caf7e466dba86eba5eb38da2834548f670d04e8d/catalog-info.yaml#L13

PR #277, opened 2023-07-24
Update example code to be correct

DominicOram:
To test:
* Import the code into an IDE and confirm it has correct imports etc.


tpoliaw (approved):

DominicOram:
Thanks @tpoliaw, what's your workflow here? Are you happy for me to squash and merge?

tpoliaw:
I think squash and merge is the only strategy enabled. Would be good to remove the `To test:.. ` comment from the auto-generated message though.

PR #281, opened 2023-07-26
Changes to ref p38 dodal

keithralphs:


helm/blueapi/templates/deployment.yaml
```diff
@@ -35,6 +35,9 @@ spec:
           - secret:
               name: {{ . }}
           {{- end }}
+      - name: plans
+        hostPath:
+          path: "/dls_sw/p38/software/bluesky_plans"
```
DiamondJoseph, line 40: Un-hardcode this and get (this/these) from values.yaml if it exists

helm/blueapi/templates/deployment.yaml
```diff
@@ -47,6 +50,12 @@ spec:
             - name: worker-config
               mountPath: "/config"
               readOnly: true
+            - name: plans
+              mountPath: "/plans"
+              readOnly: true
+          env:
+            - name: PYTHONPATH
+              value: /plans
```
DiamondJoseph, line 58: Calllum wants to only allow python modules and not just a random folder

DiamondJoseph (changes requested):

stan-dot:
unclear ref changes, petition to close @callumforrester

callumforrester:
Yep, no longer needed

PR #283, opened 2023-07-31
Speculative visit directory service

callumforrester:
*PLEASE DO NOT MERGE!*
This PR is intended as a speculative prototyping exercise only.

**Design**

In order to comply with the DLS visit directory structure, we need to group acquisitions into files. There are many solutions to this, the basics of this one:

image

1. A new service is responsible for determining where data will go for a particular acquisition. This functionality is separated from all other applications for two reasons:
    * So that other data collection applications can also use it each one is guaranteed unique data locations. 
    * This involves making directories and therefore requires write access to places where data goes. For security reasons, only trusted code should have that level of privilege, and it should be kept small and isolated.
3. Every time a group of devices is staged, blueapi will call out to this service (or use a dummy version for offline development) and get a unique data directory for this stage. This is achieved via a plan preprocessor that detects groups of stages and makes the call. 
4. Compatible Ophyd detectors will be aware of a mutable state which is updated when this call occurs, so they can write to the "current" visit.
5. The current group is bundled into the run start metadata. Allowing for a group:run mapping of 1:1 or 1:many. The use case for the latter is writing many runs to a single file for performance reasons.
6. Any aggregating service, such as a nexus file writer, can take the group information from the run starts and use it how it likes. Our initial thinking was to map each group to a nexus file and reach run to an entry, but that is TBD.

**Design Issues**

I remain _very_ sceptical about point 2! It introduces a shared mutable state between all detectors just to get around the fact that there is no direct, universal way to pass the current visit down to them, by design! The other alternative I explored is to put all support for configuring individual detectors in one place, in blueapi, and if your detector isn't supported it will just write as it normally does. Simple example

```python
def configure_data_writing(
    devices: Iterable[Device],
    collection: DataCollection,
) -> None:
    for device in devices:
        if isinstance(device, FileStoreBase):
            path_template = str(collection.raw_data_files_root)

            # Configure Ophyd Device to setup HDF5 writer
            device.reg_root = "/"
            device.read_path_template = path_template
            device.write_path_template = path_template
```

**What's Prototyped Here?**

The actual contents of this PR:

* A very simple visit service that can create and inform clients about unique directories for "acquisition groups".
* A plan preprocessor that calls out to that service (or use an internal dummy version) to create a new group when a set of devices is staged. It also puts metadata into the run start document to show which group it is a part of. Each start document now contains a key called `data_collection_number`, which can be passed back to the visit service for more information. 
* A dummy device that has access to the cached result of the service call.
* A series of tests ensuring that the correct mapping of run -> file is achieved for various use cases, such as multi-run plans.
* A small amendment to the bluesky context that makes blueapi use this preprocessor on every plan it runs (if the plan does not contain any `stage`s, it will be unaffected)

**How to test**

Checkout this branch. To test with the dummy detector, run

```shell
tox -e pytest -- tests/plugins/test_data_writing.py
```

To start the visit service:

```shell
python -m blueapi.plugins.data_writing_server
```

It should run on port 8089, you can test it with

```shell
curl -X POST "http://localhost:8089/collection/mybeamline"
```

Which should create a new group and return a blob of JSON about it.

Running a scan with blueapi should now produce a start document that contains an attribute called `data_collection_number` which we can use to aggregate runs.

**Still to do**

This PR contains a dummy detector with the correct hooks. It would be good to get this working with an actual Ophyd v2 detector, and ideally a v1 one too. We could then run a full collection via blueapi that uses this functionality @coretl @callumforrester  

More discussion on point 2. of the design would be helpful, I think. @coretl @DiamondJoseph @keithralphs @callumforrester  

We should try to get the nexus writer to interpret the group number in the run start document and ensure there are no problems there @DiamondJoseph 

Possibly leverage the `data_session` metadata in the run start schema. @callumforrester  @DiamondJoseph


src/blueapi/core/device_walk.py
```diff
@@ -0,0 +1,17 @@
+from typing import Iterable, Tuple
+
+from blueapi.core import Device, HasComponents
+
+DeviceTreeNode = Tuple[Device]
+
+
+def walk_devices(devices: Iterable[Device]) -> Iterable:
```
coretl, line 8: Where is this used?

tests/plugins/file_writing_detector.py
```diff
@@ -0,0 +1,27 @@
...
+        self.stage_sigs[self.image_count] = 0
+        self._provider = provider
+
+    def trigger(self, *args, **kwargs):
+        return self.image_count.set(self.image_count.get() + 1)
+
+    def stage(self) -> List[object]:
+        collection = self._provider.current_data_collection
+        if collection is None:
+            raise Exception("No active collection")
```
coretl, line 25: This seems pretty readable to me, requires explicit knowledge of a DataCollectionProvider when making the class, but that makes it obvious where the collection number is coming from

src/blueapi/plugins/data_writing.py
```diff
@@ -0,0 +1,106 @@
...
+import bluesky.plan_stubs as bps
+from aiohttp import ClientSession
+from bluesky.utils import make_decorator
+
+from blueapi.core import MsgGenerator
+from blueapi.plugins.data_writing_server import DataCollection
+
+from .data_writing_server import DataCollection, DataCollectionSetupResult
+
+DATA_COLLECTION_NUMBER = "data_collection_number"
```
DiamondJoseph, line 15: Is there a reason to favour data_collection_number over data_session, which is an existing field on StartDocument? From the event_model docs: "An optional field for grouping runs. The meaning is not mandated, but this is a data management grouping and not a scientific grouping. It is intended to group runs in a visit or set of trials."

src/blueapi/core/context.py
```diff
@@ -51,12 +55,21 @@ class BlueskyContext:
...
     devices: Dict[str, Device] = field(default_factory=dict)
     plan_functions: Dict[str, PlanGenerator] = field(default_factory=dict)
 
     _reference_cache: Dict[Type, Type] = field(default_factory=dict)
 
+    def wrap(self, plan: MsgGenerator) -> MsgGenerator:
+        wrapped_plan = functools.reduce(
+            lambda wrapped, next_wrapper: next_wrapper(wrapped),
+            self.plan_wrappers,
+            plan,
```
DiamondJoseph, line 69: Might be obvious to people coming from python, but making plan be a kwarg `initializer=plan` would stop me running off to read the docs for functools every time I see it.

src/blueapi/plugins/data_writing_server.py
```diff
@@ -0,0 +1,76 @@
...
+def create_collection(group: str) -> DataCollectionSetupResult:
+    num = next(SCAN_NUMBER)
+    if group not in COLLECTIONS:
+        COLLECTIONS[group] = {}
+    if num in COLLECTIONS[group]:
+        raise Exception("Collection already exists")
+
+    data_root = get_data_root(group)
+    raw_data_files_root = data_root / f"{group}-{num}"
+    nexus_file_path = data_root / f"{group}-{num}.nxs"
```
DiamondJoseph, line 39: Nexus service currently has config for Nexus file extension, if we're going to also pass it from the visit service it should be because it is not hard coded and may change. I don't know if it ever will, so may be able to make this class a little shorter and just have data_files_root

callumforrester: I think that's fine, tell services roughly where to put things then leave the details to them

src/blueapi/plugins/data_writing.py
callumforrester: Nope, data session is fine, finer details of the API can be changed, this is just a rough sketch

src/blueapi/core/device_walk.py
callumforrester: Nowhere any more, was used in a previous iteration of a prototype.

tests/plugins/file_writing_detector.py
callumforrester: From discussion, happy with this because it is theoretically easy to strip out later, as long as:

- We keep in mind that we want to strip it out later
- We make an interface or ABC for the singleton in ophyd-async
- We make a DLS-specific implementation in blueapi
- We maintain a parallel implementation in ophyd-async for offline use, that just writes to a preconfigured directory or similar
- We inject references to the singleton into each device

src/blueapi/plugins/data_writing_server.py
```diff
@@ -0,0 +1,76 @@
+import itertools
```
callumforrester, line 1: This whole file is just a prototype version of the visit service, now embedded in GDA, and should be deleted!

src/blueapi/plugins/data_writing.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class DataCollectionProvider(ABC):
```
callumforrester, line 18: This should be removed and replaced with <thing Tom is about to push>

callumforrester: https://github.com/bluesky/ophyd-async/commit/7b5035067be3d7347fa56a16422e5decdc13433d

src/blueapi/plugins/data_writing.py
```diff
@@ -0,0 +1,106 @@
...
+                yield from bps.wait_for([provider.update])
+            if provider.current_data_collection is None:
+                raise Exception("There is no active data collection")
+            message.kwargs[
+                DATA_COLLECTION_NUMBER
+            ] = provider.current_data_collection.collection_number
+        yield message
+
+
+data_writing_decorator = make_decorator(data_writing_wrapper)
```
callumforrester, line 106: This is the only actual part of this file that should stick around. I'm not convinced it should be called `plugins/data_writing.py` either. Maybe `preprocessors/xyz.py`

src/blueapi/core/device_walk.py
callumforrester: Should be deleted

callumforrester: In fact I think this whole file can go

src/blueapi/core/context.py
```diff
@@ -51,12 +55,21 @@ class BlueskyContext:
...
         default_factory=lambda: RunEngine(context_managers=[])
     )
+    plan_wrappers: Sequence[PlanWrapper] = field(default_factory=list)
     plans: Dict[str, Plan] = field(default_factory=dict)
     devices: Dict[str, Device] = field(default_factory=dict)
     plan_functions: Dict[str, PlanGenerator] = field(default_factory=dict)
 
     _reference_cache: Dict[Type, Type] = field(default_factory=dict)
 
+    def wrap(self, plan: MsgGenerator) -> MsgGenerator:
```
callumforrester, line 65: This should be tested

src/blueapi/core/bluesky_types.py
```diff
@@ -129,3 +132,8 @@ def watch(self, __func: Callable) -> None:
         """
 
         ...
+
+
+@runtime_checkable
+class HasComponents(Protocol):
+    component_names: List[str]
```
callumforrester, line 139: This can also be deleted

src/blueapi/worker/task.py
```diff
@@ -29,7 +29,8 @@ def do_task(self, ctx: BlueskyContext) -> None:
         func = ctx.plan_functions[self.name]
         prepared_params = self._ensure_params(ctx)
         plan_generator = func(**prepared_params.dict())
-        ctx.run_engine(plan_generator)
+        wrapped_plan_generator = ctx.wrap(plan_generator)
+        ctx.run_engine(wrapped_plan_generator)
```
callumforrester, line 33: This needs testing too

src/blueapi/plugins/data_writing.py
```diff
@@ -0,0 +1,106 @@
...
+        scan_number = next(self._scan_number)
+        self._current_collection = DataCollection(
+            collection_number=scan_number,
+            group=self._collection_group,
+            raw_data_files_root=Path(f"/tmp/{self._collection_group}"),
+            nexus_file_path=Path(f"/tmp{self._collection_group}.nxs"),
+        )
+
+
+def data_writing_wrapper(
```
callumforrester, line 80: There may be a nicer way of doing this, @rosesyrett to investigate and make my flowchart look less ugly

src/blueapi/plugins/data_writing.py
rosesyrett: I can't see a direct replacement to the DataCollectionProvider in the file you've linked there...

your DataCollectionProvider gives a DataCollection, which has 4 things: collection_number, group, raw_data_files_root and nexus_file_path. Tom's work that you've linked has a DirectoryProvider which gives a DirectoryInfo object, which just contains a directory path and filename prefix.

So it's not a direct mapping. Perhaps I'll chat to you about this tomorrow morning in more detail.

src/blueapi/plugins/data_writing.py
DiamondJoseph: If it's going to be a plan preprocessor should it be moved to dls-bluesky-core?

callumforrester: I thought about this. It is specifically a part of blueapi, that fudges plans to write data in our structure, maybe it shouldn't be allowed to get outside of blueapi?

DiamondJoseph: I'm fine with it existing only within BlueAPI and being applied on plan import: prevents possibility of attempting to apply multiple times (pretty sure from glancing through the logic that would just doubly increment the data collection number each run, but that's disruptive enough) or being missed. 

Should anyone want to be a. using vanilla Bluesky/not BlueAPI, b. wanting to write to DLS filesystem, they can extract and duplicate the logic. But I don't think that'll be very often.

src/blueapi/plugins/data_writing.py
rosesyrett: Cool, got it

rosesyrett: I've come across a bit of an annoyance; I want to subclass `DirectoryProvider` so it gets the directory from the visit service. However, I want to use aiohttp or some similar asynchronous library... `DirectoryProvider` has one synchronous method, `__call__`.

My solution for now, is to make a subclass `GDADirectoryProvider` (happy to get other suggestions) and have a async `update` method on this. Then the synchronous `__call__` can do something like `asyncio.wait_for` on it.

callumforrester: Yep, I think that's what I told @coretl I had in mind at the time

coretl: As discussed, the pre-processor should do the `async` access to the directory service, then cache the result in `DirectoryProvider`, then the `__call__` method just returns the cached value

keithralphs (commented):
to be discussed further...

src/blueapi/core/context.py
callumforrester: Reduce explicitly disallows keyword arguments

src/blueapi/data_management/gda_directory_provider.py
```diff
@@ -0,0 +1,146 @@
...
+    async def create_new_collection(self) -> DataCollectionIdentifier:
+        count = self._count
+        self._count += 1
+        return DataCollectionIdentifier(collectionNumber=count)
+
+    async def get_current_collection(self) -> DataCollectionIdentifier:
+        return DataCollectionIdentifier(collectionNumber=self._count)
+
+
+class VisitDirectoryProvider(DirectoryProvider):
```
callumforrester, line 70: Needs tests

src/blueapi/core/context.py
```diff
@@ -86,6 +104,18 @@ def with_config(self, config: EnvironmentConfig) -> None:
             elif source.kind is SourceKind.DODAL:
                 self.with_dodal_module(mod)
 
+        call_in_bluesky_event_loop(self.connect_devices(self.sim))
+
+    async def connect_devices(self, sim: bool = False) -> None:
```
callumforrester, line 109: Should make private and rename to `_connect_ophyd_async_devices`

stan-dot:
willing to test with with the webcam at visr https://github.com/DiamondLightSource/ViSR

@coretl  and @callumforrester  is this a good idea?

callumforrester:
This is very old and out-of-date, everything in it now exists in blueapi in some form.

PR #284, opened 2023-07-31
Working minimal authn/authz configuration for blueapi

rosesyrett:
This adds some authentication/authorization config to the helm chart


RAYemelyanova:
CI failing for unrelated reason which would probably make it fail on main also. Investigating now.

helm/blueapi/templates/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
-        {{- include "blueapi.selectorLabels" . | nindent 8 }}
+        {{- include "blueapi.metadataLabels" . | nindent 8 }}
```
DiamondJoseph, line 19: {{- include "blueapi.selectorLabels" . | nindent 8 }}
{{- include "blueapi.extraLabels . | nindent 8 }}

I think it'd make more sense to present the extraLabels alongside the selectorLabels, rather than folding both into a single function: there are selectorLabels which are used by this release to map between resources and extraLabels, which have some meaning in a wider context.

helm/blueapi/Chart.yaml
```diff
@@ -17,3 +17,8 @@ type: application
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 # This is updated automatically by CI, the appVersion is automatically inserted by CI
 version: 0.1.0
+
+dependencies:
+  - name: rabbitmq
+    version: 11.16.2
+    repository: oci://registry-1.docker.io/bitnamicharts
```
DiamondJoseph, line 24:
```suggestion
dependencies:
  - name: rabbitmq
    version: 11.16.2
    repository: oci://registry-1.docker.io/bitnamicharts
    enabled: rabbitmq.enabled
```

If we add this to the conditions, we can also add 
rabbitmq:
  enabled: false

to our values.yaml to disable rabbitmq (once we have BlueAPI not requiring a message bus)

helm/blueapi/values.yaml
```diff
@@ -75,13 +75,41 @@ worker:
...
+        module: blueapi.startup.example_plans
       - kind: planFunctions
         module: blueapi.plans
   stomp:
     auth:
-      username: guest
-      passcode: guest
-    host: rabbitmq
+      username: foo
+      passcode: bar
```
DiamondJoseph, line 86: I'm going to make a ticket for describing how to make SealedSecret for BlueAPI and RabbitMQ shared information, I think for this default/example config, we can keep guest/guest and add to the readme for deploying that these values will need to be overwritten if rabbitmq.enabled and not exposed on localhost

helm/blueapi/templates/_helpers.tpl
```diff
@@ -50,6 +50,16 @@ app.kubernetes.io/name: {{ include "blueapi.name" . }}
...
+{{/*
+Metadata labels
+*/}}
+{{- define "blueapi.metadataLabels" -}}
+{{ include "blueapi.selectorLabels" . }}
+{{- range $key, $value := .Values.extraLabels }}
+{{- $key }}: {{ $value | quote }} # N.B. ensures your labels are correctly String->String
+{{- end }}
+{{- end }}
+
```
DiamondJoseph, line 62:
```suggestion
{{/*
Metadata labels
*/}}
{{- define "blueapi.extraLabels" -}}
{{- range $key, $value := .Values.extraLabels }}
{{- $key }}: {{ $value | quote }} # N.B. ensures your labels are correctly String->String
{{- end }}

```

helm/blueapi/values.yaml
DiamondJoseph, line 86:
```suggestion
      username: guest
      passcode: guest
```

helm/blueapi/values.yaml
```diff
@@ -75,13 +75,41 @@ worker:
...
+extraLabels: {}
+extraVolumes: {}
+extraContainers: {}
+
+rabbitmq:
+  persistence:
+    enabled: true # TODO: Best approach for storage
+  auth:
+    username: foo
+    password: bar
```
DiamondJoseph, line 100:
```suggestion
    username: guest
    password: guest
```

DiamondJoseph (changes requested):

stan-dot:
not sure what is this PR accomplishing at the moment, as the auth logic moved to deployments. petition to close @callumforrester

callumforrester:
I'll defer to @DiamondJoseph

DiamondJoseph:
This PR seems to mostly be configuring RMQ, is outdated and doesn't have any Auth considerations.

PR #291, opened 2023-08-10
Add page on adding plans and devices to blueapi

callumforrester:
Document how to add plans and devices to blueapi.
This is an early draft meant to be criticised. The design it describes is also not the final design and can definitely be improved upon. Would be good to reduce the overall number of steps required to get a plan in!


callumforrester:
Preview of the page here: https://diamondlightsource.github.io/blueapi/plan-docs/index.html

docs/user/how-to/add-plans-and-devices.rst
```diff
@@ -0,0 +1,97 @@
...
+Home of Code
+------------
+
+The code can be in any pip-installable package, such as:
+
+* A package on pypi
+* A Github repository
+* A local directory with a ``pyproject.toml`` file or similar.
+
+The easiest place to put the code is a repository created with the Python skeleton. Which can then become any of the above.
```
DominicOram, line 15: Could: Link to the python skeleton

docs/user/how-to/add-plans-and-devices.rst
```diff
@@ -0,0 +1,97 @@
...
+You would add the following into your configuration file:
+
+.. code:: yaml
+
+    env:
+      sources:
+        - kind: dodal
+          # note, this code does not have to be inside dodal just because it uses
+          # the dodal kind
+          module: dodal.my_beamline
```
DominicOram, line 89: Should: It's not immediately clear that this is referencing the file with `MyTypeOfDetector` in it

DominicOram (changes requested):
Thanks Callum, this is a good start. There's a lot of implied steps to this that are simple with some python knowledge but I don't think would be obvious to someone with a pure Java background like how to set up a python environment with all the constituent pieces for this to work. I realise this might not be the best place for a tutorial like that but I think we need one.

Ideally I think this guide (or maybe it's better suited to a tutorial) will walk you through the next steps on how to run `blueapi` pointing at this and how to then run a plan

DominicOram:
> This is an early draft meant to be criticised. The design it describes is also not the final design and can definitely be improved upon. Would be good to reduce the overall number of steps required to get a plan in!

I don't think we should let this block us from merging it, we can change it later if we want.

callumforrester:
@DominicOram I've added a link to the skeleton and its "new project" documentation, which will hopefully help with the missing steps for Python novices part. Although I wouldn't be surprised if those tutorials have their own assumptions that we will discover.

DominicOram (approved):
Thanks, I think this a good start we should get merged

PR #294, opened 2023-08-16
Refactor scan plan metadata to more effectively utilise underlying scan_nd implementation

DiamondJoseph:
- Remove fields of metadata duplicated by scan_nd (detectors) 
- Override incorrect values of metadata set by scan_nd (plan_name, plan_args)
- Provide metadata that is available to our implementation but not necessarily for scan_nd (shape)
- Rename metadata kwarg to match with Bluesky plans


src/blueapi/plans/plans.py
```diff
@@ -14,7 +14,7 @@ def scan(
     detectors: List[Readable],
     axes_to_move: Mapping[str, Movable],
     spec: Spec[str],
-    metadata: Optional[Mapping[str, Any]] = None,
+    md: Optional[Mapping[str, Any]] = None,
```
callumforrester, line 17: Would still prefer this is called `metadata` as it is presented to the user when they query for the plan arguments. These wrappers around `bps` and `bp` were written for two reasons:

1. add type annotations
2. make nicer argument structures with more descriptive names

DiamondJoseph: Done, I was going to argue that we don't have a count plan wrapping count, but we do. So I've made that also a little more in line

src/blueapi/plans/plans.py
```diff
@@ -93,5 +97,16 @@ def count(
     Yields:
         Iterator[MsgGenerator]: _description_
     """
+    plan_args = {
+        "detectors": list(map(repr, detectors)),
+        "num": num,
+    }
+    if delay:  # If bp.count added delay to plan_args, we could remove all md handling
+        plan_args.update({"delay": delay})
```
callumforrester, line 105: For optional plan args I'm in favour of allowing `"arg": null` in the JSON rather than maybe not having it there at all. It's generally easier to deserialize if the documents are all consistent for a given plan.

DiamondJoseph:
@callumforrester I want to move these plans into core-bluesky and import them to BlueAPI in its config. But I'd like this to get merged so that it is a direct move from one repo to the other without a minor refactor from "my head" along the way

callumforrester (approved):

PR #301, opened 2023-08-23
Extract plans to Core plan library

DiamondJoseph:
- Future plan libraries do not need to depend on BlueAPI
- Plans can be versioned/updated separately from BlueAPI
- Equivalent to dodal device module


DiamondJoseph:
For conversation:

Name of the plans module, given understandable voting down of core_bluesky, but analogous mx-specific plans in mx_bluesky.
- dls_bluesky_core (current working title: is dls but is also in the dls org; is bluesky stuff but is not in the bluesky org/repo; is "core" in that it is across beamlines/groups, contains functionality that is intended to be root functionality to be extended, is verbose)
- dls_bluesky_plans (not only plans? plans, stubs, and useful functionality around plans? But does putting non-plan things in it reduce the value of extracting it from BlueAPI? is still verbose)
- diamond_bluesky_core
- diamond_common_bluesky
- 
Organisation of plans and stubs: currently separated into plans/, stubs/ where previously were all in plans: plan_stubs are conceptually separate from plans in that they do not produce a Run, and bluesky makes the distinction, I think we should also, but then we have to import both modules into our BlueAPI instance: are we ever going to want just the plans or just the stubs? I can't think of a reason why we would, but having a single imported module of dls_bluesky_core and then in scripts `import dls_bluesky_core.plans as bp` may be more neat.

Organisation of plans into separate files: currently we have 1 plan that is a wrapped version of a Bluesky plan, and we may add to these plans, which are conceptually different from any plans that are "in house", and plan(s) that use ScanSpecs is also conceptually different in that the actual shape of the plan is extracted to the composing object. 
- plans/wrapped.py
- plans/scanspec.py
seems a sensible beginnings of organising these plans?

callumforrester (approved):

Issue #143, opened 2023-04-21
Expand PlanResponse to include Plan Args


DiamondJoseph:

callumforrester:
Continued in https://github.com/DiamondLightSource/blueapi/issues/368

PR #303, opened 2023-08-24
Return more information about plans when queried

### LINKED ISSUE #297: Give better introspection of plans
As a user the only current way I can find of listing plans is `blueapi controller plans`. This only gives me the name of the plan. It would be good to get (either in this call or something like `help my_plan` call):
* The types of the params it wants
* The docstring of the plan

DiamondJoseph:
- Closes #297, #143
- Returns information about call arguments when queried for plans

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -141,7 +141,9 @@ class BlueskyContext:
             __config__=BlueapiPlanModelConfig,
             **self._type_spec_for_function(plan),
         )
-        self.plans[plan.__name__] = Plan(name=plan.__name__, model=model, description=plan.__doc__)
+        self.plans[plan.__name__] = Plan(
+            name=plan.__name__, model=model, description=plan.__doc__
+        )
         self.plan_functions[plan.__name__] = plan
         return plan
 
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -141,7 +141,7 @@ class BlueskyContext:
             __config__=BlueapiPlanModelConfig,
             **self._type_spec_for_function(plan),
         )
-        self.plans[plan.__name__] = Plan(name=plan.__name__, model=model)
+        self.plans[plan.__name__] = Plan(name=plan.__name__, model=model, description=plan.__doc__)
         self.plan_functions[plan.__name__] = plan
         return plan
 
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -21,7 +21,7 @@ from typing import (
 
 from bluesky import RunEngine
 from pydantic import create_model
-from pydantic.fields import FieldInfo
+from pydantic.fields import FieldInfo, ModelField
 
 from blueapi.config import EnvironmentConfig, SourceKind
 from blueapi.utils import BlueapiPlanModelConfig, load_module_all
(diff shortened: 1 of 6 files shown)


DiamondJoseph:
For consideration: is this too verbose when e.g. calling /plans and getting all plans returned? 
/plans could return a PlanModel which is just the name as it currently is, and /plan/<foo> could return this more verbose struct
Should we not strip 'title' out of plan args, which makes the struct slightly longer with no benefit, but is a standard JSON field.

Is repr(type) as in detectors suitable?
```
"detectors": {
                "items": {"_detectors": "<class " "'bluesky.protocols.Readable'>"},
                "type": "array",
            },

```

src/blueapi/service/model.py
```diff
@@ -54,10 +54,15 @@ class PlanModel(BlueapiBaseModel):
     """
 
     name: str = Field(description="Name of the plan")
+    properties: Dict[str, Any] = Field(description="Plan arguments description")
+    required: List[str] = Field(description="Names of required arguments")
```
callumforrester, line 58: I think it would be better if these were nested, say inside another object called `PlanSchema`? Makes it easier to build client code that uses the schema if it's isolated in one place.

src/blueapi/service/model.py
```diff
@@ -54,10 +54,15 @@ class PlanModel(BlueapiBaseModel):
...
 
     @classmethod
     def from_plan(cls, plan: Plan) -> "PlanModel":
-        return cls(name=plan.name)
+        schema = plan.model.schema()
+        properties = schema.get("properties") or {}
+        required = schema.get("required") or []
+        return cls(name=plan.name, properties=properties, required=required)
```
callumforrester, line 65: To give some context: I had always hoped to unify `Plan` and `PlanModel`. The problem was that you could easily serialize `plan.schema` the way you've done it here but not so easily deserialize it back to a pydantic model. So the only options were to make it un-deserializable, make a new model without a schema model in it, or move the schema model somewhere else.

Don't think this is any more soluble than before, but just in case, got any ideas?

src/blueapi/core/context.py
```diff
@@ -198,6 +198,13 @@ def valid(cls, value):
                         raise ValueError(f"Device {value} is not of type {target}")
                     return val
 
+                @classmethod
+                def __modify_schema__(
+                    cls, field_schema: dict[str, Any], field: Optional[ModelField]
+                ):
+                    if field:
+                        field_schema.update({field.name: repr(target)})
+
```
callumforrester, line 207: Is this needed because pydantic just doesn't make a schema by default if it isn't there?

callumforrester (changes requested):
Looks good, just a few minor comments

callumforrester:
@DiamondJoseph Regarding API verbosity, I think it's fine for now. Can filter client-side where needed.

src/blueapi/service/model.py
keithralphs: Decided in meeting to use JSONSchema representation of the arguments under the `schema` key at the same level as the `name` key

src/blueapi/core/context.py
DiamondJoseph: Yes, with allow_arbitrary_types, and types that aren't primitives or Pydantic models, we have to provide a serialisation method

src/blueapi/service/model.py
```diff
@@ -54,10 +54,18 @@ class PlanModel(BlueapiBaseModel):
     """
 
     name: str = Field(description="Name of the plan")
+    description: Optional[str] = Field(description="Docstring of the plan")
+    argument_schema: dict[str, Any] = Field(
```
callumforrester, line 58: Just for the sake of consistency, we call them parameters rather than arguments throughout the blueapi codebase

callumforrester:
Happy to approve this once CI is happy

DiamondJoseph:
@callumforrester CI is happy (was just unhappy as argument_schema sorts above description, name while parameter_schema sorts below description, name)

callumforrester (approved):

PR #304, opened 2023-08-31
Allow passing kwargs to make_all_devices for dodal modules

Tom-Willemsen:
In particular, in hyperion it would be useful to be able to pass through `wait_for_connection` and `fake_with_ophyd_sim` - but keeping this generic to any `kwarg` (like `make_all_devices` itself) seems nicer than restricting it to just those two.


callumforrester (approved):
Agree with merging this, but TBD in next week's discussion. I would like to find something else for the long term.

PR #310, opened 2023-09-21
Fix CLI errors on handling events

callumforrester:
* Fix CLI errors when handling different types of event
* Support drawing progress bars in CLI
* Support drawing tables with best effort callback


DiamondJoseph (approved):
LGTM

PR #313, opened 2023-10-19
Add support for scratch area to automatically add packages when running on Kubernetes

callumforrester:
Recent discussion has emphasized the need for an area to checkout Python packages and do dev work against a live instance of blueapi running on Kubernetes. This has previously been done by pushing alpha tags of various packages and continually rebuilding and redeploying the blueapi container, which has become a lengthy and unweildy process.

This PR introduces the "scratch area", a shared directory that can be mounted from the container running on kubernetes, and can install any (pip installable) Python libraries that are cloned into it. In the cases where a dev version of an existing dependency is required (e.g. Dodal), the cloned version will overwrite the existing version installed in the container.
Removing a package and restarting the container works like a factory reset. See documentation addtions for more information.

Changes:
* Add confiugration for a scratch directory
* If configured, make blueapi pip install all packages in it on startup
* If configured, automatically create the scratch directory on startup
* Mount scratch in the helm chart if configured
* Tests and docs for new functionality

### DIFF
--- a/helm/blueapi/values.yaml
+++ b/helm/blueapi/values.yaml
@@ -70,7 +70,7 @@ affinity: {}
 
 scratch:
   {}
-  # hostPath: /dls_sw/p38/software/
+  # hostPath: /usr/local/blueapi-software-scratch
 
 worker:
   api:
--- a/helm/blueapi/values.yaml
+++ b/helm/blueapi/values.yaml
@@ -68,6 +68,10 @@ affinity: {}
 
 #existingSecret: see templates/secret.yaml
 
+scratch:
+  {}
+  # hostPath: /dls_sw/p38/software/
+
 worker:
   api:
     host: 0.0.0.0 # Allow non-loopback traffic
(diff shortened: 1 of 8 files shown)


helm/blueapi/values.yaml
```diff
@@ -68,6 +68,10 @@ affinity: {}
 
 #existingSecret: see templates/secret.yaml
 
+scratch:
+  {}
+  # hostPath: /dls_sw/p38/software/
```
callumforrester, line 73: Intend to leave this in as an example unless anyone can think of a better one

graeme-winter: Something non-diamond specific may be useful in the long game, e.g. where we deploy such software at a non-diamond location, or for third-party users (in the event that this is useful for others)

`p38` requires some rather internal knowledge to make sense of

graeme-winter: Also, if this really is a scratch area, perhaps 

```
/usr/local/blueapi-software-scratch
```

would be a more explicit example (appreciate that you make assumptions in the kubernetes environment about this being available on any host, but that can be arranged)

callumforrester: Good idea, chagned to that

DiamondJoseph (approved):
Seems to work for both existing library dependencies (dodal) and for modules to import (i22_bluesky)

PR #315, opened 2023-10-24
Use external service to configure ophyd-async detector data writing

callumforrester:
Production version of #283 

This PR allows coordination with a central service for creating unique groups of data (called collections) and configures ophyd-async detectors to write their data to the same location for a given collection. 

Changes:
- Add mechanism to preprocess all plans with set bluesky [preprocessors](https://blueskyproject.io/bluesky/plans.html#plan-preprocessors)
- Create directory provider that knows how to talk to GDA's visit directory API and provide a unique collection number to group data files
- Create dummy directory provider that works in a similar way without the need for an external server (useful for development)
- Create preprocessor that uses the directory provider and groups detectors by staging, also bundles the data group information into run start documents on a best effort basis.
- Add tests


src/blueapi/service/handler.py
```diff
@@ -80,7 +87,48 @@ def setup_handler(
...
+            )
+
+            set_directory_provider_singleton(provider)
+        except ImportError:
+            logging.error(
+                "Unable to set directory provider for ophyd-async devices, "
+                "a newer version of dodal is required"
+            )
+
+        plan_wrappers.append(lambda plan: attach_metadata(plan, provider))
```
rosesyrett, line 122: should we not also call `set_directory_provider_singleton(provider)` from `blueapi.data_management.gda_directory_provider` also? Here, we're only setting the directory provider in dodal.

src/blueapi/service/handler.py
```diff
@@ -80,7 +87,48 @@ def setup_handler(
...
+        try:
+            from dodal.parameters.gda_directory_provider import (
+                set_directory_provider_singleton,
+            )
+
+            set_directory_provider_singleton(provider)
+        except ImportError:
+            logging.error(
+                "Unable to set directory provider for ophyd-async devices, "
+                "a newer version of dodal is required"
```
rosesyrett, line 119: maybe this is too much but, should we specify here instead;

"Unable to set directory provider for ophyd-async devices, "
"most likely due to dodal not being installed from the directory_provider branch"

src/blueapi/data_management/gda_directory_provider.py
```diff
@@ -0,0 +1,145 @@
...
+    global _SINGLETON
+
+    _SINGLETON = provider
+
+
+def get_directory_provider() -> DirectoryProvider:
+    if _SINGLETON is not None:
+        return _SINGLETON
+    else:
+        return StaticDirectoryProvider(tempfile.TemporaryFile(), "")
```
rosesyrett, line 145: I can't seem to find any instances of this being used. It seems like when we start up blue api and a handler is made, in that big 'if' block that checks for config, we make a provider (if we have config) and then add attach_metadata to the plan wrappers with the provider we've just made. With that in mind, is there any point to making the directory provider in blue api a global variable?

This goes hand in hand with my earlier comment about set_directory_provider_singleton not actually being used.

so perhaps the best thing to do, is get rid of this block (from line 132 to 145) and don't fiddle with a global directory provider; just set the one in dodal.

rosesyrett (changes requested):
Minor changes requested. Looks good, thank you for the good tests as well.

I think we should remove the directory_provider as a global variable in blueapi, and just set it in dodal when we instantiate blueapi (as we are currently doing; we currently don't use the global variable)

src/blueapi/service/handler.py
callumforrester: Yeah I wondered about that, depends how soon that changes (i.e. that branch gets merged). I hope relatively soon.

src/blueapi/data_management/gda_directory_provider.py
callumforrester: You're right, this should be removed. Will put a fuller explanation in https://github.com/DiamondLightSource/blueapi/pull/315#discussion_r1371443992

src/blueapi/service/handler.py
callumforrester: This works because dodal is a library and blueapi is an application.

Blueapi runs in a Python runtime environment and imports dodal . That environment is the only "thing" in play. Blueapi and dodal essentially merge into one thing at runtime and that thing is called Python. That means we're not setting the global variable "in dodal", we're setting it in Python and it will be global for both dodal code and blueapi code. Does that make sense?

I appreciate that https://github.com/DiamondLightSource/blueapi/pull/315#discussion_r1371455708 may have caused some confusion because I had two global variables. We only need one and it needs to be in the dodal code because the dodal code is what depends on it.

src/blueapi/service/handler.py
rosesyrett: ah nevermind then, ignore me

src/blueapi/service/handler.py
rosesyrett: sure, that makes sense. Effectively, when we call `dodal.parameters.gda_directory_provider.set_directory_provider_singleton(provider)`, we set the global variable `_PROVIDER`. So, am I right in saying we don't need to define `_PROVIDER` in blueapi, we can just expect that at runtime (when we call this function) we will have a global variable, `_PROVIDER` (which is actually defined in dodal)?

rosesyrett (approved):

PR #321, opened 2023-10-25
Make devices connect with a timeout

callumforrester:


rosesyrett:
any tests for this will fail: see https://github.com/bluesky/ophyd-async/issues/39

Going to follow this up eventually by moving the stuff in ophyd_async_connect into ophyd_async.

PR #322, opened 2023-10-26
Fix dependency on dodal

callumforrester:
A dependency on a specific branch of dodal was mistakenly merged in #315. This PR reverts that dependency to use the release on PyPi.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
     "ophyd",
     "nslsii",
     "pyepics",
+    "aioca",
     "pydantic<2.0",
     "stomp.py",
     "aiohttp",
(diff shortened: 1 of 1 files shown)


DominicOram:
Could we get an automated way to confirm that dependencies are all pinned on release so this doesn't happen again?

callumforrester:
I may be misunderstanding, but given that you use blueapi as a library, wouldn't pinning everything on release hugely increase the chances of a dependency conflict somewhere?

DominicOram:
Sorry, typo. I mean nothing is pinned or it is pinned with a `>=`

callumforrester:
Definitely, although we will need to also get `dls-bluesky-core` unpinned first. It's stuck that way because it depends on unreleased bugfixes in bluesky itself. Feels like we need to find a better way to manage this sort of thing.

DominicOram (approved):

DominicOram:
Works for Hyperion but I'm aware that this means `blueapi` does not have the changes in https://github.com/DiamondLightSource/dodal/pull/193 now. Could another core developer comment on whether this is an issue, @rosesyrett maybe?

PR #324, opened 2023-10-27
Modified attach_metadata processor to only yield from, not yield

rosesyrett:
Relates to: https://github.com/bluesky/bluesky/issues/1625

### DIFF
--- a/src/blueapi/preprocessors/attach_metadata.py
+++ b/src/blueapi/preprocessors/attach_metadata.py
@@ -1,4 +1,5 @@
 import bluesky.plan_stubs as bps
+import bluesky.preprocessors as bpp
 from bluesky.utils import make_decorator
 
 from blueapi.core import MsgGenerator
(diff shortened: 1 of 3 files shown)


src/blueapi/preprocessors/attach_metadata.py
```diff
@@ -41,32 +42,10 @@ def attach_metadata(
...
-            # Handle the case where we're opening a run but no detectors
-            # have been staged yet. Common for nested runs.
-            if not group_in_progress:
-                yield from bps.wait_for([provider.update])
-            directory_info = provider()
-            message.kwargs[DATA_SESSION] = directory_info.filename_prefix
-
-        # This is a preprocessor so we yield the original message.
-        yield message
+    directory_info = provider()
```
callumforrester, line 45: As a result of no longer using the `update()` method, the type annotation of provider in the function signature can be changed to `DirectoryProvider`

src/blueapi/service/handler.py
```diff
@@ -92,18 +94,9 @@ def setup_handler(
...
-            )
-        else:
-            visit_service_client = LocalVisitServiceClient()
-
-        provider = VisitDirectoryProvider(
-            data_group_name=config.env.data_writing.group_name,
-            data_directory=config.env.data_writing.visit_directory,
-            client=visit_service_client,
+        provider = StaticDirectoryProvider(
+            filename_prefix="blueapi",
```
callumforrester, line 98: This could actually be `f"{config.env.data_writing.group_name}-blueapi"`, which would mean the files are called `<BEAMLINE>-blueapi-<UUID>.h5`, possibly slightly easier to interpret.

callumforrester (commented):
Looks good, suspect the tests need updating.

src/blueapi/service/handler.py
rosesyrett: Does it mean that?

When I've been testing this locally (i.e. with static directory provider) I don't recall the filenames of e.g. the hdf writer having uuid's in them. I think this will just produce a `<BEAMLINE>-blueapi.h5` file, which is fine for now.

callumforrester (approved):

PR #325, opened 2023-10-27
Reintroduce VisitDirectoryProvider and update in preprocessor

rosesyrett:


rosesyrett:
Have verified this works on p38. Unfortunately the filenames currently look like this:

BL38P-1d11-hdf.h5
BL38P-2d11-hdf.h5

which is a bit confusing and I think it actually originates from ophyd-async and how it propagates down the information from directory provider... I'll investigate but the easiest solution is probably a quick ophyd-async PR.

callumforrester (approved):

PR #328, opened 2023-11-01
Pin fastapi to <0.99 to use openapi <3.1

joeshannon:


rosesyrett (approved):
I think this looks good :+1:

PR #329, opened 2023-11-02
Add document listener

callumforrester:
Chagnes:

* Add CLI command to listen for documents from the message bus and print them to the console

### DIFF
--- a/helm/blueapi/templates/deployment.yaml
+++ b/helm/blueapi/templates/deployment.yaml
@@ -72,12 +72,8 @@ spec:
               value: {{ .Values.scratch.containerPath }}
         {{- if .Values.listener.enabled -}}
         - name: {{ .Chart.Name }}-document-listener
-          securityContext:
-            {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          resources:
-            {{- toYaml .Values.resources | nindent 12 }}
           args:
             - "-c"
             - "/config/config.yaml"
--- a/helm/blueapi/templates/deployment.yaml
+++ b/helm/blueapi/templates/deployment.yaml
@@ -70,6 +70,23 @@ spec:
           env:
             - name: SCRATCH_AREA
               value: {{ .Values.scratch.containerPath }}
+        {{- if .Values.listener.enabled -}}
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          resources:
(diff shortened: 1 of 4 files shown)


src/blueapi/cli/cli.py
```diff
@@ -125,6 +128,22 @@ def get_devices(obj: dict) -> None:
...
+    amq_client = AmqClient(StompMessagingTemplate.autoconfigured(config.stomp))
+
+    def on_event(context: MessageContext, event: Union[WorkerEvent, ProgressEvent, DataEvent],) -> None:
+        converted = json.dumps(event.dict(), indent=2)
+        print(converted)
+
+    print(f"Subscribing to all bluesky events from {config.stomp.host}:{config.stomp.port}")
+    with amq_client:
+        amq_client.subscribe_to_all_events(on_event)
+        input()
```
tpoliaw, line 146: Is it worth adding a message along the lines of 'Press enter to exit'?

tpoliaw (commented):
Looks good other than the lint complaints

src/blueapi/cli/cli.py
callumforrester: Wonderfully paradoxical sentence but yes

helm/blueapi/templates/deployment.yaml
```diff
@@ -70,6 +70,23 @@ spec:
           env:
             - name: SCRATCH_AREA
               value: {{ .Values.scratch.containerPath }}
+        {{- if .Values.listener.enabled -}}
+        - name: {{ .Chart.Name }}
```
callumforrester, line 74: Should be different to the other container
```suggestion
        - name: {{ .Chart.Name }}-document-listener
```

helm/blueapi/templates/deployment.yaml
```diff
@@ -70,6 +70,23 @@ spec:
...
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
```
callumforrester, line 76: I think this can be removed, the container requires no special access

helm/blueapi/templates/deployment.yaml
```diff
@@ -70,6 +70,23 @@ spec:
...
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
```
callumforrester, line 80: Resources should be much smaller than the resources for the blueapi main container

rosesyrett:
I don't particularly care that codecov is failing. Writing an integration test for this sort of thing is hard, because we need a test that simultaneously; 1. puts things on the message bus and 2.listens to it. This can be a ticket for later, but for now I'm happy to merge this.

rosesyrett (approved):

helm/blueapi/templates/deployment.yaml
rosesyrett: I'll just remove the resource requirements entirely and let k8s work it out. Do you think that's reasonable?

The alternative is including a listener.resources section in the values.yaml which I don't know if we should do since it's such a small process

rosesyrett (approved):

PR #330, opened 2023-11-07
 Allow passing kwargs to make_all_devices for dodal modules (again)

Tom-Willemsen:
Functionality added in https://github.com/DiamondLightSource/blueapi/pull/304 and [being used in hyperion](https://github.com/DiamondLightSource/hyperion/blob/92b23dac5efc1aa78910d2f0f4bd28725a9ec836/src/hyperion/utils/context.py#L81) got removed in https://github.com/DiamondLightSource/blueapi/pull/315/commits/076da45193d3f3460d0481d2bc1a814a6193bcc3 . It doesn't look like there was a clear reason for this removal of functionality so I think it should be reinstated.

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -138,10 +138,10 @@ class BlueskyContext:
     def with_device_module(self, module: ModuleType) -> None:
         self.with_dodal_module(module)
 
-    def with_dodal_module(self, module: ModuleType) -> None:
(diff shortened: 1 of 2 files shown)


callumforrester (approved):
For context this was removed as part of a large refactor in https://github.com/DiamondLightSource/blueapi/pull/315. As you say no clear reason is given, which is on us. From memory I believe I didn't realise that was in `main`, and thought it was part of a previous version of that PR.

PR #331, opened 2023-11-09
Add configuration option to disable status updates via message bus

callumforrester:
As well as data documents, blueapi produces events when the status objects monitored by the run engine are updated. These events are useful for creating progress bars and similar updates. Unfortunately it seems very easy to unintentionally  make plans/devices produce a very large number of these updates. The handling of all of these results in log spam and high CPU usage. 

We're seeing this now on I22 and have seen similar problems before (see #111). I think the easy way to make debugging easier is to make the status update handling optional and easy to turn off via config. To that end...

Changes:
- Add config option to disable status events
- Make the worker only hook into the run engine if this option is marked as true

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -50,6 +50,7 @@ class WorkerEventConfig(BlueapiBaseModel):
     """
     Config for event broadcasting via the message bus
     """
+
     broadcast_status_events: bool = True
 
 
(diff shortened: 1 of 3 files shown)


rosesyrett (approved):
Looks good; thanks for doing this. Am I alright to merge it in, make a release and try see if that fixes the problems we are seeing on i22?

PR #332, opened 2023-11-10
Minor fix to templating

rosesyrett:
In a previous PR, I had failed to test that my template for adding a listener actually worked. This is a minor change to fix the issue. I have verified the helm chart installs successfully with this change.


rosesyrett:
I'm going to be naughty and merge this without approvals as it's a small change, and I've verified that it works locally with building the helm chart.

Issue #220, opened 2023-05-19
Service does not gracefully start or shut down if activemq not available


rosesyrett:
## Overview
Found out by accident that if I kill my activemq container but forget to shut down the blueapi REST service, the blueapi terminal just spams messages. Upon further testing I also found out, if there is no activemq, you cannot start the blueapi service either (although at least it doesn't spam messages). Steps to reproduce both cases are included below.

### On shutdown

To reproduce, run the following with an activemq instance working:
```
blueapi serve
```
You will see something like:
```
INFO:     Started server process [731956]
INFO:     Waiting for application startup.
INFO:blueapi.worker.multithread:Setting up event loop
INFO:blueapi.messaging.stomptemplate:Connecting...
INFO:blueapi.worker.multithread:Worker starting
INFO:blueapi.worker.reworker:Worker starting
INFO:blueapi.worker.reworker:Awaiting task
INFO:stomp.py:established connection to host localhost, port 61613
INFO:     Application startup complete.
INFO:     Uvicorn running on http://localhost:8000 (Press CTRL+C to quit)
```
Then terminate your activemq instance. You will see the following:
```
WARNING:blueapi.messaging.stomptemplate:Stomp connection lost, will attempt reconnection with policy StompReconnectPolicy(initial_delay=0.0, attempt_period=10.0)
INFO:blueapi.messaging.stomptemplate:Connecting...
INFO:stomp.py:established connection to host localhost, port 61613
ERROR:blueapi.messaging.stomptemplate:Reconnect failed
Traceback (most recent call last):
  File "/home/rose/Documents/projects/blueapi/src/blueapi/messaging/stomptemplate.py", line 194, in _on_disconnected
    self.connect()
  File "/home/rose/Documents/projects/blueapi/src/blueapi/messaging/stomptemplate.py", line 160, in connect
    self._conn.connect(wait=True)
  File "/venv/lib/python3.11/site-packages/stomp/connect.py", line 151, in connect
    Protocol11.connect(self, *args, **kwargs)
  File "/venv/lib/python3.11/site-packages/stomp/protocol.py", line 339, in connect
    self.transport.wait_for_connection()
  File "/venv/lib/python3.11/site-packages/stomp/transport.py", line 318, in wait_for_connection
    raise exception.ConnectFailedException()
stomp.exception.ConnectFailedException
INFO:blueapi.messaging.stomptemplate:Connecting...
WARNING:stomp.py:could not connect to host localhost, port 61613
```

[...]

PR #335, opened 2023-11-10
Fix shutdown

abbiemery:
Relates to #220 . This change is to enable blueapi to be shutdown with crtl+c when the activemq drops and it can no longer connect to it.

### DIFF
--- a/src/blueapi/service/handler.py
+++ b/src/blueapi/service/handler.py
@@ -80,7 +80,8 @@ class Handler:
 
     def stop(self) -> None:
         self.worker.stop()
-        self.messaging_template.disconnect()
+        if self.messaging_template.is_connected():
+            self.messaging_template.disconnect()
 
 
 HANDLER: Optional[Handler] = None
--- a/tests/messaging/test_stomptemplate.py
+++ b/tests/messaging/test_stomptemplate.py
@@ -154,7 +154,7 @@ def test_reconnect(template: MessagingTemplate, test_queue: str) -> None:
     template.disconnect()
     connected = template.is_connected()
     assert connected is False
-    
+
     template.connect()
     connected = template.is_connected()
     assert connected is True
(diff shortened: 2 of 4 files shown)


callumforrester (commented):
Looks good! Could do with a test or two

abbiemery:
I've added in a small check using the is_connected method. However, I am going to change up the stomp connection part slightly in another PR so it doesn't crash if you try start BlueApi without an activemq instance running. The tests will be changed/added too acordingly there.

tests/messaging/test_stomptemplate.py
```diff
@@ -152,7 +152,12 @@ def test_reconnect(template: MessagingTemplate, test_queue: str) -> None:
     reply = template.send_and_receive(test_queue, "test", str).result(timeout=_TIMEOUT)
     assert reply == "ack"
     template.disconnect()
+    connected = template.is_connected()
+    assert connected is False
+
     template.connect()
+    connected = template.is_connected()
+    assert connected is True
```
callumforrester, line 160: Slightly less wordy version

```suggestion
    assert not template.is_connected()
    template.connect()
    assert template.is_connected()
```

src/blueapi/service/handler.py
```diff
@@ -80,7 +80,8 @@ def _publish_event_stream(self, stream: EventStream, destination: str) -> None:
 
     def stop(self) -> None:
         self.worker.stop()
-        self.messaging_template.disconnect()
+        if self.messaging_template.is_connected():
+            self.messaging_template.disconnect()
```
callumforrester, line 84: Hmm, I definitely intended `MessagingTemplate.disconnect` to be idempotent. If it's not that's probably a bug in itself, worth raising an issue over

callumforrester (approved):
Looks good and should be merged if it fixes a common failure case.
See minor nitpicks

src/blueapi/service/handler.py
abbiemery: This was pretty much the crux of the issue. It was trying to disconnect something that was already disconnected and was raising exceptions.

abbiemery: https://github.com/DiamondLightSource/blueapi/issues/340

callumforrester: Hmm, I _think_ I'd prefer it be made to only disconnect if connected, similar to this. What do you think?

PR #337, opened 2023-11-15
Add event type filtering to listener command

callumforrester:


src/blueapi/config.py
```diff
@@ -71,8 +71,21 @@ class EnvironmentConfig(BlueapiBaseModel):
...
+    cli: LogLevel = "ERROR"
+
+
 class LoggingConfig(BlueapiBaseModel):
-    level: LogLevel = "INFO"
+    """
+    Config for how blueapi logs behave
+    """
+
+    level: LoggingLevelConfig = Field(default_factory=LoggingLevelConfig)
```
callumforrester, line 88: I don't think this is needed any more, the CLI and server just have entirely separate configurations, so do not need a shared configuration with separate elements.

callumforrester:
I think this PR badly needs to be rebased on top of `main`!

callumforrester:
Very out of date, probably superseded by the plans for storing the documents in a queryable database

PR #341, opened 2023-11-28
Remove example config

callumforrester:
Remove example configuration because it was confusing; people thought it was used in production somewhere rather than just an example. Replaced with an easy mechanism of viewing the config via the command line.

Changes:
- Delete example config directory and it's validation test
- Add CLI command to print the current config
- Refactor yaml printing code to be common between printing schema and config


DiamondJoseph:
I think the adsim config could be useful, even if the others belong in their Athena deployment chart.

callumforrester:
No longer relevant

Issue #317, opened 2023-10-25
Run worker in subprocess


joeshannon:

stan-dot:
#399 is relevant (?)

callumforrester:
@stan-dot very relevant, it's the next step. Also @joeshannon can this be closed now #343 is merged?

joeshannon:
Yes I think so, there is #365 to do still, but that's its own ticket.

PR #342, opened 2023-11-30
Extract handler methods to ABC

joeshannon:
Refactor in preparation for #317

### DIFF
--- a/tests/service/test_rest_api.py
+++ b/tests/service/test_rest_api.py
@@ -23,7 +23,7 @@ def test_get_plans(handler: Handler, client: TestClient) -> None:
 
     plan = Plan(name="my-plan", model=MyModel)
 
-    handler.context.plans = {"my-plan": plan}
+    handler._context.plans = {"my-plan": plan}
     response = client.get("/plans")
 
     assert response.status_code == status.HTTP_200_OK
@@ -49,7 +49,7 @@ def test_get_plan_by_name(handler: Handler, client: TestClient) -> None:
 
     plan = Plan(name="my-plan", model=MyModel)
 
-    handler.context.plans = {"my-plan": plan}
+    handler._context.plans = {"my-plan": plan}
     response = client.get("/plans/my-plan")
 
     assert response.status_code == status.HTTP_200_OK
@@ -159,7 +159,7 @@ def test_get_devices(handler: Handler, client: TestClient) -> None:
 
     device = MyDevice("my-device")
 
-    handler.context.devices = {"my-device": device}
+    handler._context.devices = {"my-device": device}
     response = client.get("/devices")
 
     assert response.status_code == status.HTTP_200_OK
@@ -180,7 +180,7 @@ def test_get_device_by_name(handler: Handler, client: TestClient) -> None:
 
     device = MyDevice("my-device")
 
-    handler.context.devices = {"my-device": device}
+    handler._context.devices = {"my-device": device}
     response = client.get("/devices/my-device")
 
     assert response.status_code == status.HTTP_200_OK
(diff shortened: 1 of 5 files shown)


tests/service/test_rest_api.py
```diff
@@ -23,7 +23,7 @@ class MyModel(BaseModel):
 
     plan = Plan(name="my-plan", model=MyModel)
 
-    handler.context.plans = {"my-plan": plan}
+    handler._context.plans = {"my-plan": plan}
```
rosesyrett, line 26: I know I made this change, which makes this comment seem hypocritical; but I think tests shouldn't use the underlying ._context and other _ attributes if they can avoid it. So this is just a comment to say; please review everywhere where tests do something like handler._ ... and see if they can't use the facade methods instead.

That's all. Otherwise looking good.

rosesyrett (approved):
lgtm

tests/service/test_rest_api.py
joeshannon: Thanks - yes I was wondering about this too and I do agree.
I'm a bit unsure how to actually achieve this though. E.g. how to specify specific plans and devices used in the tests without accessing these fields directly. In the normal code flow they are loaded based on the `ApplicationConfig` but the tests at the moment don't concern themselves with that.

src/blueapi/service/facade.py
```diff
@@ -0,0 +1,73 @@
+from abc import ABC, abstractmethod
+from typing import List, Optional
+
+from blueapi.service.model import DeviceModel, PlanModel, WorkerTask
+from blueapi.worker.event import WorkerState
+from blueapi.worker.task import RunPlan
+from blueapi.worker.worker import TrackableTask
+
+
+class BlueskyContextFacade(ABC):
```
joeshannon, line 10: I'm not sure about this name now either as the methods here are related to both the `BlueskyContext` and the `Worker`.

joeshannon: Have changed name.

src/blueapi/service/facade.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+
+    @abstractmethod
+    def submit_task(self, task: RunPlan) -> str:
+        """
+        Submit a task to be run on begin_task
+        """
+
+    @abstractmethod
+    def delete_task(self, task_id: str) -> str:
```
joeshannon, line 46: In the worker this is called "clear" rather than "delete" (delete comes from REST api).

Is "clear" better?

src/blueapi/service/facade.py
```diff
@@ -0,0 +1,85 @@
...
+from blueapi.service.model import DeviceModel, PlanModel, WorkerTask
+from blueapi.worker.event import WorkerState
+from blueapi.worker.task import RunPlan
+from blueapi.worker.worker import TrackableTask
+
+
+class BlueskyHandler(ABC):
+    """Interface between web application and underlying Bluesky context and worker"""
+
+    @property
```
joeshannon, line 13: There is a mix of properties and getters now, is this cleaner overall or more confusing?

tpoliaw: I don't think it's too bad

joeshannon:
The filename should probably be changed too (currently `facade.py`).

tests/service/test_rest_api.py
callumforrester: Agree with Rose, is there a way we can restructure this test?

src/blueapi/service/facade.py
joeshannon: This only applies to pending tasks. To cancel the active task, `cancel_active_task` is used. So perhaps `clear_pending_task` to remove ambiguity?

tests/service/test_rest_api.py
joeshannon: Due to the complexity of the test changes required I think we should handle this in another PR/issue.

The tests will need to change with the next PR (that actually adds the subprocess) too.

PR #343, opened 2023-12-07
Run BlueskyContext and Worker in subprocess

joeshannon:
This approach is possibly not the best design as it results in four definitions of all the functions.

### DIFF
--- a/src/blueapi/service/handler.py
+++ b/src/blueapi/service/handler.py
@@ -145,7 +145,6 @@ def setup_handler(
 
     provider = None
     plan_wrappers = []
-
     if config:
         visit_service_client: VisitServiceClientBase
         if config.env.data_writing.visit_service_url is not None:
--- a/src/blueapi/service/handler.py
+++ b/src/blueapi/service/handler.py
@@ -28,6 +28,7 @@ class Handler(BlueskyHandler):
     _worker: Worker
     _config: ApplicationConfig
     _messaging_template: MessagingTemplate
+    _initialized: bool = False
 
     def __init__(
         self,
@@ -64,6 +65,7 @@ class Handler(BlueskyHandler):
         )
 
         self._messaging_template.connect()
+        self._initialized = True
 
     def _publish_event_streams(
         self, streams_to_destinations: Mapping[EventStream, str]
@@ -79,6 +81,7 @@ class Handler(BlueskyHandler):
         )
 
     def stop(self) -> None:
+        self._initialized = False
         self._worker.stop()
         if self._messaging_template.is_connected():
             self._messaging_template.disconnect()
@@ -133,6 +136,10 @@ class Handler(BlueskyHandler):
 
     def get_pending_task(self, task_id: str) -> Optional[TrackableTask]:
         return self._worker.get_pending_task(task_id)
+
+    @property
+    def initialized(self) -> bool:
+        return self._initialized
 
 
 HANDLER: Optional[Handler] = None
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -1,5 +1,5 @@
 from contextlib import asynccontextmanager
-from typing import Dict, Set
+from typing import Dict, Optional, Set
 
 from fastapi import Body, Depends, FastAPI, HTTPException, Request, Response, status
 from pydantic import ValidationError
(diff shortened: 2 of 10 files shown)


joeshannon:
These changes do allow the run engine and worker to run in a subprocess but I'm not sure this is the best design as it ends up with many implementations of these methods.

I've tested it though and it does work.

Probably something to discuss next week.

src/blueapi/service/main.py
```diff
@@ -49,6 +73,13 @@ async def on_key_error_404(_: Request, __: KeyError):
     )
 
 
+@app.put("/reload")
+def update_task(handler: BlueskyHandler = Depends(get_handler)):
```
callumforrester, line 77: This seems to cause a silent crash after 1-5 reloads (random), can you reproduce?

callumforrester: From discussion:
- Could switch subprocess to spawning rather than forking

src/blueapi/service/subprocess_handler.py
```diff
@@ -0,0 +1,144 @@
...
+    def start(self):
+        if self._subprocess is not None:
+            raise ValueError("Subprocess already running")
+        self._subprocess = Pool(processes=1)
+        self._subprocess.apply(setup_handler, [self._config])
+
+    def stop(self):
+        self._subprocess.apply(teardown_handler)
+        self._subprocess.terminate()
+        self._subprocess = None
```
callumforrester, line 34: These should be idempotent: If it's started then `start()` should do nothing, ditto for stop

callumforrester (commented):
Looks good, needs tests. See comment about intermittent issue.

I think after this we can clean up `subprocess_handler.py` a bit more and break up handler as discusses

src/blueapi/service/model.py
```diff
@@ -128,3 +128,21 @@ class StateChangeRequest(BlueapiBaseModel):
         description="The reason for the current run to be aborted",
         default=None,
     )
+
+
+class EnvironmentResponse(BlueapiBaseModel):
```
joeshannon, line 133: Naming of these endpoints, responses and parameters to be determined.

src/blueapi/service/handler.py
```diff
@@ -145,7 +152,7 @@ def setup_handler(
 
     provider = None
     plan_wrappers = []
-
+    print("Setting up handler")
```
joeshannon, line 155: Remove

src/blueapi/service/main.py
```diff
@@ -49,6 +84,29 @@ async def on_key_error_404(_: Request, __: KeyError):
...
+@app.put("/environment", response_model=EnvironmentRequest)
+async def set_environment(
+    state: EnvironmentRequest,
+    background_tasks: BackgroundTasks,
+    handler: BlueskyHandler = Depends(get_handler),
+) -> EnvironmentRequest:
+    def restart_handler(handler: BlueskyHandler):
+        handler.stop()
+        handler.start()
+        print("Reloaded")
```
joeshannon, line 103: Remove/convert to logging

joeshannon (commented):
Probably needs some tests of the subprocess handler.

src/blueapi/service/main.py
```diff
@@ -49,6 +73,13 @@ async def on_key_error_404(_: Request, __: KeyError):
     )
 
 
+@app.put("/reload")
```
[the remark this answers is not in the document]
joeshannon: Done, subject to naming choices.

callumforrester:
Would merging this close https://github.com/DiamondLightSource/blueapi/issues/317?

callumforrester:
I think the four definitions of functions flaw can be solved by a refactor in the future, this is a good first attempt.

joeshannon:
> Would merging this close #317?

Yes, is you commenting of that sufficient to auto close the issue when this is merged?

src/blueapi/service/main.py
joeshannon: I haven't been able to reproduce the reloading issue.

Have switched to spawn now.

src/blueapi/service/main.py
joeshannon: Done

callumforrester (approved):
Please could you make issues to do the following (unless you feel like doing them now but I think we should get this merged):

- Document this approach, probably requires an explanation page and perhaps an ADR
- Add CLI commands to inspect/restart the environment

joeshannon:
> Please could you make issues to do the following (unless you feel like doing them now but I think we should get this merged)

Yes I agree - it's probably best to draw a line under this and make new issues.

PR #346, opened 2024-01-15
Added more comprehensive docstrings to values, using helm-docs tool

rosesyrett:
Added extra descriptions in the helm documentation for the values.yaml file. Auto generated a README in the helm repository containing this information in a table.

Perhaps we should setup a CI job to do this, as part of this PR? I think this should also be done for the nexus-filewriting-service.


rosesyrett:
Note: on github currently the README looks a little squished and I'm not sure the best way of fixing this... https://github.com/DiamondLightSource/blueapi/tree/helm-docs/helm/blueapi

DiamondJoseph:
For a lot of these, it may be worth re-running helm create and seeing what comments are present, stick with what is readily available if sufficient and not roll our own.

Issue #360, opened 2024-02-05
Remove all private member access from tests


callumforrester:
There is some private member access (a.k.a `foo._bar`) going on in the test that has made it more difficult for us to migrate and refactor code. Private member access is discouraged and should be removed. Tests should be refactored to work without it. If they cannot work without it, the code they test should be refactored until they can.

PR #370, opened 2024-02-07
Remove parameter caching inside task class

### LINKED ISSUE #369: Error querying worker task via REST API
When querying `/worker/task` via the API, the server returns 500 internal server error.

This is because the currently active task object holds a cached copy of the dynamically generated pydantic model of the plan's parameters. See `blueapi.worker.task.Task`. Pickling does not work on dynamically generated classes by default.

There are a couple of options for solving:
1. Make the dynamically generated classes pickable. It seems to do so we need to define a `__reduce__` dunder for them: https://docs.python.org/3/library/pickle.html#object.__reduce__
2. Cache the dynamically generated class somewhere else

I'm in favour of 2 since `Task` was only ever supposed to be a data class

callumforrester:
Closes #369

The validated and processed parameters from a plan request were being cached inside the `Task` class because they were needed in two separate threads. Unfortunately, with the introduction of the subprocess (#343), pickling was causing issues with the cached object (see #369 for more details).

This PR is the simplest solution: remove the cache and generate the parameters twice. It also adds regression tests for the bug case in #369

### DIFF
--- a/src/blueapi/worker/reworker.py
+++ b/src/blueapi/worker/reworker.py
@@ -129,7 +129,7 @@ class RunEngineWorker(Worker[Task]):
             raise KeyError(f"No pending task with ID {task_id}")
 
     def submit_task(self, task: Task) -> str:
-        task.prepare_params(self._ctx)
+        task.prepare_params(self._ctx)  # Will raise if parameters are invalid
         task_id: str = str(uuid.uuid4())
         trackable_task = TrackableTask(task_id=task_id, task=task)
         self._pending_tasks[task_id] = trackable_task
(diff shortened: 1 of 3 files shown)


joeshannon (approved):
This change look good, thanks!

My only note is that the new tests currently pass on `main` even without the associated changes to task/reworker.

This is because `test_rest_api.py` is only using the "internal" non-subprocess handler, and it highlights that we probably need to either convert it (relates partially to #360), or introduce some system tests.

callumforrester:
Yeah I think #360 should go towards solving that, it's useful to have the test regardless

PR #375, opened 2024-02-21
Restore tests and stubs previously moved to dls-bluesky-core

DiamondJoseph:
- Wrapped Bluesky plans only required for BlueAPI, able to expose through new scratch dir

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
     "fastapi[all]<0.99",
     "uvicorn",
     "requests",
-    "dls-bluesky-core", #requires ophyd-async
     "dls-dodal",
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -26,7 +26,6 @@ dependencies = [
     "fastapi[all]<0.99",
     "uvicorn",
     "requests",
```
abbiemery, line 28: There appear to be some references left to this removed package that are failing the tests.

abbiemery (changes requested):
Seams sensible, although i'm not up on the history behind the choices. Just remove the old references and make sure the CI passes.

callumforrester:
These are now in dodal

Issue #353, opened 2024-01-25
Move preprocessors to dodal


callumforrester:
There is currently a `preprocessors` package in blueapi which holds [plan preprocessors](https://blueskyproject.io/bluesky/plans.html#plan-preprocessors) that can be applied to all plans that blueapi runs. It doesn't really belong here, it should be an optional plugin.

The preprocessor code should live in dodal. There should be a way of optionally ingesting preprocessors on startup and applying them to all plans.

callumforrester:
@DominicOram Opinions welcome

DominicOram:
We don't really use `preprocessors` at the moment so I don't really have any strong opinions. I agree that code that is actually logic of what to do on the beamline (which I assume is what your `preprocessors` are) should live in `dodal`

callumforrester:
@joeshannon and I have been looking at this. Are there any plans currently in dodal we could look at @DominicOram? In particular we're wondering where beamline-specific configuration for general plans should go (rough equivalent of GDA properties).

callumforrester:
@DiamondJoseph correct me if I'm wrong but this is now done

PR #376, opened 2024-02-21
Adjust to new DirectoryProvider structure

DiamondJoseph:
- Backs off Directory that is provided, allows a hardcoded root to get back to where it was
Allows us to unpin our dependency on dodal (through it, ophyd-async)


src/blueapi/config.py
```diff
@@ -41,9 +41,22 @@ class StompConfig(BaseModel):
...
+    Args:
+        visit_service_url (Optional[str]): the /api endpoint of a visit_service
+        beamline (str): beamline identifier to use in constructing the name of files
+        root (Optional[str]): root to use when constructing StreamResources.
+            When visits are part of the response from the visit_service,
+            this will be deprecated and replaced with a method that constructs
+            it (e.g. return f"/{year}/{visit_id})
+    """
+
+    beamline: str = "example"
```
callumforrester, line 55: Very small nitpick but I don't like having the word "beamline" in blueapi, encourages keeping it generically deployable (e.g. in a CryoEM or a small lab setup)

callumforrester:
Noting that this will need to be moved: https://github.com/DiamondLightSource/blueapi/issues/353

src/blueapi/preprocessors/attach_metadata.py
```diff
@@ -32,10 +33,9 @@ def attach_metadata(
         Iterator[Msg]: Plan messages
     """
     yield from bps.wait_for([provider.update])
-    directory_info = provider()
-    yield from bpp.inject_md_wrapper(
-        plan, md={DATA_SESSION: directory_info.filename_prefix}
-    )
+    directory_info: DirectoryInfo = provider()
+    yield from bpp.inject_md_wrapper(plan, md={DATA_SESSION: directory_info.prefix})
+    provider.clear()
```
callumforrester, line 38: Won't this mean that detectors subsequently calling `provider()` will try to write to `None`?

callumforrester (commented):
Build not passing

src/blueapi/preprocessors/attach_metadata.py
DiamondJoseph: Yes, I think the issue is actually that update() should always fetch a new instance rather than only the first time.

callumforrester: It was intended to be a caching thing, where `update()` would generate a brand new scan number and save location, so you would end up with this:

image

If it fetches a new instance every time, wouldn't the detectors each write to a different scan number, which would be different in turn from the resource doc?

src/blueapi/config.py
DiamondJoseph: prefix=f"{self._beamline}-{collection_id}"

Is already so Diamond specific whether it's _beamline or_data_group_name....

And when the pre-processor is moved to dodal surely it's being more Diamond specific there?

I think DirectoryProvider should define an abstract "generate_metadata" and just call that, dodal implementation just calls its own get_directory_provider, and then any implementations (like how BlueAPI is acting prior to moving pre-processor to Dodal) is creating and setting the implementation of Directory provider.

src/blueapi/preprocessors/attach_metadata.py
DiamondJoseph: Yes, you're right (and update() doesn't hold onto the previous one as I thought it did in the previous comment), I was confused by the change to no longer swallow the Exception in update() [as required merge into Bluesky had occured], which previously made sure the prior collection was None but no longer does! :zap:

DiamondJoseph:
IMPORTANT: Must also now unpin dependency on dodal

DominicOram (dismissed):
As it is this is fine, all the code seems to be replaced in https://github.com/DiamondLightSource/dodal/pull/445. Except there are tests in this PR that have no correspondent in `dodal`. Could you make a PR on `dodal` that moves the appropriate tests over?

DiamondJoseph:
> As it is this is fine, all the code seems to be replaced in [DiamondLightSource/dodal#445](https://github.com/DiamondLightSource/dodal/pull/445). Except there are tests in this PR that have no correspondent in `dodal`. Could you make a PR on `dodal` that moves the appropriate tests over?

I've been working on it today: the FakeDetector structure wasn't correct for how a StandardDetector actually uses a DirectoryProvider

callumforrester (approved):

PR #382, opened 2024-03-06
remove RunPlan class that wraps Task

stan-dot:
quick renaming 36 refs and delete the class

### DIFF
--- a/docs/user/reference/openapi.yaml
+++ b/docs/user/reference/openapi.yaml
@@ -15,8 +15,8 @@ components:
           title: Protocols
           type: array
       required:
-        - name
-        - protocols
+      - name
+      - protocols
       title: DeviceModel
       type: object
     DeviceResponse:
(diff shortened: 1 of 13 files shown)


src/blueapi/worker/task.py
```diff
@@ -32,16 +32,6 @@ def do_task(self, ctx: BlueskyContext) -> None:
         ctx.run_engine(wrapped_plan_generator)
 
 
-# Here for backward compatibility pending
-# https://github.com/DiamondLightSource/blueapi/issues/253
-class RunPlan(Task):
```
DiamondJoseph, line 37: Is this still required for backwards compatibility? Is it referenced anywhere?

docs/user/reference/openapi.yaml
DiamondJoseph, line 1: How did you generate this file? It'd be good to remove all the unrelated formatting only changes.

stan-dot: I think I just ran ruff / pre-commit automatically

src/blueapi/worker/task.py
stan-dot: to be honest RunEngineWorker shares a name, but might be renamed into TaskWorker

stan-dot: renamed

docs/user/reference/openapi.yaml
stan-dot: now I auto-generated a new one using 
`python -m blueapi.cli schema -u`.
formatting indeed has changed

Issue #233, opened 2023-05-24
Pending tasks are not cleared


callumforrester:
Tasks are currently only removed from pending_tasks when clear_task is called on it, meaning pending_tasks contain tasks that are not is_pending, and various states of is_complete/is_error.
I think we need pending_tasks, active_task, finished_tasks eventually.

callumforrester:
Apologies for past-me's unclear rambling, but I _think_ this means we should rename `worker.pending_tasks` to simply `worker.tasks`, since not all of them are, in fact, pending (some are already done, for example)

PR #386, opened 2024-03-12
233 pending tasks are not cleared

stan-dot:
closes issue #233

### DIFF
--- a/src/blueapi/utils/ophyd_async_connect.py
+++ b/src/blueapi/utils/ophyd_async_connect.py
@@ -30,7 +30,7 @@ async def _wait_for_tasks(tasks: Dict[asyncio.Task, str], timeout: float):
             t.cancel()
             with suppress(Exception):
                 await t
-            msg += format_error_message(tasks, t)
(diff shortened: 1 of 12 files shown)


src/blueapi/utils/ophyd_async_connect.py
```diff
@@ -45,3 +39,14 @@ async def _wait_for_tasks(tasks: Dict[asyncio.Task, str], timeout: float):
             logging.exception(f"  {tasks[t]}:", exc_info=t.exception())
     if pending or raised:
         raise NotConnected("Not all Devices connected")
+
+
+def format_error_message(tasks: Dict[asyncio.Task, str], t: asyncio.Task) -> str:
```
callumforrester, line 44: I think this should have a more descriptive name or be prefixed with an `_`

callumforrester (approved):

src/blueapi/utils/ophyd_async_connect.py
stan-dot: doing the `_` option now

PR #400, opened 2024-03-25
Update catalog-info.yaml

keithralphs:
Change owner to core data acq group and add Athena system reference


callumforrester (approved):
LGTM

Issue #206, opened 2023-05-16
Unclear error message when running plan without parameters


joeshannon:
When the controller run subcommand is executed without providing any parameters (`-p`), the following is returned:
```
$ blueapi controller run scan
Traceback (most recent call last):
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/bin/blueapi", line 8, in <module>
    sys.exit(main())
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/scratch/30day_tmp/delete15may/bs/blueapi/src/blueapi/cli/cli.py", line 67, in wrapper
    func(*args, **kwargs)
  File "/scratch/30day_tmp/delete15may/bs/blueapi/venv/lib/python3.10/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/scratch/30day_tmp/delete15may/bs/blueapi/src/blueapi/cli/cli.py", line 106, in run_plan
    json=json.loads(parameters),
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/json/__init__.py", line 339, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not NoneType
```

[...]

PR #409, opened 2024-04-02
366 add cli commands to inspectrestart the environment

stan-dot:


callumforrester (changes requested):

src/blueapi/cli/rest.py
```diff
@@ -113,3 +117,11 @@ def _request_and_deserialize(
     def _url(self, suffix: str) -> str:
         base_url = f"{self._config.protocol}://{self._config.host}:{self._config.port}"
         return f"{base_url}{suffix}"
+
+    def get_environment(self) -> EnvironmentResponse:
+        return self._request_and_deserialize("/environment", EnvironmentResponse)
+
+    def reload_environemnt(self) -> EnvironmentResponse:
```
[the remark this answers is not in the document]
stan-dot: disagree. if I am a user and do 'delete env' and then 'get env' I'd expect to see empty environment. That is not what happens. If after that I get some environment, I'd suspect that the 'delete' command didn't quite work.

REST does not support a 'reload' verb and that we use it over PUT or PATCH is arbitrary.

src/blueapi/cli/cli.py
```diff
@@ -258,6 +258,29 @@ def stop(obj: dict) -> None:
...
+)
+@click.pass_obj
+def env(obj: dict, reload: Optional[bool]) -> None:
+    """
+    Inspect or restart the environment
+    """
+
+    client: BlueapiRestClient = obj["rest_client"]
+    if reload:
+        print("reloading the environment...")
```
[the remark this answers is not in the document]
stan-dot: done

.vscode/settings.json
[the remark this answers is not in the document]
stan-dot: the only difference is explicit vs true, which is due to deprecation

src/blueapi/cli/cli.py
```diff
@@ -258,6 +258,29 @@ def stop(obj: dict) -> None:
...
+        pprint(client.reload_environemnt())
```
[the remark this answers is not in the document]
stan-dot: how do we know that it is complete from the server side?

I don't like the EnvironmentResponse. Initialized != ready. potential response could be [425 too early](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/425) or to make the pydantic response a string enum 


image

src/blueapi/cli/cli.py
```diff
@@ -258,6 +259,51 @@ def stop(obj: dict) -> None:
...
+        # Fetch the current environment status
+        environment_status = client.get_environment()
+
+        # Check if the environment is initialized
+        if environment_status.initialized:
+            print("Environment is initialized.")
+            environment_initialized = True
+        else:
+            print("Waiting for environment to initialize...")
+            sleep(5)  # Wait for 5 seconds before checking again
```
callumforrester, line 301: I think 5 seconds is a long time, would rather check every 1 second 10 times, say, and then raise a `TimeoutError`

stan-dot: ok

src/blueapi/cli/rest.py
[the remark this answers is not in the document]
callumforrester: PATCH may be a better verb long-term, but that's a separate ticket

stan-dot:
> src/blueapi/service/main.py:145:22: B008 Do not perform function call `Task` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable
tests/worker/test_reworker.py:349:18: B008 Do not perform function call `Task` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable

not sure what convention do we want to settle on here @callumforrester

tests/worker/test_reworker.py
```diff
@@ -346,7 +346,9 @@ def test_worker_and_data_events_produce_in_order(worker: Worker) -> None:
 def assert_running_count_plan_produces_ordered_worker_and_data_events(
     expected_events: list[WorkerEvent | DataEvent],
     worker: Worker,
-    task: Task = Task(name="count", params={"detectors": ["image_det"], "num": 1}),  # noqa: B008
+    task: Task = Task(
+        name="count", params={"detectors": ["image_det"], "num": 1}
+    ),  # noqa: B008
```
DiamondJoseph, line 351: Remove the noqa, make task: Task | None = None

Then in the body of the assert, task = task or Task(name="count", params={"detectors": ["image_det"], "num": 1})

DiamondJoseph: As is, task is instantiated when the method is first read, then every time it's called without an argument, the same task is being passed around, modified, run.

src/blueapi/service/main.py
```diff
@@ -140,7 +141,9 @@ def get_device_by_name(name: str, handler: BlueskyHandler = Depends(get_handler)
 def submit_task(
     request: Request,
     response: Response,
-    task: Task = Body(..., example=Task(name="count", params={"detectors": ["x"]})),  # noqa: B008
+    task: Task = Body(
+        ..., example=Task(name="count", params={"detectors": ["x"]})
+    ),  # noqa: B008
```
DiamondJoseph, line 146: as below

DiamondJoseph (changes requested):

.vscode/launch.json
callumforrester, line 1: This is no longer necessary following #438, please revert

.vscode/settings.json
callumforrester, line 1: Please could this be kept in line with the copier template?

src/blueapi/cli/rest.py
```diff
@@ -103,14 +104,27 @@ def _request_and_deserialize(
         data: Mapping[str, Any] | None = None,
         method="GET",
         raise_if: Callable[[requests.Response], bool] = _is_exception,
-    ) -> T:
+    ) -> T | BlueskyRemoteError :
```
callumforrester, line 107: Unsure why the error is returned rather than raised

src/blueapi/service/main.py
```diff
@@ -132,6 +133,9 @@ def get_device_by_name(name: str, handler: BlueskyHandler = Depends(get_handler)
     return handler.get_device(name)
 
 
+default_task = Task(name="count", params={"detectors": ["x"]})
```
callumforrester, line 136: Would prefer this be called `example_task`

callumforrester (changes requested):

.vscode/launch.json
stan-dot: ok

.vscode/settings.json
stan-dot: sure

src/blueapi/cli/rest.py
stan-dot: I need to look into how to handle that from the CLI.

but tbh Python does not have an indicator on the function that it can throw some error. 

another solution is to always return a tuple [value, error], where each could be null

src/blueapi/service/main.py
stan-dot: because it's passed as default? examples are in docs

src/blueapi/cli/rest.py
callumforrester: Why not...?
```python
try:
    client.foo()
except BlueskyRemoteError:
    handle_error()
```

stan-dot:
But the handle error must be in the cli. Otherwise you either quit the program or return something

src/blueapi/cli/rest.py
callumforrester: @stan-dot presume you meant to comment this here:

> But the handle error must be in the cli. Otherwise you either quit the program or return something

You can handle the error in the `except` block

stan-dot: ok done

.vscode/launch.json
stan-dot: done

callumforrester: There is still a diff, please ensure this file has no diff in the PR

.vscode/settings.json
callumforrester, line 1: Ditto https://github.com/DiamondLightSource/blueapi/pull/409/files#r1589220995

callumforrester (changes requested):
@stan-dot nearly there! Just a couple more things and a rebase

.vscode/settings.json
callumforrester, line 1: Still to be removed

callumforrester (changes requested):
Happy to approve once the vscode config diff goes away and a rebase is done

stan-dot:
rebase is complex

stan-dot:
will do hectic rebase tomorrow

callumforrester (approved):
LGTM, thanks!

stan-dot:
this if failing due to testing coverage. to mock the polling right, we'd need to mock the devices being restarted. the polling logic is simple and custom, not sure what should happen here

stan-dot:
the CLI demands the access to RabbitMQ. should we spin one up as part of a docker-compose config with devcontainers?

callumforrester:
@stan-dot it is already spun up, see how the stomp tests are configured

src/blueapi/cli/cli.py
```diff
@@ -105,7 +106,8 @@ def wrapper(*args, **kwargs):
         try:
             func(*args, **kwargs)
         except ConnectionError:
-            print("Failed to establish connection to FastAPI server.")
+            logging.error("Failed to establish connection to FastAPI server.")
```
callumforrester, line 109: https://github.com/DiamondLightSource/blueapi/pull/409/files#r1616783954

src/blueapi/cli/cli.py
```diff
@@ -116,7 +118,7 @@ def wrapper(*args, **kwargs):
 def get_plans(obj: dict) -> None:
     """Get a list of plans available for the worker to use"""
     client: BlueapiRestClient = obj["rest_client"]
-    pprint(client.get_plans().dict())
+    logging.info(client.get_plans().dict())
```
callumforrester, line 121: https://github.com/DiamondLightSource/blueapi/pull/409/files#r1616783954

src/blueapi/cli/cli.py
```diff
@@ -139,16 +141,17 @@ def listen_to_events(obj: dict) -> None:
...
+        logging.logger("Message bus needs to be configured")
+        return
 
     def on_event(
         context: MessageContext,
         event: WorkerEvent | ProgressEvent | DataEvent,
     ) -> None:
         converted = json.dumps(event.dict(), indent=2)
-        print(converted)
+        logging.logger(converted)
```
callumforrester, line 152: https://github.com/DiamondLightSource/blueapi/pull/409/files#r1616783954

src/blueapi/cli/cli.py
```diff
@@ -139,16 +141,17 @@ def listen_to_events(obj: dict) -> None:
             StompMessagingTemplate.autoconfigured(config.stomp)
         )
     else:
-        raise RuntimeError("Message bus needs to be configured")
+        logging.logger("Message bus needs to be configured")
```
callumforrester, line 144: https://github.com/DiamondLightSource/blueapi/pull/409/files#r1616783954

src/blueapi/cli/cli.py
```diff
@@ -139,16 +141,17 @@ def listen_to_events(obj: dict) -> None:
...
     def on_event(
         context: MessageContext,
         event: WorkerEvent | ProgressEvent | DataEvent,
     ) -> None:
         converted = json.dumps(event.dict(), indent=2)
-        print(converted)
+        logging.logger(converted)
 
-    print(
+    logging.logger(
```
callumforrester, line 154: https://github.com/DiamondLightSource/blueapi/pull/409/files#r1616783954

src/blueapi/cli/cli.py
```diff
@@ -180,9 +183,8 @@ def run_plan(
     if config.stomp is not None:
         _message_template = StompMessagingTemplate.autoconfigured(config.stomp)
     else:
-        raise RuntimeError(
-            "Cannot run plans without Stomp configuration to track progress"
-        )
+        logging.error("Cannot run plans without Stomp configuration to track progress")
```
callumforrester, line 186: https://github.com/DiamondLightSource/blueapi/pull/409/files#r1616783954

callumforrester (changes requested):
Mainly the prints. Otherwise happy!

@stan-dot if you want to make the case for replacing all the prints with logging calls I'm open to that discussion, however I think it should be a separate ticket and separate PR since it's already polluted this one quite a bit.

src/blueapi/cli/cli.py
```diff
@@ -116,7 +118,7 @@ def wrapper(*args, **kwargs):
 def get_plans(obj: dict) -> None:
     """Get a list of plans available for the worker to use"""
     client: BlueapiRestClient = obj["rest_client"]
-    pprint(client.get_plans().dict())
+    print(client.get_plans().dict())
```
callumforrester, line 121: Dictionaries should still be `pprint`s, ditto for the other pprints

callumforrester (changes requested):

src/blueapi/cli/cli.py
stan-dot: ok sure

stan-dot: changed all dict prints into pprints

src/blueapi/cli/cli.py
```diff
@@ -180,9 +185,8 @@ def run_plan(
     if config.stomp is not None:
         _message_template = StompMessagingTemplate.autoconfigured(config.stomp)
     else:
-        raise RuntimeError(
-            "Cannot run plans without Stomp configuration to track progress"
-        )
+        print("Cannot run plans without Stomp configuration to track progress")
```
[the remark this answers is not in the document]
stan-dot: ok sys.stderr.write(message)

callumforrester (changes requested):
TODO: Fix conflicts and raise exceptions rather than print error messages

stan-dot:
but then we get something like #206

callumforrester (approved):

DiamondJoseph (changes requested):

tests/test_cli.py
```diff
@@ -200,3 +197,212 @@ def test_valid_stomp_config_for_listener(runner: CliRunner):
...
+    assert mock_get_environment.call_count == 3
+
+    # Verify if sleep was called between polling iterations
+    assert mock_sleep.call_count == 2  # Since the last check doesn't require a sleep
+
+    # Check if the final environment status is printed correctly
+    # assert "Environment is initialized." in result.output
+    assert (
+        reload_result.output
+        == "Reloading the environment...\nEnvironment reload initiated.\nWaiting for environment to initialize...\nWaiting for environment to initialize...\nEnvironment is initialized.\ninitialized=True\n"  # noqa: E501
```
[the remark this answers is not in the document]
stan-dot: running into issues trying to fix this, I still get an error. keeping like it was as I see no instances of a similar thing happening elsewhere in the test part of the codebase

tests/test_cli.py
```diff
@@ -200,3 +197,212 @@ def test_valid_stomp_config_for_listener(runner: CliRunner):
...
+
+@pytest.mark.handler
+@responses.activate
+@patch("blueapi.service.handler.Handler")
+@patch("blueapi.cli.cli.sleep", return_value=None)
+def test_env_reload_server_side_error(
+    mock_sleep: MagicMock, mock_handler: Mock, handler: Handler, runner: CliRunner
+):
+    # Setup mocked error response from the server
+    responses.add(
```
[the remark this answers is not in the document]
stan-dot: done

DiamondJoseph (approved):

PR #414, opened 2024-04-09
Add logo to README and concept image

joeshannon:
This is following the approach adopted by other DLS copier based projects e.g. scanspec.


README.md
```diff
@@ -1,3 +1,6 @@
+<img src="https://raw.githubusercontent.com/DiamondLightSource/blueapi/main/docs/images/blueapi-logo.svg"
+     style="background: none" width="120px" height="120px" align="right">
+
```
callumforrester, line 3: @coretl Can't remember if you were in favour of this in the end. @joeshannon for now I would suggest just doing whatever the copier template is doing, but  happy to merge as-is

callumforrester (approved):

README.md
coretl: This is fine, depending on the project you may need to reduce to `100px` or `110px` to make it look right, but it seems to look ok here

Issue #426, opened 2024-04-19
Cannot run plans without a message bus (server)


callumforrester:
Blueapi can start up without connecting to a message bus but fails to run any plans because the message bus is currently interconnected with the worker. 

## To Reproduce

Terminal 1
```shell
# start blueapi server
blueapi serve
```

Terminal 2
```shell
# Simple plan
blueapi controller run sleep '{"time": 5}'

# More complex plan
blueapi controller run count '{"detectors": ["det"]}'
```

## Possible solution
Make the message template an optional part of the handler and make it `None` if connection fails. Perform None-checks on it wherever it is used. Hopefully if its type is declared as optional in all the right places mypy should point out all the usages where it is assumed to be present.
 

## Acceptance Criteria
- Can run the demo set of plans without connecting blueapi to a message bus
- Tests to prove that the message bus usages are guarded

ZohebShaikh:
In the Terminal 2 the example is given for the CLI which should raise exceptions when trying to run https://github.com/DiamondLightSource/blueapi/issues/427#issuecomment-2078755742
The blueapi server can run plans without a message bus and there are already tests to check it in the test_rest_api.py it uses a `handler = Handler(context=context, messaging_template=MagicMock())`
MagiMock which will run with a None as well

There are tests added to check if the message bus use is guarded

Issue #427, opened 2024-04-19
Cannot run plans without a message bus (client)


callumforrester:
Relates to #426, the blueapi CLI listens to updates from the server via the message bus so it throws an error if you try to use blueapi without a bus being present. 

## To Reproduce

Terminal 1
```shell
# start blueapi server
blueapi serve
```

Terminal 2
```shell
# Simple plan
blueapi controller run sleep '{"time": 5}'

# More complex plan
blueapi controller run count '{"detectors": ["det"]}'
```

## Possible solution
The CLI does have a message bus-free mode that can be enabled with a flag, we should also enable it if the connection fails and print a warning to the user. In this case the CLI does not block until the plan completes but expects the user to run further commands to poll the server. The warning message should make this clear. 

## Acceptance Criteria
- Can run plans via the CLI without a message bus without getting stomp errors
- Appropriate changes to CLI tests

callumforrester:
> The CLI does have a message bus-free mode that can be enabled with a flag

As it turns out I was misremembering. So will accept the CLI raising early if it can't connect to the bus

ZohebShaikh:
The use of message bus is guarded and there are tests added to check this for the CLI .
The changes mentioned are in this PR: https://github.com/DiamondLightSource/blueapi/pull/433

Issue #285, opened 2023-08-01
Enable running without a message bus


DiamondJoseph:

callumforrester:
Blueapi can now tolerate a message bus connection failure on startup, but is not very usable. I've opened two separate tickets to fix that: #426 and #427.

I do still think this issue is valid as we want blueapi to be able to start up this way by design rather than just when the message bus fails, will post details below.

callumforrester:
As a developer I want to configure the bluepai server and CLI to work without a message bus so that I can easily run it up without messing around with external dependencies. 

## Possible Solution
The way I suggest to do this is make `StompConfig` in `config.py` an optional part of `ApplicationConfig` and run mypy. It will highlight where the stomp config is being used on the assumption that it is definitely there and where None-checks are needed. Ultimately stomp can be made optional.

## Acceptance Criteria
- If a user does not supply a stomp config in the config file, blueapi server does not attempt to connect to a message bus on startup 
- If a user does not supply a stomp config in the config file, blueapi client does not attempt to connect to a message bus on running a plan
- There are tests that prove both paths

PR #433, opened 2024-04-24
Remove message bus dependency

ZohebShaikh:
closes https://github.com/DiamondLightSource/blueapi/issues/426 https://github.com/DiamondLightSource/blueapi/issues/427 https://github.com/DiamondLightSource/blueapi/issues/285


tests/test_cli.py
```diff
@@ -169,3 +169,22 @@ def test_config_passed_down_to_command_children(
             "params": {"time": 5},
         }
     }
+
+
+def test_invalid_stomp_config_for_listner(runner: CliRunner):
+    result = runner.invoke(main, ["controller", "listen"])
+    assert result.exit_code == 1
```
joeshannon, line 176: Instead of using the exit code, `result.exception` could be checked instead. Then could raise a custom exception and check that it is the expected type.

tests/test_cli.py
```diff
@@ -169,3 +169,22 @@ def test_config_passed_down_to_command_children(
...
+        result = runner.invoke(
+            main,
+            [
+                "-c",
+                "tests/example_yaml/valid_stomp_config.yaml",
+                "controller",
+                "listen",
+            ],
+        )
+    assert result.exit_code == 1
```
joeshannon, line 190: The `controller listen` command connects to a message bus and then prints messages to stdout. It uses the `input` builtin to wait for the user to press Enter to terminate the process

Therefore could add change the invoke to
```
        result = runner.invoke(
            main,
            [
                "-c",
                "tests/example_yaml/valid_stomp_config.yaml",
                "controller",
                "listen",
            ], input="\n"
        )
```
to tell the testing to immediately terminate the process.

This will then cause the `exit_code` to be 0 instead of 1 which makes sense here.

tests/test_cli.py
```diff
@@ -169,3 +169,22 @@ def test_config_passed_down_to_command_children(
...
     }
+
+
+def test_invalid_stomp_config_for_listner(runner: CliRunner):
+    result = runner.invoke(main, ["controller", "listen"])
+    assert result.exit_code == 1
+
+
+def test_valid_stomp_config_for_listner(runner: CliRunner):
+    with patch("uvicorn.run", side_effect=None):
```
joeshannon, line 180: This patch is not really required here as the listen command only requires a message bus and not the blueapi fastapi server.

joeshannon (changes requested):

tests/test_cli.py
```diff
@@ -169,3 +169,22 @@ def test_config_passed_down_to_command_children(
...
         }
     }
+
+
+def test_invalid_stomp_config_for_listner(runner: CliRunner):
+    result = runner.invoke(main, ["controller", "listen"])
+    assert result.exit_code == 1
+
+
+def test_valid_stomp_config_for_listner(runner: CliRunner):
```
DiamondJoseph, line 179: As this test requires that Stomp is available, it should be marked as such for Pytest
@pytest.mark.stomp

DiamondJoseph (changes requested):

src/blueapi/cli/cli.py
```diff
@@ -172,8 +173,13 @@ def run_plan(
...
     logger = logging.getLogger(__name__)
-
-    amq_client = AmqClient(StompMessagingTemplate.autoconfigured(config.stomp))
+    _message_template = StompMessagingTemplate.autoconfigured(config.stomp)
+    if _message_template is not None:
+        amq_client = AmqClient(_message_template)
+    else:
+        raise RuntimeError(
+            "Message bus needs to be configured to get done message after run"
+        )
```
DiamondJoseph, line 182:
```suggestion
        raise RuntimeError(
            "Cannot run plans without Stomp configuration to track progress"
        )
```

tests/example_yaml/rest_config.yaml
```diff
@@ -1,3 +1,9 @@
 api:
   host: a.fake.host
   port: 12345
+stomp:
+  auth:
+    username: admin
+    passcode: admin
+  host: localhost
+  port: 61613
```
DiamondJoseph, line 9: nit: restore blank final line

tests/example_yaml/valid_stomp_config.yaml
```diff
@@ -0,0 +1,6 @@
+stomp:
+  auth:
+    username: admin
+    passcode: admin
+  host: localhost
+  port: 61613
```
DiamondJoseph, line 6: nit: restore blank final line

tests/messaging/test_stomptemplate.py
```diff
@@ -218,3 +222,15 @@ def server(ctx: MessageContext, message: str) -> None:
         template.send(reply_queue, "ack", correlation_id=ctx.correlation_id)
 
     template.subscribe(destination, server)
+
+
+def test_autoconfigured_none_is_none():
```
DiamondJoseph, line 227:
```suggestion
def test_no_default_configuration():
```

src/blueapi/cli/cli.py
joeshannon: It still might be worth changing this to a custom exception to confirm that the tests are hitting this code, as a `RuntimeError` could be raised by something else.

ZohebShaikh: That is logical, I have added the test for it

src/blueapi/messaging/stomptemplate.py
```diff
@@ -101,7 +101,9 @@ def __init__(
         self._subscriptions = {}
 
     @classmethod
-    def autoconfigured(cls, config: StompConfig) -> MessagingTemplate:
+    def autoconfigured(cls, config: StompConfig | None) -> MessagingTemplate | None:
+        if config is None:
+            return None
```
callumforrester, line 106: Should: Similar to `AmqClient`, I think accepting `None` in the signature but immediately rejecting it in the body is confusing. I'd prefer that this only accept `config: StompConfig` and the code calling `autoconfigured` performs the `None`-check

src/blueapi/cli/cli.py
```diff
@@ -143,10 +143,11 @@ def on_event(
...
 
-    print(
-        "Subscribing to all bluesky events from "
-        f"{config.stomp.host}:{config.stomp.port}"
-    )
+    if config.stomp is not None:
+        print(
+            "Subscribing to all bluesky events from "
+            f"{config.stomp.host}:{config.stomp.port}"
+        )
```
callumforrester, line 150: Must: This command does nothing without a message bus so should probably just raise if there's no stomp config

DiamondJoseph (changes requested):

callumforrester (changes requested):
Thanks for this, just a few minor points

tests/test_cli.py
```diff
@@ -169,3 +169,44 @@ def test_config_passed_down_to_command_children(
...
+    assert (
+        isinstance(result.exception, RuntimeError)
+        and str(result.exception) == "Message bus needs to be configured"
+    )
+
+
+def test_cannot_run_plans_without_stomp_config(runner: CliRunner):
+    with patch("uvicorn.run", side_effect=None):
+        result = runner.invoke(main, ["serve"])
+    assert result.exit_code == 0
```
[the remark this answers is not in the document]
DiamondJoseph: As discussed, we want a server that is configured with the same communication methods as the client, so that behaviour when other communication methods are added is clear without editing the tests.

DiamondJoseph (approved):
Good to go once Callum's changes have been made

callumforrester:
@ZohebShaikh please could you also reference the issues this fixes in the PR description?

DiamondJoseph (approved):

callumforrester (approved):

joeshannon (approved):

PR #441, opened 2024-04-29
testing client gen

stan-dot:


stan-dot:
need to get mypy and ruff to ignore the auto-gen folder

stan-dot:
> rc/blueapi/cli/cli_new.py:195: error: Argument 1 to "submit_task_tasks_post" of "DefaultApi" has incompatible type "blueapi.worker.task.Task"; expected "blueapi.openapi_client.models.task.Task"  [arg-type]
src/blueapi/cli/cli_new.py:200: error: Argument 1 to "update_task_worker_task_put" of "DefaultApi" has incompatible type "blueapi.service.model.WorkerTask"; expected "blueapi.openapi_client.models.worker_task.WorkerTask"  [arg-type]
src/blueapi/cli/cli_new.py:230: error: Argument 1 to "set_state_worker_state_put" of "DefaultApi" has incompatible type "WorkerState"; expected "StateChangeRequest"  [arg-type]
src/blueapi/cli/cli_new.py:240: error: Argument 1 to "set_state_worker_state_put" of "DefaultApi" has incompatible type "WorkerState"; expected "StateChangeRequest"  [arg-type]
src/blueapi/cli/cli_new.py:257: error: Argument "state_change_request" to "set_state_worker_state_put" of "DefaultApi" has incompatible type "blueapi.service.model.StateChangeRequest"; expected "blueapi.openapi_client.models.state_change_request.StateChangeRequest"  [arg-type]
src/blueapi/cli/cli_new.py:270: error: Argument "state_change_request" to "set_state_worker_state_put" of "DefaultApi" has incompatible type "blueapi.service.model.StateChangeRequest"; expected "blueapi.openapi_client.models.state_change_request.StateChangeRequest"  [arg-type]
Found 6 errors in 1 file (checked 73 source files)
ERROR: InvocationError for command /opt/hostedtoolcache/Python/3.11.9/x64/bin/mypy src tests (exited with code 1)

one possible issue is the types overlap. possibly it's not duck-typed

stan-dot:
potentially might need lines to delete the duplicate models and import from the original ones. but that would break the isolation

stan-dot:
ok now [the only errors](https://github.com/DiamondLightSource/blueapi/actions/runs/8907617094/job/24461805752?pr=441) are that the library needs to be mocked. this needs consultation wrt out testing mocking strategy. we likely don't want to have a running server to unit test cli

stan-dot:
I could mock some method from inside the testing library, not sure which one atm. possibly the more 'raw' one the better (so requests(?))

stan-dot:
it's a maze though

stan-dot:
there are various possible proxy urls

stan-dot:
possible solution - to mock the specific method on urllib3 pool manager
`@patch('urllib3.PoolManager.request')
def test_my_client_method(self, mock_request):
    # Setup a mock response
    mock_response = MagicMock()
    mock_response.status = 200  # Set the HTTP status code
    mock_response.data = b'{"result": "success"}'  # Mock a JSON response

    # Configure the mock to return the response
    mock_request.return_value = mock_response

    # Call the method under test
    response = self.client.my_method()  # Replace 'my_method' with your actual method

    # Assert the expected outcome
    self.assertEqual(response['result'], 'success')
    # You can also check if the request was made as expected
    mock_request.assert_called_once_with(
        'GET',  # or 'POST', etc.
        'http://example.com/api',  # Expected URL
        body=None,  # Expected body
        preload_content=True,  # Other parameters as expected
        timeout=None,
        headers={}
    )
`

stan-dot:
found out the error:
`
@main.group()
@click.pass_context
def controller(ctx: click.Context) -> None:
    """Client utility for controlling and introspecting the worker"""

        return

    ctx.ensure_object(dict)
    config: ApplicationConfig = ctx.obj["config"]
    # ctx.obj["rest_client"] = BlueapiRestClient(config.api)
    # todo consider the API being not that useful
    ctx.obj["rest_client"] = DefaultApi(config.api)`
    
the context expects a specific format and the context does mot match

stan-dot:
major progress, need some newline errors

stan-dot:
update: need to add various RESTResponseObject fields to the magic mock response - such as reason or JSONCONTENt type

joeshannon:
Have we investigated if the code can be generated at "build" time with python? This is the approach we have used for autogenerated Java clients and has the advantage that the autogenerated code doesn't have to be committed as if they were source files.

stan-dot:
realization - need to nicely handle the stomp error in the CLI first (even withoutthe test)

stan-dot:
what is the expected behavior there? 
image


should we be able to run the cli plans without the bus running?

callumforrester:
@stan-dot see this  https://github.com/DiamondLightSource/blueapi/issues/459 and all the issues linked from it

callumforrester:
> Have we investigated if the code can be generated at "build" time with python? This is the approach we have used for autogenerated Java clients and has the advantage that the autogenerated code doesn't have to be committed as if they were source files.

@joeshannon yes, I agree, I wonder if there is a way to generate it when we run `pip install` for dev and build it into the wheel for prod...

stan-dot:
> This is the approach we have used for autogenerated Java clients and has the advantage that the autogenerated code doesn't have to be committed as if they were source files
why is this an advantage?

stan-dot:
> In dev if you don't need a message bus simply don't define one.

but the CLI expects it @DiamondJoseph 
https://github.com/DiamondLightSource/blueapi/issues/459#issuecomment-2100008467

stan-dot:
`def run_plan(
    obj: dict, name: str, parameters: str | None, timeout: float | None
) -> None:
    """Run a plan with parameters"""
    client: BlueapiRestClient = obj["rest_client"]

    if config.stomp is not None:
        _message_template = StompMessagingTemplate.autoconfigured(config.stomp)
    else:
        raise RuntimeError(
            "Cannot run plans without Stomp configuration to track progress"
        )
    amq_client = AmqClient(_message_template)
    
    `
    
    @callumforrester what do you think?

stan-dot:
trying to run without the stomp listener

stan-dot:
actually mocking the stomp message bus might be one way to do this

stan-dot:
there is a mock for amqpt

or
https://microcks.io/
or
https://pypi.org/project/amqp-mock/

stan-dot:
proceeding with testing the microcks as it has CNCF sandbox status https://www.cncf.io/sandbox-projects/

stan-dot:
mocking with this: https://microcks.io/documentation/guides/rabbitmq-support/

stan-dot:
stomp config conundrum

stan-dot:
no idea why did this close

stan-dot:
[pydantic v1 generator](https://openapi-generator.tech/docs/generators/python-pydantic-v1/) wasn't all that great to work it and only now the assumption 'not to put generated code into git' has surfaced. 

idea - to work on this with the generic, hopefully better supported [python generator](https://openapi-generator.tech/docs/generators/python/) , once the #449 is done

stan-dot:
as mentioned in the issue, this has been a learning experience, though it will be better to do this from scratch with python generator (not pydantic-v1) and from the start as without git ignored

PR #446, opened 2024-04-30
Install git in runtime container

### LINKED ISSUE #445: Git required in container
As part of the migration to python-copier-template the runtime container base was switched from `python:3.11` to `python:3.11-slim`.
The slim container doesn't contain git which is currently required to install additional modules at runtine (this is to support the scratch directory).

joeshannon:
This is required for installing python packages at runtime.

It was previously present in the non-slim python container used before but this was accidentally removed with the switch to the copier template.

Fixes #445.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,6 +21,10 @@ RUN pip install .
 # The runtime stage copies the built venv into a slim runtime container
 FROM python:${PYTHON_VERSION}-slim as runtime
 # Add apt-get system dependecies for runtime here if needed
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    # Git required for installing packages at runtime
+    git \
+    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


joeshannon (commented):
I think this should work, I did look in the action logs though and it wasn't clear where git was being installed.

Also the `apt` in the container above has `&& rm -rf /var/lib/apt/lists/*` at the end of it - do we really need this here too?

joeshannon:
> I think this should work, I did look in the action logs though and it wasn't clear where git was being installed.

It was being installed but I missed it before as it seemed to be building the runtime container in parallel with the developer container.

coretl:
Yes, reading the docker best practices you get examples that look like that:
https://docs.docker.com/develop/develop-images/guidelines/#sort-multi-line-arguments

joeshannon:
Thanks!

PR #449, opened 2024-05-01
remove pydantic pinning

stan-dot:


DiamondJoseph:
I'm surprised it was as simple as just removing the pinning- the service still starts with our old-style BaseModels in Pydantic v2? Or are we picking up a Pydantic v1 pin from some other (dodal, ophyd-async?) dependency?

callumforrester:
Yeah, unsure if this is happy because it just works or lack of tests...

stan-dot:
should a test 'staging' deployment be used for this?

callumforrester:
@stan-dot I think it's fine if someone has the time to play around with a local dev version of the server.

stan-dot:
@callumforrester I can try this on a branch with the ViSR project is that ok?

callumforrester:
@stan-dot I think this configuration will still install pydantic v1 because the pinned fastapi version depends on it, see #454

stan-dot:
so it's dependent on GDA. rip
https://github.com/DiamondLightSource/blueapi/issues/454#issuecomment-2098016338

I'm unassigning myself from this issue as it does not seem resolvable in the nearest future

DiamondJoseph:
It's also pinned in ScanSpec, dodal and other places. We need to have a mass move over

stan-dot:
superceded by #564

PR #450, opened 2024-05-02
Auto restart deployment on config change

joeshannon:
Add new restartOnConfigChange property.

This will cause the deployment to be restarted if the config is changed,
e.g. the set of plans or device modules are updated, when running
helm upgrade.

Use a variation of approach at:
https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

Fixes 451.

### DIFF
--- a/helm/blueapi/templates/deployment.yaml
+++ b/helm/blueapi/templates/deployment.yaml
@@ -11,8 +11,11 @@ spec:
       {{- include "blueapi.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-      {{- with .Values.podAnnotations }}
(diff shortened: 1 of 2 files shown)


DiamondJoseph (approved):

DiamondJoseph (approved):
Performs how I'd expect it to (`helm upgrade` with changes to the configmap.yaml deploys the updated ConfigMap, causes the Deployment to become outdated and be re-deployed). Does require that changes to the ConfigMap are done through Helm upgrade, but that's something we should enforce anyway.

We don't seem to have a github issue for this

PR #452, opened 2024-05-02
tesitng if this quick schema checking is ok

stan-dot:
also doing #235 a bit fwiw?

### DIFF
--- a/.github/workflows/generate_and_verify_schema.yml
+++ b/.github/workflows/generate_and_verify_schema.yml
@@ -1,0 +1,36 @@
+name: Validate OpenAPI Schema
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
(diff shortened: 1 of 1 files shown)


callumforrester:
What does this do that https://github.com/DiamondLightSource/blueapi/blob/main/tests/service/test_openapi.py doesn't?

stan-dot:
the difference is this verifies that it's preserving it across time.

that python file merely checks if the schema conforms to the schema in the general sense

joeshannon:
> the difference is this verifies that it's preserving it across time.
> 
> that python file merely checks if the schema conforms to the schema in the general sense

Could you elaborate on the "across time" part? I don't think I understand. The existing python test is also generating the schema and comparing it to the file in `docs/user/reference/openapi.yaml`.

stan-dot:
ah sorry I was imprecise about the 'across time'. I didn't read the second test in the file closely.

stan-dot:
I'd still make the case that the job of generating schema is different from the test.

We could keep the test file but keep it in a different job, different kind of test.

callumforrester:
So the purpose of the test is explicitly _not_ to autogenerate a schema. If you make a change to the code that changes the schema then you, the human, should be aware of it. The change may be breaking so the most important thing is that the schema is only changed deliberately.

stan-dot:
I fulyl agree that the human must be in the loop here

callumforrester:
Fixed by #238 

For client generration, we would ideally like it to be in the build system rather than Github CI

PR #462, opened 2024-05-10
Remove ophyd_async_connect

### LINKED ISSUE #461: Remove code to wait for device connection
The device connection is now handled by `device_instantiation` in dodal. This function also provides the option on whether to wait for connection, so it is not needed here too.

Additionally it can lead to undefined (currently) behaviour if the device is initially created with `fake_with_ophyd_sim = True` but then later connected again by blueapi with `fake_with_ophyd_sim = False`.

For full customisation and flexibility of lazy connect we need #440.

joeshannon:
This also leaves sim on BlueskyContext redundant so that is removed too.

Fixes #461.


callumforrester (approved):
LGTM, thanks!

PR #464, opened 2024-05-13
querying for unstarted tasks

stan-dot:


stan-dot:
forgot to update the Worker interface

stan-dot:
debugging the tests for the new feature now

stan-dot:
image


@keithralphs what is the desired business logic for this kind of test?

callumforrester (changes requested):
I would suggest an alternative implementation:

Replace `is_pending` and is `is_complete` on `Task` with a status enum and propagate that through the rest of the code. It simplifies your if check in the worker and also means you don't have to pass an ambiguous status string around.

stan-dot:
ok this will take a while and a fair bit of logic too

stan-dot:
there are two different models - classs TaskStatus in a worker but also TrackableTask.

this is a mess and too many classes.

stan-dot:
workerEvent should just include a reference to that

stan-dot:
you need to tell me the finite state machine logic of complete failed and error states for this @callumforrester .

callumforrester:
@stan-dot the reason `WorkerEvent` doesn't reference the task directly is to handle the null case. Where the worker is not currently doing a task.

callumforrester:
Another source of complexity is that the status of the worker and the status of the task the worker is currently doing are separate things and separately queryable. I would like it if it were simpler, I'm open to suggestions

src/blueapi/worker/reworker.py
```diff
@@ -136,6 +136,21 @@ def submit_task(self, task: Task) -> str:
         self._tasks[task_id] = trackable_task
         return task_id
 
+    def get_tasks_by_status(self, status: str) -> list[TrackableTask[Task]]:
```
callumforrester, line 139: Status should be an enum

stan-dot: will do hectic refactoring tomorrow

stan-dot: update: changing the argus to be a required instance of the enum , without the big refactor

stan-dot:
moved to #470 f

PR #467, opened 2024-05-13
added none if device not found

### LINKED ISSUE #444: Don't Fall Over if a Device Does not Connect
As a user I would still like to use blueapi if a particular device fails to connect, assuming there are still plans available that can be used without it, so I can continue operations even if a PV is wrong or a device is unavailable.

This is relevant to, but distinct from #440. If this issue is implemented we expect blueapi to start even if a device does not connect but fall over if we try to use it. Implementing #440 would mean it gives a sensible error if we try to use it.

Implementing this behaviour would also be inkeeping with #285. Blueapi's design philosophy seems to be evolving towards: "start even if there's nothing you can use, and report it".

## Possible Solution

Currently unsure if the code change required is in blueapi or dodal

## Acceptance Criteria
- If an error occurs instantiating or connecting to a device, blueapi still starts up
- In this circumstance, the devices that did successfully instantiate or connect are still available in blueapi
- There are tests that prove this

ZohebShaikh:
closes #444

### DIFF
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -9,7 +9,7 @@ import click
 from requests.exceptions import ConnectionError
 
 from blueapi import __version__
-from blueapi.cli.amq import AmqClient
+from blueapi.cli.event_bus_client import EventBusClient
 from blueapi.config import ApplicationConfig, ConfigLoader
 from blueapi.core import DataEvent
 from blueapi.messaging import MessageContext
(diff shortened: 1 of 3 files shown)


callumforrester (changes requested):
I would prefer we keep the original behaviour here, the problem in #444 is not about retrieving devices from the device store. It is about what ends up in the device store in the first place. If the user asks for a device that doesn't exist, they _should_ get an error, not a `None`. It's when `dodal` makes the devices in the first place that has to change. If dodal fails to make some of the devices, it should still return the ones it _does_ manage to make.

PR #468, opened 2024-05-17
added indexing to handle tuple

ZohebShaikh:
closes https://github.com/DiamondLightSource/blueapi/issues/444

With PR for dodal at https://github.com/DiamondLightSource/dodal/pull/550

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -106,8 +106,14 @@ class BlueskyContext:
     def with_dodal_module(self, module: ModuleType, **kwargs) -> None:
         from dodal.utils import make_all_devices
 
-        for device in make_all_devices(module, **kwargs)[0].values():
+        devices, exceptions = make_all_devices(module, **kwargs)
+
+        for device in devices.values():
             self.device(device)
+
+        num_failed = len(exceptions)
+
+        LOGGER.warning(f"{num_failed} device(s) failed to connect: {exceptions}")
 
     def plan(self, plan: PlanGenerator) -> PlanGenerator:
         """
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -7,14 +7,12 @@ from types import ModuleType, UnionType
 from typing import Any, Generic, TypeVar, Union, get_args, get_origin, get_type_hints
 
 from bluesky.run_engine import RunEngine
+from ophyd_async.core import NotConnected
 from pydantic import create_model
 from pydantic.fields import FieldInfo, ModelField
 
 from blueapi.config import EnvironmentConfig, SourceKind
-from blueapi.utils import (
-    BlueapiPlanModelConfig,
-    load_module_all,
-)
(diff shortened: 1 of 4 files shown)


ZohebShaikh:
The tests are failing now because it is testing it with the old version of dodal

src/blueapi/core/context.py
```diff
@@ -106,7 +106,7 @@ def with_device_module(self, module: ModuleType) -> None:
     def with_dodal_module(self, module: ModuleType, **kwargs) -> None:
         from dodal.utils import make_all_devices
 
-        for device in make_all_devices(module, **kwargs).values():
+        for device in make_all_devices(module, **kwargs)[0].values():
             self.device(device)
```
callumforrester, line 110: I think we want the issues logged

```suggestion
        devices, exceptions = make_all_devices(module, **kwargs)
        for device in devices.values():
            self.device(device)
        
        num_failed = len(exceptions)
        LOGGER.warning(f"{num_failed} devices failed to connect: {exceptions}")
```

callumforrester (changes requested):
Tests would be useful, even they are just copy/pastes of the dodal tests for now, we can work on consolidation later.

src/blueapi/core/context.py
```diff
@@ -106,9 +106,15 @@ def with_device_module(self, module: ModuleType) -> None:
     def with_dodal_module(self, module: ModuleType, **kwargs) -> None:
         from dodal.utils import make_all_devices
 
-        for device in make_all_devices(module, **kwargs).values():
+        devices, exceptions = make_all_devices(module, **kwargs)
+
+        for device in devices.values():
             self.device(device)
 
+        num_failed = len(exceptions)
```
DiamondJoseph, line 114: Should we register devices that had exceptions?

DiamondJoseph: Has the signature of make_all_devices changed? Seems it used to return `dict[str, Device]` but now returns `tuple[dict[str, Device], dict[str, Exception]]`?

callumforrester: It changes in the corresponding dodal PR, see description: https://github.com/DiamondLightSource/dodal/pull/550

callumforrester:
Tests will pass following a dodal release

src/blueapi/core/context.py
```diff
@@ -104,11 +103,17 @@ def with_device_module(self, module: ModuleType) -> None:
...
+        devices, exceptions = make_all_devices(module, **kwargs)
 
-        for device in make_all_devices(module, **kwargs).values():
+        for device in devices.values():
             self.device(device)
 
+        # If exceptions have occurred, we log them but we do not make blueapi
+        # fall over
+        if len(exceptions) > 0:
+            LOGGER.warning(f"{len(exceptions)} device(s) failed to connect")
```
DiamondJoseph, line 114:
```suggestion
            LOGGER.warning(f"{len(exceptions)} exceptions while attempting to connect ")
```

DiamondJoseph (approved):

PR #473, opened 2024-05-20
apply changes again clean

stan-dot:


callumforrester:
@stan-dot the PR title/description will become a commit message, please can you make them more informative?

stan-dot:
I mean I put a commit name and only then it became a PR name. I thought that the commit names get squashed anyway

callumforrester (approved):
LGTM! Thanks

stan-dot:
ok it does not make sense to validate that 'the wrong plan gets rejected', as that is not a responsibility of CLI

stan-dot:
@callumforrester so locally I can start `blueapi serve` without example config. but then it is necessary for running the controller. 

but then the values are discordant.

when I try to run blueapi serve with the file, I get:
image

src/blueapi/cli/cli.py
```diff
@@ -191,10 +191,21 @@ def store_finished_event(event: WorkerEvent) -> None:
...
-    resp = client.create_task(task)
-    task_id = resp.task_id
+    task_id = ""
+    parsed_params = json.loads(parameters) if isinstance(parameters, str) else {}
+    try:
+        task = Task(name=name, params=parsed_params)
+        resp = client.create_task(task)
+        task_id = resp.task_id
+    except ValidationError:
+        pprint(f"failed to validate the task parameters, {task_id}")
```
callumforrester, line 201: Can you also print/raise the validation error so the user gets details of what parameters they got wrong?

stan-dot: yeah, sure

PR #475, opened 2024-05-21
Rename update_task and add some docstrings to REST endpoint methods

joeshannon:
The method names and docstrings are used in the OppenAPI schema (and therefore Swagger UI).

"update_task" was a bit vague as calling this actually starts the specified task.

### DIFF
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -88,6 +88,7 @@ async def on_key_error_404(_: Request, __: KeyError):
 def get_environment(
     handler: BlueskyHandler = Depends(get_handler),
 ) -> EnvironmentResponse:
+    """Get the current state of the environment, i.e. initialization state."""
     return handler.state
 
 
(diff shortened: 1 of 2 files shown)


src/blueapi/service/main.py
```diff
@@ -167,10 +170,11 @@ def delete_submitted_task(
     response_model=WorkerTask,
     responses={status.HTTP_409_CONFLICT: {"worker": "already active"}},
 )
-def update_task(
+def set_active_task(
     task: WorkerTask,
     handler: BlueskyHandler = Depends(get_handler),
 ) -> WorkerTask:
+    """Begin the specified task."""
```
joeshannon, line 177: Maybe this isn't quite optimal wording...

callumforrester: How about?

```suggestion
    """Promote a task to active status, the worker should begin it as soon as possible"""
```

callumforrester (approved):
LGTM

src/blueapi/service/main.py
DiamondJoseph: Promote implies (to me) that it can displace other currently running tasks.
"Attempt to set the task to be run on the worker"?

joeshannon:
Finally revisited this PR. Have used a combination of both wording suggestions.

PR #476, opened 2024-05-21
Hook directory provider into run engine

callumforrester:


stan-dot:
ophyd-async does not use directory provider anymore, but pathprovider instead - at least the name of this PR needs to change

callumforrester:
We need to wait for https://github.com/bluesky/ophyd-async/issues/314 and ACQP-150 (jira) before we can really start this anyway now. I'll close it.

PR #477, opened 2024-05-21
fix one test to see if it's acceptable

stan-dot:


stan-dot:
from discussion: this is not about changing tests, but the implementation

stan-dot:
it's not finished yet! need to change the implementation. not sure where though, as the inheritance tree is deep.

stan-dot:
mixed this up with some other branch

PR #480, opened 2024-05-21
Set umask in container

### LINKED ISSUE #478: Mitigate permission issues in scratch area
Often we are stuck with files with inconsistent permissions in the scratch area when on the shared filesystem.

A small improvement would be to set the umask to match the DLS standard (0002).

joeshannon:
Performing an editable install on a shared filesystem scratch area causes cache/version files to be written in the source repository.

With the container's default umask of 0022 the files are not group writable which can lead to frustration when attempting to clean up the scratch area.

Use the DLS "standard" of 0002.

Additionally modify the looping to only pickup directories within the scratch area.

Fixes #478.

### DIFF
--- a/container-startup.sh
+++ b/container-startup.sh
@@ -1,18 +1,19 @@
 #!/usr/bin/env bash
 
+# Set umask to DLS standard
+umask 0002
+
 if [[ -z "${SCRATCH_AREA}" ]]; then
   SCRATCH_AREA="/blueapi-plugins/scratch"
 fi
 
-mkdir -p ${SCRATCH_AREA}
-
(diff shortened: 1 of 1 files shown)


callumforrester (approved):

joeshannon:
Would it help to additionally add a `pip freeze`here at the end to be clear about what we have installed?

callumforrester:
Unsure, should be a separate PR if so

DiamondJoseph (approved):
Should this be an option in the python-copier-template?

callumforrester:
That's a question for @coretl and @GDYendell

DiamondJoseph:
I'll make an issue and discussion can take place there

PR #482, opened 2024-05-21
deleted

stan-dot:


stan-dot:
Tests are failing, might just need to move those to tests as mocks

stan-dot:
or change the name, 'startup' is misleading

stan-dot:
alternatively we could just use all-dodal stuff
image

callumforrester:
These are used, they produce the default, simulated environment when you run `blueapi serve` without config. I'll accept issues/PRs to modernise them/replace them with `ophyd-async`, especially now we have sim signals.

stan-dot:
closed in favour of #484

PR #488, opened 2024-05-30
 Update stomptemplate.py to check connection existing before disconnecting

### LINKED ISSUE #401: Stomp Tests Timeout Forever
If I run the stomp tests without a valid message bus running, they should fail within the 10 second timeout (they usually fail earlier when disconnection occurs). They currently hang forever, this behaviour seems to have been present since #345 so probably has something to do with the fact that we are now logging exceptions rather than raising them. This would be unfortunate as it means our solution that made the CLI easier to use makes the tests harder to use.

This ticket is therefore to come up with a solution to #220 that also works for this.

stan-dot:
fixes #401

### DIFF
--- a/tests/worker/test_reworker.py
+++ b/tests/worker/test_reworker.py
@@ -380,7 +380,7 @@ E = TypeVar("E")
 def take_n_events(
     stream: EventStream[E, Any],
     num: int,
-) -> "Future[list[E]]":
+) -> Future[list[E]]:
     count = itertools.count()
     return take_events(stream, lambda _: next(count) >= num)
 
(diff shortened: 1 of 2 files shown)


tests/worker/test_reworker.py
```diff
@@ -380,15 +380,15 @@ def assert_running_count_plan_produces_ordered_worker_and_data_events(
 def take_n_events(
     stream: EventStream[E, Any],
     num: int,
-) -> "Future[list[E]]":
+) -> Future[list[E]]:
```
callumforrester, line 383: @DiamondJoseph was this there for a reason? I forget

callumforrester (changes requested):
Please give this PR a better title

tests/worker/test_reworker.py
DiamondJoseph: Them being quoted? None that I can recall. If it's working without then great, but I worry given the forward ref issues we had in blueapi recently (but no forward refs here, causality is maintained).

stan-dot: I had to change that otherwise the pre-commit wouldn't pass

PR #491, opened 2024-05-31
Update stomptemplate.py to fix timeout bug

stan-dot:


callumforrester:
Happy with this if tests are fixed

PR #492, opened 2024-05-31
Configurable global plan wrapper

callumforrester:
Restore the global plan wrapper setting and make it configurable. Eventually we want to replace it with a combination of:
- https://github.com/DiamondLightSource/dodal/issues/576
- https://github.com/bluesky/ophyd-async/issues/314
But we need a working solution in the interim

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -73,6 +73,7 @@ class EnvironmentConfig(BlueapiBaseModel):
         Source(kind=SourceKind.PLAN_FUNCTIONS, module="dls_bluesky_core.stubs"),
     ]
     events: WorkerEventConfig = Field(default_factory=WorkerEventConfig)
+    global_plan_wrapper: str | None = None
 
 
 class LoggingConfig(BlueapiBaseModel):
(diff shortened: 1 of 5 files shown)


callumforrester:
@DiamondJoseph has convinced me not to do this for now, instead we are going to manually add `@attach_data_session_metadata_decorator` to all plans we want to use this for. This gives us our working interim solution.

The long-term solution is to sync the directory provider with the `RunEngine` and make detectors delay opening an HDF5 file until after a run starts, which is documented in these two issues:
- https://github.com/DiamondLightSource/dodal/issues/576
- https://github.com/bluesky/ophyd-async/issues/314

callumforrester:
@DiamondJoseph and @keithralphs have all the plans we want wrapped been wrapped yet?

DiamondJoseph:
IIRC Keith has a copy of `count` that has type annotations, decorators on a branch for.... i22?

keithralphs:
Yes we made the edits for these last Tuesday in readiness for when <a thing that Joseph told me but I've since forgotten> was completed. Hopefully, these changes were made on my laptop, or they may nave since been destroyed

keithralphs:
I can confirm that I do have the code on my laptop and it was waiting for https://github.com/DiamondLightSource/dodal/pull/590 so once a release of dodal including this is made, I can push the change to i22_bluesky that depends on the release.

PR #508, opened 2024-06-18
Add the unstarted tasks endpoint

stan-dot:


src/blueapi/service/main.py
```diff
@@ -173,6 +175,34 @@ def delete_submitted_task(
...
 
 
+def validate_task_status(v: str) -> TaskStatusEnum:
+    v_upper = v.upper()
+    if v_upper not in TaskStatusEnum.__members__:
+        raise ValueError("Invalid status query parameter")
+    return TaskStatusEnum(v_upper)
+
+
+@app.get("/tasks/", response_model=TasksListResponse, status_code=status.HTTP_200_OK)
```
callumforrester, line 185: Should: I believe this will actually require a trailing slash on `curl` commands etc. 

```suggestion
@app.get("/tasks", response_model=TasksListResponse, status_code=status.HTTP_200_OK)
```

src/blueapi/service/main.py
```diff
@@ -173,6 +175,34 @@ def delete_submitted_task(
...
+def get_tasks(
+    task_status: str,
```
callumforrester, line 187: Should: I'm surprised we didn't already have a `/tasks` endpoint, shall we make the query parameter optional so it provides all tasks if no desired state is supplied?

callumforrester (changes requested):
Getting there, thanks!

src/blueapi/service/main.py
callumforrester: (you could also just make an issue for this if preferred, so I'll retroactively change it to a "could")

src/blueapi/service/main.py
stan-dot: all the other get routes have no trailing slash at the end

src/blueapi/service/main.py
stan-dot: added this endpoint

src/blueapi/service/main.py
```diff
@@ -173,6 +175,38 @@ def delete_submitted_task(
...
+@app.get("/tasks", response_model=TasksListResponse, status_code=status.HTTP_200_OK)
+def get_tasks(
+    task_status: str | None = None,
+    handler: BlueskyHandler = Depends(get_handler),
+) -> TasksListResponse:
+    """
+    Retrieve tasks based on their status.
+    The status of a newly created task is 'unstarted'.
+    """
+    tasks = []
```
callumforrester, line 194: Nit: This line is redundant and can be removed.

callumforrester (approved):
Thanks for this

PR #514, opened 2024-06-19
Fix reloading the environment if there is an error

### LINKED ISSUE #512: Failure to reload the subprocess gets blueapi into an unrecoverable state
To reproduce:
1. Run blueapi against a scratch area
2. Edit the code to introduce a fatal bug (e.g. a syntax error)
3. Reload
4. Fix the bug
5. Try to reload again

joeshannon:
This adds a new error field to the environment. This is now checked against when attempting to delete the environment as if there is an error the handler is never set to initialized.

Note that the property has only been added to SubprocessHandler as the BlueskyHandler ABC is likely to be removed.

Fixes #512.


joeshannon (commented):
I have only added the error property to the subprocess handler (and not the ABC). This has required changes to some of the typing in `main.py`.

I'm a bit unsure but I thought it might save some work and if the handler is to be removed soon anyway then it could make sense.

What do you think @callumforrester @DiamondJoseph .

Note for me: the schema needs to be regenerated too.

joeshannon:
> I have only added the error property to the subprocess handler (and not the ABC).

Looking at the test failures now it seems the handler implementation is variable, perhaps I should just add to the ABC.

src/blueapi/service/subprocess_handler.py
```diff
@@ -113,6 +119,11 @@ def get_task_by_id(self, task_id: str) -> TrackableTask | None:
     def initialized(self) -> bool:
         return self._initialized
 
+    @property
+    # This method is not currently on ABC
+    def error(self) -> bool:
```
callumforrester, line 124: Should we just have the fact that there is/isn't an error or actually report an error message via the API?

joeshannon: Possibly, although I don't know how straightforward it is to get a meaningful message. The exception message might not give enough context, and the whole exception stack trace is a bit unwieldy.

callumforrester: Currently, when a plan fails, we pass just the exception message and expect the user to look at the logs if they require more info. Could do the same here?

joeshannon: Ok, yeah will look at doing that.

src/blueapi/service/main.py
```diff
@@ -100,7 +100,7 @@ def restart_handler(handler: BlueskyHandler):
         handler.stop()
         handler.start()
 
-    if handler.initialized:
+    if handler.state.initialized or handler.state.error_message:
```
callumforrester, line 103: Something can be false for reasons other than being `None`

```suggestion
    if handler.state.initialized or handler.state.error_message is not None:
```

callumforrester (changes requested):
Only a small nitpick, looks good, thanks!

src/blueapi/service/main.py
joeshannon: Thanks, yes the falsy empty string is still catching me out!

callumforrester (approved):

Issue #504, opened 2024-06-13
Consider moving the worker around (run engine) into a separate microservice 


stan-dot:
![image](https://github.com/DiamondLightSource/blueapi/assets/56644812/8a5d96bb-6b55-4e13-8275-5bfb0801c4eb)

We should strive to avoid bloat like with GDA and possibly those microservices could be deployed separately. 

```mermaid
graph TD
    CLI[CLI Tool] -->|Submits jobs| A[FastAPI Frontend]
    CLI -->|Previews IoT devices| IoT[IoT Device Service]
    CLI -->|Listens to| D[Message Bus]
    A -->|Reads jobs| B[Job Queue]
    B --> C[Worker Service]
    C -->|Emits results| D
    D --> E[Result Processing Service]
    C -->|Protobuf| F[Other Microservices]
    style CLI fill:#fff,stroke:#000,stroke-width:2px
    style A fill:#fff,stroke:#000,stroke-width:2px
    style B fill:#fff,stroke:#000,stroke-width:2px
    style C fill:#fff,stroke:#000,stroke-width:2px
    style D fill:#fff,stroke:#000,stroke-width:2px
    style E fill:#fff,stroke:#000,stroke-width:2px
    style F fill:#fff,stroke:#000,stroke-width:2px
    style IoT fill:#fff,stroke:#000,stroke-width:2px

```

callumforrester:
The subprocess already does some of this. We could separate the management and worker processes out completely, I'm not opposed to that. See also #371 

What is the IoT device service for?

stan-dot:
Yeah that inaccurate but basically ophyd async

callumforrester:
Are the ophyd-async devices instantiated twice? Once in the worker service and once in the IoT service? There's no arrow between them.

stan-dot:
I was not precise, now I changed it from 'service' to 'representation' to be more abstract. in essence this would entail splitting of the context:
```

@dataclass
class BlueskyContext:
    """
    Context for building a Bluesky application.

    The context holds the RunEngine and any plans/devices that you may want to use.
    """

    run_engine: RunEngine = field(
        default_factory=lambda: RunEngine(context_managers=[])
    )
    plans: dict[str, Plan] = field(default_factory=dict)
    devices: dict[str, Device] = field(default_factory=dict)
    plan_functions: dict[str, PlanGenerator] = field(default_factory=dict)

    _reference_cache: dict[type, type] = field(default_factory=dict)

```
[...]

Issue #363, opened 2024-02-05
Remote git server as source for plans and devices


callumforrester:
# Remote Git Repository as Source for Plans and Devices

## Background

We introduced a scratch area in https://github.com/DiamondLightSource/blueapi/pull/313 so that plan/device code could be loaded from external files on startup. Typically we check out these files on a shared filesystem for prototyping. We have found several issues with this approach:

- Longer startup times as the code is loaded in
- Blueapi is difficult to hot-reload (hopefully fixed with https://github.com/DiamondLightSource/blueapi/issues/317)
- The shared files may get conflicting edits by multiple people
    - Especially confusing if one person left uncomitted changes
    - May also lead to permissions issues if someone creates, and owns, a new file
- It can be hard to restore blueapi to a known state without deleting WIP code in the scratch area

We are therefore looking for a more flexible, long-term solution that will enable more robust deployments at the facility level. 

### Outline

Add the option to configure blueapi to pull down and install git repositories into its environment on startup. Define a "workspace" i.e. a group of remote git repositories that blueapi uses. On startup/reload, it will pull these as-configured (latest `main`, latest tag, some specific branch or hash). It can be configured via REST endpoints. It will also cache what it pulls in case the remote goes down.

![blueapi-code-upload-simple(1)](https://github.com/DiamondLightSource/blueapi/assets/29771545/6cd4189b-cd04-4bf3-8c20-c50996c5a325)

### Example Workflow for Prototyping Changes

An example workflow for prototyping changes to plans/devices is therefore:
- Checkout the relevent repo
- Make a new branch
- Make some experimental changes and push
- Point blueapi at the new branch

This adds some extra overhead to get experimental changes in, however they can be addressed outside of blueapi. For example, we have considered an auto-commit program for less advanced users. From their perspective they click "save" and a few seconds later the changes are live.

### Benefits

- We can track ongoing changes to deployments at the facility level
- No shared checkouts are being edited
- There is always the option of easily restoring to a working state (`main` branch)
- Caching reduces startup times most of the time

### Outstanding Concerns

- Is this too complex?
- It's risky to depend on remote github/gitlab repositories for runtime operation (solvable by hosting an intermediate git server, but that adds yet more complexity)
- Are we just using git as a database?

stan-dot:
re: 'point blueapi at the new branch' 
how would this be done? with a config value? there is no 'config' area in the current CLI.

re: risk
[...]

PR #515, opened 2024-06-20
Support lazy devices

stan-dot:
Outlined the places where the code needs to be modify to support the requested features.

Some alignment must happen on what is the best format to store the devices that are lazy. 

I'd rather do it in the start of the PR. 
#363 and #504  would change how this works however I think that this PR is mergeable much earlier than that.

Then, on my end I think the lazy devices would be best presented as:
-    early_devices: dict[str, Device] = field(default_factory=dict)
-    lazy_devices: dict[str, Device] = field(default_factory=dict)
properties on the Context object.

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -31,20 +31,6 @@ LOGGER = logging.getLogger(__name__)
 DevicesDict = dict[str, Device]
 
 
-def bisect_devices_dict(input_dict: DevicesDict) -> tuple[DevicesDict, DevicesDict]:
-    lazy_dict = {}
-    non_lazy_dict = {}
-
-    for key, value in input_dict.items():
-        # Assuming 'lazy' is an attribute or a condition you can check
-        if hasattr(value, "lazy") and value.lazy:
-            lazy_dict[key] = value
-        else:
-            non_lazy_dict[key] = value
-
-    return (non_lazy_dict, lazy_dict)
-
-
 @dataclass
 class BlueskyContext:
     """
@@ -126,7 +112,7 @@ class BlueskyContext:
         devices, exceptions = make_all_devices(module, **kwargs)
 
         # for non-lazy devices, we instantiate them
-        early_devices, _ = bisect_devices_dict(devices)
+        early_devices = devices.items().filter(lambda x: not x.lazy)
 
         for device in early_devices.values():
             self.register_device(device)
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -42,6 +42,7 @@ class BlueskyContext:
     )
     plans: dict[str, Plan] = field(default_factory=dict)
     devices: dict[str, Device] = field(default_factory=dict)
+    # todo add some format to keep the lazy vs non-lazy devices
     plan_functions: dict[str, PlanGenerator] = field(default_factory=dict)
 
     _reference_cache: dict[type, type] = field(default_factory=dict)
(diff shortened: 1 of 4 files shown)


callumforrester:
PR title

src/blueapi/core/context.py
```diff
@@ -116,7 +139,7 @@ def with_dodal_module(self, module: ModuleType, **kwargs) -> None:
             )
             LOGGER.exception(NotConnected(exceptions))
 
-    def plan(self, plan: PlanGenerator) -> PlanGenerator:
+    def register_plan(self, plan: PlanGenerator) -> PlanGenerator:
```
callumforrester, line 142: I don't object to this renaming, but will this or `register_device` break hyperion? @DominicOram @olliesilvester

src/blueapi/core/context.py
```diff
@@ -97,16 +114,22 @@ def plan_2(...) -> MsgGenerator:
...
+    def get_lazy_devices(self) -> DevicesDict:
+        return {k: v for k, v in self.devices.items() if v.lazy}
+
     def with_dodal_module(self, module: ModuleType, **kwargs) -> None:
         devices, exceptions = make_all_devices(module, **kwargs)
 
-        for device in devices.values():
-            self.device(device)
+        # for non-lazy devices, we instantiate them
+        early_devices, _ = bisect_devices_dict(devices)
```
callumforrester, line 129: This seems to be the only place you use the bisected dictionaries, in which case, is there any point in bisecting and storing two dictionaries? Could you just have one big dictionary and filter for eager devices here?

callumforrester (changes requested):

src/blueapi/core/context.py
olliesilvester: I don't think we use this in Hyperion yet, should be fine

stan-dot:
Todo: just fix tests, then push and mark PR as ready

stan-dot:
I commented in the ophyd-async repo towards the resolution of the test errors here

coretl:
The relevant ophyd-async ticket is https://github.com/bluesky/ophyd-async/issues/387, not https://github.com/bluesky/ophyd-async/issues/413.

The short term fix is to cap `numpy<2.0.0` like https://github.com/bluesky/ophyd-async/pull/388.

src/blueapi/core/context.py
```diff
@@ -97,16 +100,22 @@ def plan_2(...) -> MsgGenerator:
...
+    def get_lazy_devices(self) -> DevicesDict:
+        return {k: v for k, v in self.devices.items() if v.lazy}
+
     def with_dodal_module(self, module: ModuleType, **kwargs) -> None:
         devices, exceptions = make_all_devices(module, **kwargs)
 
-        for device in devices.values():
-            self.device(device)
+        # for non-lazy devices, we instantiate them
+        early_devices = devices.items().filter(lambda x: not x.lazy)
```
callumforrester, line 115: Nit: For consistency I suggest we use the blanket term "eager" in place of "early" and "non-lazy"

stan-dot: done

stan-dot:
@callumforrester so we need to support both ophydv1 and ophyd-async devices?

stan-dot:
image
@callumforrester  why do we import this as a sum of interfaces? I thought this kind of thing already exists inside bluesky?

stan-dot:
I just realized that there is no clear way to see if a device is lazy or not. this might need to wait until https://github.com/DiamondLightSource/dodal/issues/483 is done

callumforrester:
@stan-dot yes both should be supported for now, dodal is going to fully move off ophys v1 eventually, but probably not soon

stan-dot:
closing as out of date - the change of blueapi plan to register_plan is unrelated. furthermore no PR is meaningful here until the dodal issue is done

https://github.com/DiamondLightSource/dodal/issues/483

Issue #399, opened 2024-03-19
Remove Startup Script


callumforrester:
The [startup script](https://github.com/DiamondLightSource/blueapi/blob/main/container-startup.sh) performs an editable `pip install` of every directory in scratch and then starts blueapi. This was needed because the `pip install` needed to happen before the process containing the run engine started. Following #343, the RE is run in a subprocess, which means the pip install can be performed from the main process rather than a script.

We've seen some reliability issues with the script which this may improve, but the primary business need is paving the way to auto-reloading things in scratch without needing to restart the entire blueapi process (just the subprocess).

Tagging @joeshannon in case he can think of any other issues.

stan-dot:
is this just waiting for #365  then?

callumforrester:
@stan-dot no, this can be done at any time, documentation would be nice but is not blocking

PR #520, opened 2024-06-24
Automatically init scratch area and remove startup script

### LINKED ISSUE #510: InitContainer to set up scratch area
We should use [initContainers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) to set up the scratch area, i.e. clone a series of repositories if they do not already exist.

## Acceptance Criteria

- Blueapi starts on kubernetes with an initContainer that reads a list of required repos (defined in `values.yaml`)
- The container clones any repos that are not already present
- If cloning fails (e.g. because of a network error), then blueapi should still start up

callumforrester:
Fixes #510 and #399 

Write python code to do the job of `container-startup.sh` and add code to clone the required repos. Replace `container-startup.sh`  with this code. Stop blueapi from initializing scratch automatically on container startup, replace this job with an [initContainer](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/). Copy the blueapi venv to an [emptyDir](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) so the initContainers can manipulate it.


callumforrester:
This will need testing on k8s

helm/blueapi/templates/deployment.yaml
```diff
@@ -41,6 +41,31 @@ spec:
...
           type: Directory
       {{- end }}
+      - name: venv
+        emptyDir:
+          sizeLimit: 5Gi
+      initContainers:
+      {{- if .Values.scratch.hostPath }}
+      - name: init-venv
+        image: busybox:1.28
+        command: ['sh', '-c', "cp -r /venv /emptyDir/venv"]
```
callumforrester, line 51: My current concern with this is the added startup time

src/blueapi/cli/scratch.py
```diff
@@ -0,0 +1,72 @@
...
+from blueapi.config import ScratchConfig
+
+
+def setup_scratch(config: ScratchConfig) -> None:
+    _validate_directory(config.root)
+
+    logging.info(f"Setting up scratch area: {config.root}")
+
+    # Set umask to DLS standard
+    os.umask(0o002)
```
joeshannon, line 17: Is the `o` meant to be here?

src/blueapi/config.py
```diff
@@ -85,6 +85,16 @@ class RestConfig(BlueapiBaseModel):
     protocol: str = "http"
 
 
+class ScratchRepository(BlueapiBaseModel):
```
joeshannon, line 88: Eventually we may want to specify a branch too?

I suppose at this stage people will still be editing the scratch directory directly for modifications so it's probably good that the init doesn't mess with the working tree if the repo is already present.

src/blueapi/cli/scratch.py
callumforrester: Umask takes octal values, right? So this is octal 002, should it actually be `0o0002`?

src/blueapi/cli/scratch.py
```diff
@@ -0,0 +1,72 @@
...
+def ensure_repo(
+    remote_url: str,
+    local_directory: Path,
+) -> None:
+    if not local_directory.exists():
+        logging.info(f"Cloning {remote_url}")
+        Repo.clone_from(remote_url, local_directory)
+        logging.info(f"Cloned {remote_url} -> {local_directory}")
+    elif local_directory.is_dir():
+        Repo(local_directory)
```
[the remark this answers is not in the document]
callumforrester: Yes, it should raise an error if there's no `.git` directory

src/blueapi/config.py
callumforrester: I went back and forth on this, yeah, not even sure if we want a branch. If we're aiming for a world of automatic commits, a linear history is the easiest thing to manage...

src/blueapi/cli/scratch.py
joeshannon: Yeah you're correct - I've just been so used to calling it from bash as a "normal" number that it surprised me.

callumforrester:
Still to do:
- Cloned repos have wrong permissions
- If cloning/pip install errors, container starts anyway, do we want this?

tests/cli/test_scratch.py
```diff
@@ -0,0 +1,215 @@
...
+        with file_path.open("w") as stream:
+            stream.write("foo")
+
+    mock_repo.clone_from.side_effect = lambda url, path: write_repo_files()
+
+    ensure_repo("http://example.com/foo.git", repo_root)
+    assert file_path.exists()
+    assert file_path.is_file()
+    st = os.stat(file_path)
+    # For some reason the mode has two extra bytes at the front
```
joeshannon, line 127:
```suggestion
```

tests/cli/test_scratch.py
```diff
@@ -0,0 +1,215 @@
...
+    assert (st.st_mode & 0x00FF) & stat.S_IWGRP
```
joeshannon, line 128:
```suggestion
    assert st.st_mode & stat.S_IWGRP
```
Think the higher octets can be ignored.

src/blueapi/config.py
joeshannon: As discussed this is sufficient for now.

joeshannon (approved):

Issue #518, opened 2024-06-24
Add ability to enumerate instances of types to returned JSONSchemas


keithralphs:
At the moment on the /plans and /plans/(name} endpoints, the returned JSONSchemas that describe the plan(s) provide the types of objects with bluesky interfaces (e.g. detectors, axes etc.) that represent items bluesky talks to rather than the enumerated names of these items. This is perfectly reasonable and suitable for the command line inteface, GDA etc., but is not ideal for generating basic UIs from plan descriptions via e.g. JSONForms. For this use case, a list of the names of the devices avaliable on the beamline rather than their type is what is required so that these can be automatically presented to the user for selection to be passed back when the form is submitted.

The capablity to present the information in this form should be added to the plans endpoints either as an option (.g. ?enumerated=true) or as a separate version of the endpoints to allow both the current and desired functionality to be supported.  The extra information should probably be provided as well as rather than instead of the class names. This will allow the addition of a new plan to the beamline to automatically extend the range of basic web UIs available as well providing richer information for future fully fledged UIs, whilst still supporting existing interfaces.

callumforrester:
> to allow both the current and desired functionality to be supported

Is there a use case for the current functionality? If not I say we drop it and just have the desired functionality.

keithralphs:
> > to allow both the current and desired functionality to be supported
> 
> Is there a use case for the current functionality? If not I say we drop it and just have the desired functionality.

Dunno, I just assumed it was that way because the current CI used it, if not then maybe.

callumforrester:
I think it is just the simplest possible implementation

PR #522, opened 2024-06-25
Correct generated schema for Reference types

tpoliaw:
I think this resolves the plan schemas returned from the plans endpoint. Do we
want to include some field in the schema to mark that the field should be a
reference rather than the value itself?

### DIFF
--- a/tests/service/test_rest_api.py
+++ b/tests/service/test_rest_api.py
@@ -110,7 +110,7 @@ def test_get_plan_with_device_reference(handler: Handler, client: TestClient) ->
                         "title": "Delay",
                     },
                     "detectors": {
-                        "items": {"type": "<class " "'bluesky.protocols.Readable'>"},
+                        "items": {"type": "bluesky.protocols.Readable"},
                         "title": "Detectors",
                         "type": "array",
                     },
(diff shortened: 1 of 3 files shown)


tpoliaw:
Should the type value change as well? The `<class 'bluesky.protocols.Readable'>` form is not that useful - would it be better as `bluesky.protocols.Readable`?

tests/service/test_rest_api.py
```diff
@@ -110,9 +110,7 @@ def test_get_plan_with_device_reference(handler: Handler, client: TestClient) ->
                         "title": "Delay",
                     },
                     "detectors": {
-                        "items": {
-                            "_detectors": "<class " "'bluesky.protocols.Readable'>"
-                        },
+                        "items": {"type": "<class " "'bluesky.protocols.Readable'>"},
```
DiamondJoseph, line 113: I think long term we probably want `"type": "bluesky.protocols.Readable"`, such that we can `GET /devices?type=bluesky.protocols.Readable`

callumforrester (approved):
LGTM, thanks!

> Do we
want to include some field in the schema to mark that the field should be a
reference rather than the value itself?

Possibly, but let's do that in another PR, could also be done as part of #518

DiamondJoseph:
We can infer it from whether the type is a known type or otherwise referenced in the schema.

Any time we have as a plan parameter a BaseModel/Dataclass/TypedDict, I think the generated schema will include `$ref` and the full definition of that type (although we should check). Would we ever be passing in something that is not either a default JSON type, a BaseModel/Dataclass/TypedDict or a device reference?

DiamondJoseph (approved):

PR #524, opened 2024-06-25
Ignore PyCharm .idea directory

tpoliaw:


callumforrester (approved):
We'll get you yet :grin:

PR #526, opened 2024-06-25
feat: Enumerate devices in plan schemas

tpoliaw:
When a plan is expecting a device name, there is a limited number of
valid devices available to choose from. Making these available in the
schema allows clients to restrict input to valid options only.

Addresses #518

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -280,6 +280,11 @@ class BlueskyContext:
                     json_schema = handler(core_schema)
                     json_schema = handler.resolve_ref_schema(json_schema)
                     json_schema["type"] = qualified_name(target)
+                    json_schema["enum"] = [
+                        name
+                        for name, device in self.devices.items()
+                        if isinstance(device, cls.origin or target)
+                    ]
                     if cls.args:
                         json_schema["types"] = [qualified_name(arg) for arg in cls.args]
                     return json_schema
--- a/tests/system_tests/plans.json
+++ b/tests/system_tests/plans.json
@@ -8,7 +8,18 @@
                 "properties": {
                     "detectors": {
                         "items": {
-                            "type": "bluesky.protocols.Readable"
+                            "type": "bluesky.protocols.Readable",
+                            "enum": [
+                                "sample_pressure",
+                                "x_err",
+                                "theta",
+                                "z",
+                                "y",
+                                "x",
+                                "current_det",
+                                "image_det",
+                                "sample_temperature"
+                            ]
                         },
                         "title": "Detectors",
                         "type": "array"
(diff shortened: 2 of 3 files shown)


tpoliaw:
Example schema (for `set_relative`)

```
{
    'additionalProperties': False,
     'properties': {'group': {'title': 'Group',
                              'type': 'string'},
                    'movable': {'enum': ['sample_pressure',
                                         'x_err',
                                         'theta',
                                         'z',
                                         'y',
                                         'x',
                                         'sample_temperature'],
                                'target_type': 'bluesky.protocols.Movable',
                                'title': 'Movable',
                                'type': 'string'},
                    'value': {'title': 'Value'},
                    'wait': {'title': 'Wait',
                             'type': 'boolean'}},
     'required': ['movable'],
     'title': 'set_relative',
     'type': 'object'}
}
```

src/blueapi/core/context.py
```diff
@@ -201,9 +201,18 @@ def valid(cls, value):
...
                 ):
+                    target_type = f"{target.__module__}.{target.__qualname__}"
                     if field:
                         field_schema.update(
-                            {"type": f"{target.__module__}.{target.__qualname__}"}
+                            {
+                                "type": "string",
+                                "enum": [
+                                    k
+                                    for k, v in self.devices.items()
```
callumforrester, line 211: I don't believe there is any guaranteed order for registration of plans and devices, meaning that some devices may not be registered yet.

tpoliaw: Isn't this method only called when the schema is called for? Does anything query schemas before everything is loaded?

keithralphs:
> Example schema (for `set_relative`)
> 
> ```
> {
>     'additionalProperties': False,
>      'properties': {'group': {'title': 'Group',
>                               'type': 'string'},
>                     'movable': {'enum': ['sample_pressure',
>                                          'x_err',
>                                          'theta',
>                                          'z',
>                                          'y',
>                                          'x',
>                                          'sample_temperature'],
>                                 'target_type': 'bluesky.protocols.Movable',
>                                 'title': 'Movable',
>                                 'type': 'string'},
>                     'value': {'title': 'Value'},
>                     'wait': {'title': 'Wait',
>                              'type': 'boolean'}},
>      'required': ['movable'],
>      'title': 'set_relative',
>      'type': 'object'}
> }
> ```

This seems to have an extra curly bracket in it (according to JSON validation) , typo maybe?, also shouldn't the False in additional properties be false? Apart from them it seems to render correctly in the browser via JSON when these are corrected.

tpoliaw:
Copy/paste/reformat error on my part I think copying from the output. Is the json returned in the response correct?

keithralphs:
> Copy/paste/reformat error on my part I think copying from the output. Is the json returned in the response correct? And yes, I copied it from the test which uses Python

Haven't tried it dynamically yet, just copy/pasted it myself, corrected the errors and checked :). I'll run it up later and test it live.

src/blueapi/core/context.py
callumforrester: No, the schemas are generated upfront when the plan is registered with the context. I don't mind if you make them lazy instead though.

callumforrester (approved):

Issue #516, opened 2024-06-21
Test JINJA auto-generated GUIs


stan-dot:
Would approximate something like [this](https://github.com/renceInbox/fastapi-todo). Much more friendly to Python developers, as we already use Go-template syntax inside Helm charts and Python. Using React brings problems

stan-dot:
should this be closed in favor of the react setup (#536 ) ? 

@callumforrester @keithralphs ?

callumforrester:
Yes, thanks!

PR #529, opened 2024-06-27
516 test jinja auto generated guis

stan-dot:


callumforrester:
@stan-dot can this be closed along with #516?

PR #533, opened 2024-07-01
Use equality instead of identity to compare types

tpoliaw:
In some cases, possibly when MsgGenerator is defined or imported from
different modules, the types can be equivalent but are not the same
instance. In these cases, the identity check would fail and some plans
would be excluded where they shouldn't be. Using equality checks instead
mean it shouldn't matter where the MsgGenerator type was defined as all
instances should be equal.

### DIFF
--- a/src/blueapi/core/bluesky_types.py
+++ b/src/blueapi/core/bluesky_types.py
@@ -75,7 +75,7 @@ def _follows_bluesky_protocols(obj: Any) -> bool:
 
 def is_bluesky_plan_generator(func: PlanGenerator) -> bool:
     try:
-        return get_type_hints(func).get("return") is MsgGenerator
+        return get_type_hints(func).get("return") == MsgGenerator
     except TypeError:
         # get_type_hints fails on some objects (such as Union or Optional)
         return False


tpoliaw:
_copied from the Jira ticket_

> RE the acceptance criteria, I don't think the imports need to be changed. I could add tests to check that functions defined with both versions of MsgGenerator are included but given in some cases, the different imports still resolve to the same type, I'm not sure they're going to actually test the failure case.

DiamondJoseph (approved):

callumforrester (approved):
Thanks for looking at this!

DiamondJoseph:
This issue is now extending to equality when deployed on i22.

```
In [25]: for obj in load_module_all(mod):
    ...:         print(f"{obj.__annotations__['return']}")
    ...: 
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]
collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]

In [26]: from dodal.common import MsgGenerator as m1

In [27]: m1
Out[27]: collections.abc.Generator[bluesky.utils.Msg, typing.Any, None]

In [28]: from dls_bluesky_core.core import MsgGenerator as m2

In [29]: m2
Out[29]: typing.Generator[bluesky.utils.Msg, typing.Any, NoneType]

In [30]: for obj in load_module_all(mod):
    ...:         print(f"{obj.__annotations__['return'] == m2}")
    ...: 
False
False
False
False
False
False
False
False

In [31]: for obj in load_module_all(mod):
    ...:         print(f"{obj.__annotations__['return'] == m1}")
    ...: 
True
True
True
True
True
True
True
True

In [32]: 
```

PR #538, opened 2024-07-04
Remove handler

### LINKED ISSUE #434: Get rid of Handler
As a developer I find the `Handler` super-object confusing and difficult to work with. Blueapi should be made up of individual components, a bit like a spring boot application so it is easier to develop.

Now we run the worker in a subprocess (#317), information goes through a long chain of redundant calls including two implementations of `Handler` that we can probably simplify, and get rid of the `Handler` objects in the first place.

Probably also makes #371 easier.

## Possible Solution

Currently we have:

```mermaid
sequenceDiagram
    box Blueapi
    participant FastAPI Endpoint
    participant SubprocessHandler
    end

    box Worker Process
    participant Free Function
    participant Handler
    participant Worker
    end
    
    FastAPI Endpoint->>SubprocessHandler: query?
    SubprocessHandler->>Free Function: query?
    Free Function->>Handler: query?
    Handler->>Worker: query?
    Worker->>Handler: response
    Handler->>Free Function: response
    Free Function->>SubprocessHandler: response
    SubprocessHandler->>FastAPI Endpoint: response

    note over Handler:Redundant
    note over SubprocessHandler:Redundant
```

Both handlers are just forwarding calls at this point and can probably go, leaving us with:

```mermaid
sequenceDiagram

    box Blueapi
    participant FastAPI Endpoint
    end

    box Worker Process
    participant Free Function
    participant Worker
    end
    
    FastAPI Endpoint->>Free Function: query?
    Free Function->>Worker: query?
    Worker->>Free Function: response
    Free Function->>FastAPI Endpoint: response
```

The complexity will be in the `setup_handler` code which creates a handler and configures the objects within it. That should move and should be more separated out rather than in one place. My suggestion would be to have `@lru_cache`ed functions to create each component (e.g. `Worker`, `MessageTemplate`, ...).

The other area of complexity will be updating the many tests that mock out `Handler`, but those tests are no longer accurate anyway since they pre-date the `SubprocessHandler`. Realistically a lot of them will need rewrites. It may be easier to do #360 first to make the tests less intrusive in the way they poke `Handler`.

## Acceptance Criteria
- `BlueskyHandler`, `SubprocessHandler` and `Handler` deleted
- Tests updated
- Tests cover subprocess case
- Any references to `Handler` in docs removed

joeshannon:
Remove handler

The handler has caused additional complexity with multiple levels of
indirection.

Remove the ABC and include the free functions in an "interface" module.

The main app and cli tests are rewritten to look more like unit tests.

This PR is to be followed up by further refactoring to replace the
singleton class with cached functions and restructuring the modules.

Fixes #434.


src/blueapi/service/interface.py
```diff
@@ -0,0 +1,205 @@
...
+
+"""This module provides interface between web application and underlying Bluesky
+context and worker"""
+
+
+class _Singleton:
+    context: BlueskyContext
+    worker: Worker
+    messaging_template: MessagingTemplate | None = None
+    initialized = False
```
callumforrester, line 27: I was experimenting to see if we could make our earlier idea of using `lru_cached` functions work, see here https://github.com/DiamondLightSource/blueapi/pull/557

I think this fixes the testing issue by clearing the caches on teardown?

src/blueapi/service/runner.py
```diff
@@ -0,0 +1,112 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+def _init_worker():
+    # Replace sigint to allow subprocess to be terminated
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
+
+
+class Runner:
```
callumforrester, line 25: I would prefer this have a longer, more descriptive name, maybe `WorkerProcessDispatcher`?

src/blueapi/service/interface.py
callumforrester, line 1: How about the following structure?

- `src`
  - `blueapi`
    - `service`
      - `rest`
        - `main`
        - `openapi`
      - `worker`
        - `main` (formerly `interface`)
        - `runner`
        - `model`

callumforrester (changes requested):

src/blueapi/service/interface.py
joeshannon: I think it will help yes, have merged.

src/blueapi/service/runner.py
joeshannon: Yeah I did try this although it is used a lot in already very long function definitions in the rest app main.py and made them a bit more unwieldy, or that might have just been me. Could go with `WorkerDispatcher` for now?

src/blueapi/service/interface.py
joeshannon: As discussed today - this is a good idea but I would prefer to merge the current PR first and move towards this (and the lru_cache) iteratively.

src/blueapi/service/runner.py
callumforrester: Yes please

src/blueapi/service/interface.py
callumforrester: Can you make issues for the next improvements?

joeshannon: Created #567 and #568.

src/blueapi/service/runner.py
joeshannon: Done

callumforrester:
Had a play locally, when I reload the environment it doesn't appear to import new changes.

callumforrester (approved):
LGTM!

PR #539, opened 2024-07-05
Extend formatting options to listen subcommand

tpoliaw:
Listen streams a mix of WorkerEvent, ProgressEvent and DataEvent
instances for anything that the server is doing. This extends the type
handling in the format to each of these types to display the relevant
level of information for each.

Informational messages are changed to be printed to stderr so that the
main stdout output can be piped to an external application (eg jq for
json mode).

### DIFF
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import sys
 from functools import wraps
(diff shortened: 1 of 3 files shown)


callumforrester (approved):
LGTM, thanks

callumforrester (approved):

callumforrester (approved):

PR #548, opened 2024-07-08
Refactor client logic out of CLI

### LINKED ISSUE #547: Move client logic out of CLI
The CLI module currently contains a lot of logic that can control blueapi using a mixture of REST and listening to STOMP messages. It would be useful to move this logic out into its own class so it is reusable. This refactor could pave the way for 
- #535 
- #485 
- #489 
- #472 
- General hope of having a Python client for scripting the new architecture

callumforrester:
Fixes #547


DiamondJoseph (approved):

callumforrester:
Some of the coverage issues might be helped by #538, so will defer until then

Issue #273, opened 2023-07-05
Update to Pydantic2


DiamondJoseph:
Requires Dodal to have removed their pinning to v1.

PR #549, opened 2024-07-09
Configure Dependabot to respect pinned dependencies

DiamondJoseph:
Doubly pins some dependencies (in pyproject.toml and in dependabot.yaml) so that these pins will not be automatically removed by dependabot, and dependabot PRs can be consumed without losing the purposeful pinning of known incompatibilities.

See also #273 
TODO: Make issues for the other known problems to track when they are capable of being unpinned

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,3 +22,6 @@ updates:
       dev-dependencies:
         patterns:
           - "*"
+    ignore:
+      - dependency-name: "pydantic"
+        versions: [">=2.0"]


coretl (commented):
Is the double pinning needed to stop dependabot generating an inconsistent set (so pip won't even install them as their dependencies clash) or a runtime incompatible set (so pip can install them but tests will fail)?

DiamondJoseph:
> Is the double pinning needed to stop dependabot generating an inconsistent set (so pip won't even install them as their dependencies clash) or a runtime incompatible set (so pip can install them but tests will fail)?

The former, it means that our dev-requirements.txt shouldn't update to or past the versions that we know are incompatible in ways that aren't caught by tests (e.g. we were pinning fastapi so that the codegen swagger Java plugin was not causing issues with the enum handling: we are now moving away from pinning that particular dependency but those sorts of things)

pyproject.toml
```diff
@@ -17,15 +17,15 @@ dependencies = [
     "nslsii",
     "pyepics",
     "aioca",
-    "pydantic<2.0",
+    "pydantic<2.0",      # Pinned by our required fastapi version, leave pinned until can check incompatibility
     "stomp-py",
     "aiohttp",
     "PyYAML",
-    "click<8.1.4",
+    "click",
```
stan-dot, line 24: nit: why click no longer pinned?

pyproject.toml
```diff
@@ -17,15 +17,15 @@ dependencies = [
...
     "fastapi[all]<0.99", # Later versions use a newer openapi schema, which is incompatible with swagger see https://github.com/swagger-api/swagger-codegen/issues/10446
     "uvicorn",
     "requests",
-    "dls-bluesky-core",  #requires ophyd-async
+    "dls-bluesky-core",
```
stan-dot, line 28: nit: isn't that deprecated soon? shouldn't it have a comment like pydantic does?

stan-dot (approved):

pyproject.toml
DiamondJoseph: Click was pinned for an issue with private methods that is no longer relevant in the latest version, I've checked and we're able to now unpin it.

pyproject.toml
DiamondJoseph: It is deprecated and will be removed soon, the comment just isn't relevant anymore as the dependency on ophyd-async (also) comes from dodal

stan-dot: ok but a new comment wouldn't hurt

Issue #553, opened 2024-07-10
Fix compatibility with bluesky 1.13.0a4 (super-state-machine) removal


joeshannon:
Bluesky 1.13.0a4 contains a change (https://github.com/bluesky/bluesky/pull/1708) which removes `super-state-machine` as a dependency and instead bundles it in a private vendor package. This causes blueapi failure as it was relying on the transitive dependency.

The following components are currently used in blueapi
```
super_state_machine.errors.TransitionError
super_state_machine.extras.PropertyMachine
super_state_machine.extras.ProxyString
```

A couple of quick fixes include adding `super-state-machine` as a direct blueapi dependency or using the private package from bluesky e.g.
```
from bluesky._vendor.super_state_machine.extras import PropertyMachine
```

joeshannon:
It has now been added as a direct dependency.

This issue can now be closed as #555 has been made to eventually remove it.

PR #554, opened 2024-07-10
Add super-state-machine as direct dependency

joeshannon:
This was previously available as a transitive dependency from bluesky but has now been removed (bluesky/bluesky#1708).

This is a temporary workaround to make the build work again and may be reverted if the types are made available from bluesky in future.

See #553.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,7 @@ dependencies = [
     "requests",
     "dls-bluesky-core",  #requires ophyd-async
     "dls-dodal>=1.24.0",
+    "super-state-machine", # See GH issue 553
 ]
 dynamic = ["version"]
 license.file = "LICENSE"


callumforrester:
I think this is the right thing to do, I think we also need to raise an issue with [`bluesky`](https://github.com/bluesky/bluesky) to either expose these types publically or completely remove them and make the `RE` state a `str` or `Enum` or similar.

This is only used for type annotations, i.e. the thing returned by `RE.state` may be a `super-state-machine` type, which is why we need them, so we can properly annotate our code when we [ask the run engine for its state](https://github.com/DiamondLightSource/blueapi/blob/97b43cec747b22c94de266e05db286ba98f81aba/src/blueapi/worker/event.py#L11)

DiamondJoseph (approved):

PR #557, opened 2024-07-11
Replace subprocess singleton with lru_cached functions

callumforrester:


src/blueapi/service/interface.py
```diff
@@ -20,52 +18,81 @@
...
+    return _CONFIG
+
+
+def set_config(new_config: ApplicationConfig):
+    global _CONFIG
+
+    _CONFIG = new_config
+
+
+@lru_cache
```
DiamondJoseph, line 39:
```suggestion
@lru_cache(maxsize=1)
```

Is this clearer for what the intended behaviour is?

src/blueapi/service/interface.py
```diff
@@ -75,131 +102,93 @@ def _publish_event_streams(streams_to_destinations: Mapping[EventStream, str]) -
...
 
 
-def _ensure_worker_started() -> None:
-    if _Singleton.initialized:
-        return
-    raise Exception("Worker must be stared before it is used")
+def health_check() -> Mapping[str, bool]:
+    """Health of various components, returns a mapping of
+    component name to whether it is up and running"""
+    return ...
```
DiamondJoseph, line 194: Seems to be unused, is this for a seperate PR?

src/blueapi/service/runner.py
```diff
@@ -30,83 +28,81 @@ class Runner:
...
+            )
+            LOGGER.exception(e)
 
     def run(self, function: Callable, arguments: Iterable | None = None) -> Any:
         if self._use_subprocess:
             return self._run_in_subprocess(function, arguments)
         else:
-            if arguments is None:
-                arguments = []
+            arguments = arguments or []
```
DiamondJoseph, line 87: this function should be refactored to `def run (self, function: Callable[[P], T], arguments: P) -> T:`
Where P = ParamSpec("P"), T = TypeVar("T") if possible

DiamondJoseph (changes requested):

src/blueapi/service/interface.py
```diff
@@ -20,52 +18,81 @@
 context and worker"""
 
 
-class _Singleton:
-    context: BlueskyContext
-    worker: Worker
-    messaging_template: MessagingTemplate | None = None
-    initialized = False
+@runtime_checkable
+class CacheClearable(Protocol):
```
joeshannon, line 22: I'm not sure how this is used but can leave it in for now.

src/blueapi/service/interface.py
joeshannon: Possibly although that parameter seems to be more relevant when the function has parameters. I think in this case it's ok as is.

src/blueapi/service/runner.py
```diff
@@ -30,83 +28,81 @@ class Runner:
...
+        self.stop()
+        self.start()
+        LOGGER.info("Runner reloaded")
 
     def start(self):
-        if self._subprocess is None and self._use_subprocess:
-            self._subprocess = Pool(initializer=_init_worker, processes=1)
-            self._subprocess.apply(
-                logging.basicConfig, kwds={"level": self._config.logging.level}
+        try:
```
joeshannon, line 54: This does change start/stop to not be idempotent, but perhaps that doesn't matter?

src/blueapi/service/runner.py
joeshannon: Possibly in the future.

src/blueapi/service/interface.py
```diff
@@ -20,52 +18,81 @@
...
         )
-        _Singleton.messaging_template.connect()
-    _Singleton.initialized = True
+        template.connect()
+        return template
+    else:
+        return None
+
+
+def setup(config: ApplicationConfig) -> None:
```
joeshannon, line 78: Do we want to prevent this being called twice (without teardown)?

I think if it was called a second time with a different config, the config would be set but the worker would still be cached and using the old config?

joeshannon (commented):
Thanks @callumforrester this looks much better. I'll merge this to my branch next week to give some time to address these comments.

src/blueapi/service/interface.py
callumforrester: I started going down the route of getting the main process to query the subprocess' health, but then, yes, decided that should be a separate issue.

src/blueapi/service/interface.py
callumforrester: It is literally for duck typing, I am at once asserting that these things have a `cache_clear` method and using it. It doesn't add much safety, but hopefully a bit of readability.

src/blueapi/service/runner.py
joeshannon: Should we be concerned about this at this stage?
Otherwise I'll probably carry on and it can be re-reviewed later.

callumforrester: Unsure, I agree it should be re-reviewed later, in the context of the main PR

PR #563, opened 2024-07-18
562 investigate splitting into entrypoints with a shared dockerfile base

stan-dot:
refactor around the worker and protobuf. not complete yet. 

based on `todo.md`

- [ ] can deploy one package just with different entrypoints, even the same dockerfile, just different entrypoints
- [ ] move the blueapi initalization logic away from CLI - ApplicationConifg is too self-indulgent
- [ ] all handler references should be replaced by the grpc client OR just the refernce to the context
- [ ] move the blueworker config types to proto
- [ ] no refernce to pydantic inside the worker

Keeping most of the context inside the worker (such as devices and plans) as the blueapi only needs parts of the plan. we can generate the json schema needed for autogeneration inside the worker and serialize it there. That way we contain the chaos of uploading arbitrary code to the worker only.

We'd represent the props to run a plan in a 
`message Plan{
    string name = 1;
    google.protobuf.Struct props = 2;
 }`


stan-dot:
this is not useful and not easily mergeable in the current state, a fresh approach on another day would be better

PR #565, opened 2024-07-19
Update schema test after count plan update

tpoliaw:
The count plan was updated to take a set instead of a list which meant
the generated schema changed to include the 'uniqueItems' key.


callumforrester (approved):

tpoliaw:
Looks like this has solved itself by being removed completely

PR #572, opened 2024-07-24
Replace interface singleton class with cached functions

joeshannon:
- [x] #570

### DIFF
--- a/src/blueapi/service/runner.py
+++ b/src/blueapi/service/runner.py
@@ -6,7 +6,10 @@ from multiprocessing.pool import Pool as PoolClass
 from typing import Any
 
 from blueapi.config import ApplicationConfig
-from blueapi.service.interface import InitialisationException, start_worker, stop_worker
+from blueapi.service.interface import (
+    setup,
+    teardown,
+)
 from blueapi.service.model import EnvironmentResponse
 
 # The default multiprocessing start method is fork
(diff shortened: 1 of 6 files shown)


src/blueapi/service/runner.py
```diff
@@ -32,51 +35,52 @@ class WorkerDispatcher:
...
     ) -> None:
         self._config = config or ApplicationConfig()
         self._subprocess = None
         self._use_subprocess = use_subprocess
-        self._state = EnvironmentResponse(initialized=False)
+        self._state = EnvironmentResponse(
+            initialized=False,
+        )
+
+    def reload(self):
```
callumforrester, line 49: Could: There is a functionality equivalent implementation in `main.py`, should probably just call this.

src/blueapi/service/runner.py
```diff
@@ -32,51 +35,52 @@ class WorkerDispatcher:
...
-                initialized=False, error_message=f"Error configuring blueapi: {e}"
+                initialized=False,
+                error_message=str(e),
             )
-            LOGGER.exception(self._state.error_message)
-            return
-        self._state = EnvironmentResponse(initialized=True)
+            LOGGER.exception(e)
 
     def stop(self):
```
callumforrester, line 68: I forget whether we came to any conclusion about whether or not this made `start`/`stop` non-idempotent and whether or not we were okay with that...

callumforrester (approved):

src/blueapi/service/runner.py
joeshannon: Done

src/blueapi/service/runner.py
joeshannon: I think for now we are ok with it. There is some risk that if things are called in the wrong order e.g. if `worker()` was called before `setup()` then things might be in an inconsistent state but I think we are ok with this risk for now.

PR #573, opened 2024-07-24
simplification refactor - canceling the worker abstract class

stan-dot:
rationale in the issue


src/blueapi/worker/task_worker.py
```diff
@@ -27,17 +32,29 @@
...
+    """
+
+    task_id: str
+    task: T
+    is_complete: bool = False
+    is_pending: bool = True
+    errors: list[str] = Field(default_factory=list)
+
+
+class TaskWorker:
```
callumforrester, line 57: Shall we just call it `Worker` now?

stan-dot: worker is a generic python process term. https://docs.python.org/3/library/multiprocessing.html#using-a-pool-of-workers

taskworker is clearer

callumforrester (approved):

Issue #559, opened 2024-07-15
Investigate using `helm-unittest` and `kubeconform`


callumforrester:
These two packages could be used to increase confidence when deploying changes to the helm chart, it is currently very hard to test changes:
- https://github.com/helm-unittest/helm-unittest
- https://github.com/yannh/kubeconform

DiamondJoseph:
Closed by #871

PR #582, opened 2024-08-02
Fix extraEnvVars in helm chart

### LINKED ISSUE #581: Extra env vars cause helm to crash
The `extraEnvVars` field in the helm chart causes helm to panic when not empty, for example

```yaml
extraEnvVars: |
 - name: BEAMLINE
   value: i04
```

leads to

```shell
install.go:222: [debug] Original chart version: ""
install.go:239: [debug] CHART PATH: /home/vid18871/projects/bluesky/blueapi/helm/blueapi


Error: template: blueapi/templates/deployment.yaml:97:27: executing "blueapi/templates/deployment.yaml" at <.Values.extraEnvVars>: can't evaluate field Values in type string
helm.go:84: [debug] template: blueapi/templates/deployment.yaml:97:27: executing "blueapi/templates/deployment.yaml" at <.Values.extraEnvVars>: can't evaluate field Values in type string
```

callumforrester:
Fixes #581 

`extraEnvVars` was being treated as a string in some places and a list in others. This PR makes it a list everywhere, any values files that look like this should be updated to remove the `|`:

```yaml
extraEnvVars: |
  - name: BEAMLINE
    value: i04
```

### DIFF
--- a/helm/blueapi/templates/deployment.yaml
+++ b/helm/blueapi/templates/deployment.yaml
@@ -92,10 +92,8 @@ spec:
             - "-c"
             - "/config/config.yaml"
             - "serve"
-          {{- with .Values.extraEnvVars }}
           env:
-            {{- tpl .Values.extraEnvVars . | nindent 10 }}
-          {{- end }}
+            {{- toYaml .Values.extraEnvVars | nindent 12 }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}


DominicOram:
Maybe naive question but is possible to have written a test that would have caught this? For hyperion we occasionally run our deploy script into `/tmp` for testing

callumforrester:
Yes, but I need to look into it further: https://github.com/DiamondLightSource/blueapi/issues/559

DiamondJoseph:
I think this was required for other edge cases where we're injecting something in values.yaml into extraEnvVars as part of the `tpl` call? Or else maybe if extraEnvVars is empty?

callumforrester:
Hmm, it may have been needed for rabbitmq password injection?

DiamondJoseph:
Looking at i22,

```
extraEnvVars: |
    - name: SCRATCH_AREA
      value: {{ .Values.scratch.containerPath }}
    - name: RABBITMQ_PASSWORD
      valueFrom:
        secretKeyRef:
          name: rabbitmq-password
          key: rabbitmq-password

```

Has this been checked if it worked with the {{.Values.scratch.containerPath}} injection? Or are we going to replicate the value here?

callumforrester:
With #520, the `$SCRATCH_AREA` variable is no longer needed,  I assume the tpl was because of `$RABBITMQ_PASSWORD`

callumforrester:
@DiamondJoseph and @DominicOram are we happy for this to go in?

DiamondJoseph (approved):

PR #584, opened 2024-08-02
Create very simple RPC so the subprocess loads functions

### LINKED ISSUE #590: Subprocess functions can't be picked when decorated for observability
In @keithralphs' attempt to add tracing in #586, errors occurred because the functions in `interface.py` are not loaded by the subprocess, they are pickled and sent over. When decorated for tracing they become unpicklable and cause an error. @keithralphs can fill in the details.

## Possible Solution

We can pass a function name to the subprocess and make it import and load the functions internally. If I'm honest, that's how I thought `multiprocessing` worked in the first place...

This is also a step in the right direction for https://github.com/DiamondLightSource/blueapi/issues/504

## Acceptance Criteria

- Functions are loaded inside the subprocess
- #586 is now achievable to @keithralphs and @DiamondJoseph's satisfaction

callumforrester:
Fixes #590

### DIFF
--- a/src/blueapi/service/runner.py
+++ b/src/blueapi/service/runner.py
@@ -1,10 +1,10 @@
 import logging
 import signal
-from collections.abc import Callable, Iterable
+from collections.abc import Callable
 from importlib import import_module
 from multiprocessing import Pool, set_start_method
 from multiprocessing.pool import Pool as PoolClass
-from typing import Any
+from typing import Any, ParamSpec, TypeVar
 
 from blueapi.config import ApplicationConfig
 from blueapi.service.interface import setup, teardown
@@ -14,6 +14,9 @@ from blueapi.service.model import EnvironmentResponse
 set_start_method("spawn", force=True)
 
 LOGGER = logging.getLogger(__name__)
+
+P = ParamSpec("P")
+T = TypeVar("T")
 
 
 def _init_worker():
@@ -54,7 +57,7 @@ class WorkerDispatcher:
         try:
             if self._use_subprocess:
                 self._subprocess = Pool(initializer=_init_worker, processes=1)
-            self.run(setup, [self._config])
+            self.run(setup, self._config)
             self._state = EnvironmentResponse(initialized=True)
         except Exception as e:
             self._state = EnvironmentResponse(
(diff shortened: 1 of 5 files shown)


DiamondJoseph (approved):
Approving for the _rpc function, if @callumforrester can approve for the other changes I made ;)

joeshannon:
Is there an associated issue for this? I can't quite infer what the problem was from this PR. Presumably pickle related?

DiamondJoseph:
> I can't quite infer what the problem was from this PR. Presumably pickle related?

I assume so from Callum's first commit (but surely still has to pickle the args, kwargs just not the function?), my interest in it is it makes upcoming observability handling easier

src/blueapi/service/runner.py
```diff
@@ -106,3 +111,24 @@ def state(self) -> EnvironmentResponse:
...
+
+
+def _rpc(
+    module_name: str, function_name: str, *args: P.args, **kwargs: P.kwargs
+) -> Any:
+    mod = import_module(module_name)
+    func: Callable[P, T] = _validate_function(
+        mod.__dict__.get(function_name, None), function_name
+    )
+    return func(*args, **kwargs)
```
callumforrester, line 126: @DiamondJoseph as discussed...

I don't think `P` is useful here, it doesn't bind the function and pyright complains about it. Because this is RPC the parameter and return types are not knowable at "compile" time, which means you will need to write runtime code to verify that your type hints reflect reality. I suggest allowing `args` and `kwargs` to be `Any` in this instance because verifying their type is hard and not _that_ useful:

```suggestion
def _rpc(
    module_name: str, function_name: str, *args: Any, **kwargs: Any
) -> Any:
    mod = import_module(module_name)
    func: Callable[P, T] = _validate_function(
        mod.__dict__.get(function_name, None), function_name
    )
    return func(*args, **kwargs)
```

What is easier, and probably more useful, is to validate the _return_ type at runtime:

```suggestion
def _rpc(
    module_name: str, function_name: str, expected_return_type: T, *args: P.args, **kwargs: P.kwargs
) -> T:
    mod = import_module(module_name)
    func: Callable[P, T] = _validate_function(
        mod.__dict__.get(function_name, None), function_name
    )
    value = func(*args, **kwargs)
    if isinstance(value, expected_return_type):
        return value
    else:
        raise TypeError()
```

PR #587, opened 2024-08-05
Move propagated context wrapper to method

DiamondJoseph:


DiamondJoseph:
Changes duplicated on target branch

PR #589, opened 2024-08-06
Replace references to blueapi messaging package 

### LINKED ISSUE #588: Extract publishing of documents etc. to message bus into a separate library
...via @coretl:

As a developer I would like to install bluesky locally and develop plans/devices with my own `RunEngine` in a python shell or similar so that I can debug plans effectively without having to connect to a remote server (blueapi). I also need to ensure that the plan provokes the correct response from the rest of the system, so I will need easy-to-import helper functions that allow me to link my `RunEngine` to a message bus in the same way blueapi's is linked.

## Possible Solution

Extract the `messaging` package to a separate library, similar to https://github.com/bluesky/bluesky-kafka and provide helper functions that both blueapi and a standalone user in a shell can easily use to authenticate and connect to the message bus.

callumforrester:
Replace references to blueapi messaging package with references to the bluesky stomp library

Fixes #588


.github/actions/install_requirements/action.yml
```diff
@@ -26,7 +26,11 @@ runs:
         python-version: ${{ env.PYTHON_VERSION }}
 
     - name: Install packages
-      run: pip install ${{ inputs.pip-install }}
+      run: pip install $([ -f dev-requirements.txt ] && echo '-c dev-requirements.txt') -e .[dev]
+      shell: bash
+
+    - name: Install bluesky_stomp
+      run: pip install git+https://github.com/DiamondLightSource/bluesky-stomp.git@main
```
ZohebShaikh, line 33: This is for ad-hoc CI testing

dev-requirements.txt
```diff
@@ -18,6 +18,7 @@ black==24.8.0
 bluesky==1.13.0a4
 bluesky-kafka==0.10.0
 bluesky-live==0.0.8
+bluesky-stomp @ git+https://github.com/DiamondLightSource/bluesky-stomp@4dbdb6b144b4b03243a1784f2f53dc13cbbef30e
```
ZohebShaikh, line 21: This is ad-hoc

tests/client/test_client.py
```diff
@@ -307,9 +307,11 @@ def test_run_task_sets_up_control(
 ):
     mock_rest.create_task.return_value = TaskResponse(task_id="foo")
     mock_rest.update_worker_task.return_value = TaskResponse(task_id="foo")
+    ctx = Mock()
+    ctx.correlation_id = "foo"
+    mock_events.subscribe_to_all_events = lambda on_event: on_event(COMPLETE_EVENT, ctx)
```
ZohebShaikh, line 312: I had to make this change to ensure that the task completes and doesn't hang indefinitely. I previously tried setting a short timeout, which allowed the test to finish, but it introduced an unintended dependency on `test_valid_stomp_config_for_listener` test. This is not ideal, as tests should be independent of each other. It seems the issue is related to the task handling.

```
        == "Subscribing to all bluesky events from localhost:61613\nPress enter to "
        + 'exit{\n  "state": "IDLE",\n  "task_status": null,\n  "errors": [],\n  '
        + '"warnings": []\n}\n'
```

pyproject.toml
```diff
@@ -91,7 +91,7 @@ filterwarnings = ["error", "ignore::DeprecationWarning"]
 # Doctest python code in docs, python code in src docstrings, test functions in tests
 testpaths = "docs src tests"
 markers = [
-    "handler: marks tests that interact with the global handler object in handler.py",
+    "stomp: marks tests that require the stomp"
```
stan-dot, line 94: `the stomp` is not optimal, `the stomp client` would fit better

dev-requirements.txt
DiamondJoseph: Did you have to make any changes to the bluesky-stomp library, or can we make a 0.1.0 on that repository?

src/blueapi/cli/cli.py
```diff
@@ -147,14 +147,20 @@ def listen_to_events(obj: dict) -> None:
     config: ApplicationConfig = obj["config"]
     if config.stomp is not None:
         event_bus_client = EventBusClient(
-            StompMessagingTemplate.autoconfigured(config.stomp)
+            MessagingTemplate.for_broker(
```
stan-dot, line 150: please rename it from Template, it's a client

stan-dot: even with `import as`

src/blueapi/client/event_bus.py
```diff
@@ -28,11 +30,11 @@ def __exit__(self, exc_type, exc_value, exc_traceback) -> None:
 
     def subscribe_to_all_events(
         self,
-        on_event: Callable[[MessageContext, AnyEvent], None],
+        on_event: Callable[[AnyEvent, MessageContext], None],
     ) -> None:
         try:
             self.app.subscribe(
-                self.app.destinations.topic("public.worker.event"),
+                MessageQueue(name="public.worker.event"),
```
stan-dot, line 37: this looks so much better now :rocket:

src/blueapi/service/interface.py
```diff
@@ -51,10 +52,14 @@ def worker() -> TaskWorker:
 def messaging_template() -> MessagingTemplate | None:
     stomp_config = config().stomp
     if stomp_config is not None:
-        template = StompMessagingTemplate.autoconfigured(stomp_config)
+        template = MessagingTemplate.for_broker(
```
stan-dot, line 55: as above

dev-requirements.txt
ZohebShaikh: I think we can release the bluesky-stomp library ,I did not make any changes on it

pyproject.toml
DiamondJoseph, line 94:
```suggestion
    "stomp: marks tests that require a message bus configured for stomp" 
```

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "aiohttp",
     "PyYAML",
     "click",
-    "fastapi[all]",
+    "fastapi>=0.112.0",
```
DiamondJoseph, line 27:
```suggestion
    "fastapi[all]>=0.112.0",
```

unless there was a refactor of the dependencies? I don't know what we required from [all]

tests/core/fake_device_module.py
```diff
@@ -33,7 +33,7 @@ def _mock_with_name(name: str) -> MagicMock:
 
 
 def wrong_return_type() -> int:
-    return "0"  # type: ignore
+    return "str"  # type: ignore
```
stan-dot, line 36: can we get a comment for this? where would we use this?

tests/service/test_interface.py
```diff
@@ -273,3 +274,11 @@ def test_get_task_by_id(context_mock: MagicMock):
 def test_stomp_config():
     interface.set_config(ApplicationConfig(stomp=StompConfig()))
     assert interface.messaging_template() is not None
+
+
+@pytest.mark.stomp
+def test_stomp_connection_failure_with_invalid_config(caplog):
+    caplog.at_level(logging.exception)
+    interface.set_config(ApplicationConfig(stomp=StompConfig(host="Not Exists")))
```
stan-dot, line 282: please do not do 3 nested inits in one line

stan-dot (dismissed):
small readability and naming suggestions

src/blueapi/client/event_bus.py
DiamondJoseph, line 37:
```suggestion
                MessageTopic(name="public.worker.event"),
```

These were Topics (multiple consumers) before, not Queues (single consumers)

src/blueapi/service/interface.py
```diff
@@ -51,10 +52,14 @@ def worker() -> TaskWorker:
...
+            broker=Broker(
+                host=stomp_config.host, port=stomp_config.port, auth=stomp_config.auth
+            )
+        )
 
         task_worker = worker()
-        event_topic = template.destinations.topic("public.worker.event")
+        event_topic = MessageQueue(name="public.worker.event")
```
DiamondJoseph, line 62:
```suggestion
        event_topic = MessageTopic(name="public.worker.event")
```

DiamondJoseph (changes requested):

pyproject.toml
ZohebShaikh: I think we should remove the fastapi[all] as it is causing problems on main as well...I have created a PR to address that issue https://github.com/DiamondLightSource/blueapi/pull/607

tests/core/fake_device_module.py
DiamondJoseph: This is unrelated to this change and will go away with a rebase on main. It's a result of [this change](https://github.com/DiamondLightSource/blueapi/pull/598) that made the handling of type conversion must more capable (by using the Pydantic handling rather than rolling our own).

ZohebShaikh:
@callumforrester  Can you have a look at this PR, If everything looks good we can create a release of bluesky-stomp

src/blueapi/service/interface.py
```diff
@@ -63,8 +68,12 @@ def messaging_template() -> MessagingTemplate | None:
...
             }
         )
-        template.connect()
-        return template
+        try:
+            template.connect()
+            return template
+        except Exception as ex:
+            logging.exception(msg="Failed to connect to message bus", exc_info=ex)
+            return None
```
DiamondJoseph, line 76: This except block changes the behaviour when a message bus is configured, but the message bus configuration is not valid.

Previously this was an error state that resulted in the failed startup of the service, which means on Kubernetes it would fall over, hopefully the message bus would recover, then this service could try and connect again. Now it will happily not create a connection in this case, which means on Kubernetes we will have created a declarative configuration but are not able to deliver that valid configuration.  I think this should be reverted and allowed to throw the exception here.

src/blueapi/service/main.py
```diff
@@ -312,6 +312,8 @@ def set_state(
         and new_state in _ALLOWED_TRANSITIONS[current_state]
     ):
         if new_state == WorkerState.PAUSED:
+            if state_change_request.defer is None:
+                state_change_request.defer = False
```
DiamondJoseph, line 316: Or can make StateChangeRequest.defer `bool` not `bool | None`- it's only going to be `None` if it's explicitly set to be but I don't see the use case for doing so?

DiamondJoseph (changes requested):

tests/service/test_interface.py
ZohebShaikh: ruff prefers it on one line

tests/conftest.py
```diff
@@ -6,27 +6,6 @@
...
-        help="skip stomp tests (e.g. because a server is unavailable)",
-    )
-
-
-def pytest_configure(config):
-    config.addinivalue_line("markers", "stomp: mark test as requiring stomp broker")
-
-
-def pytest_collection_modifyitems(config, items):
-    if config.getoption("--skip-stomp"):
```
callumforrester, line 23: Now that we've removed this, we no longer need any special handling for STOMP in CI, can you remove the parts of the github actions that start up activemq and rabbitmq?

ZohebShaikh: There are 2 tests that are used to connect to the activemq/rabbitmq...I could delete them or skip both I don't think are suitable ..

callumforrester:
@ZohebShaikh Would you be able to write some tests that cover the affected area? See https://app.codecov.io/gh/DiamondLightSource/blueapi/pull/589?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=DiamondLightSource

.github/workflows/_test.yml
```diff
@@ -24,12 +24,6 @@ jobs:
...
     runs-on: ${{ inputs.runs-on }}
 
-    services:
-      activemq:
-        image: rmohr/activemq:5.14.5-alpine
-        ports:
-          - 61613:61613
-
     steps:
       - name: Start RabbitMQ
```
callumforrester, line 28: Can you also remove this? I'd like to be sure it isn't needed before merging

src/blueapi/service/interface.py
```diff
@@ -48,13 +49,17 @@ def worker() -> TaskWorker:
 
 
 @lru_cache
-def messaging_template() -> MessagingTemplate | None:
+def messaging_template() -> MessagingTemplate:
```
DiamondJoseph, line 52:
```suggestion
def messaging_template() -> MessagingTemplate | None:
```

This still returns None if stomp_config is None?

docs/reference/openapi.yaml
```diff
@@ -1,481 +1,289 @@
...
-          anyOf:
-          - type: string
-          - type: 'null'
-          description: The ID of the current task, None if the worker is idle
-          title: Task Id
-      required:
-      - task_id
-      title: WorkerTask
-      type: object
+openapi: 3.1.0
```
DiamondJoseph, line 1: These changes seem unrelated, if possible I'd like the openapi.yaml to remain unchanged, or at least reduce the size of the changes to this file. Do we know why it needs to be regenerated? If it isn't are there changes?

DiamondJoseph (changes requested):

DiamondJoseph (approved):
LGTM!

PR #611, opened 2024-09-02
Add system tests

ZohebShaikh:


ZohebShaikh:
I have not move the current test to unit test will move after #589 is merge ...

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+    WorkerTask,
+)
+from blueapi.worker.event import TaskStatusEnum, WorkerState
+from blueapi.worker.task import Task
+from blueapi.worker.task_worker import TrackableTask
+
+_SIMPLE_TASK = Task(name="sleep", params={"time": 0.0})
+_LONG_TASK = Task(name="sleep", params={"time": 1.0})
+
+rest = BlueapiRestClient()
```
callumforrester, line 25: Should: Can you use `BlueapiClient` instead? Most of the calls here would just wrap `BlueapiRestClient` anyway but it makes for a good end-to-end test.

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+    url: str, BaseModel: Any, status_code: int = status.HTTP_200_OK
+) -> Any:
+    get_response = requests.get(url)
+    model = TypeAdapter(BaseModel).validate_python(get_response.json())
+    assert get_response.status_code == status_code
+    return model
+
+
+def test_get_plans():
+    assert rest.get_plans() == PlanResponse(plans=get_plans())
```
callumforrester, line 38: Should: While this test is useful it is ultimately calling `get_plans()` on both sides of the `==`, so in some cases it boils down to `assert with_extra_steps(get_plans()) == get_plans()`. The test will work fine if there is a problem in the extra steps, but not in every case.

For example, say there is a bug which makes `get_plans()` return `[]`, this test would still pass.

I think the only good solution is to write the list of expected plans yourself, in this module. It will make the test fail if anyone changes the default set but I _think_ that's what we want, then we are quickly made aware of any changes. I welcome opinions/comments, tagging @DiamondJoseph too.

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def test_get_plans_by_name():
+    for plan in get_plans():
+        assert rest.get_plan(plan.name) == plan
```
callumforrester, line 43: Should: See above for using `get_plans()`, if it were `[]` this test would erroneously pass.

My suggestion here would be to have a parametrised test based on a list of manually curated values:

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def test_get_non_existant_plan():
+    with pytest.raises(KeyError) as exception:
+        rest.get_plan("Not exists")
+    assert exception.value.args[0] == ("{'detail': 'Item not found'}")
+
+
+def test_get_devices():
+    assert rest.get_devices() == DeviceResponse(devices=get_devices())
```
callumforrester, line 53: Should: Same as `get_plans()`

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def test_get_device_by_name():
+    for device in get_devices():
+        assert rest.get_device(device.name) == device
```
callumforrester, line 58: Should: Same as `test_get_plans_by_name()`

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+def test_get_non_existant_device():
+    with pytest.raises(KeyError) as exception:
+        assert rest.get_device("Not exists")
+    assert exception.value.args[0] == ("{'detail': 'Item not found'}")
+
+
+def test_create_task_and_delete_task_by_id():
```
callumforrester, line 67: Could: This is good but could be split into two tests, `create_task` and `create_and_delete_task`. That way, if there is a problem, it's clearer where it is.

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+    create_task = rest.create_task(_SIMPLE_TASK)
+    rest.clear_task(create_task.task_id)
+
+
+def test_create_task_validation_error():
```
callumforrester, line 72: Should: Should also have tests for getting and deleting nonexistant tasks

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+    rest.clear_task(task_id=task_2.task_id)
+
+
+def test_get_current_state_of_environment():
+    assert rest.get_environment() == EnvironmentResponse(initialized=True)
+
+
+def test_delete_current_environment():
+    rest.delete_environment()
+    time.sleep(5)
```
callumforrester, line 189: Should: If you use `BlueapiClient` as suggested above it will poll for you and give a sensible error message.

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+        task_by_pending_request.json()
+    )
+
+    assert len(task_by_pending.tasks) == 2
+    for task in task_by_pending.tasks:
+        trackable_task = TypeAdapter(TrackableTask).validate_python(task)
+        assert trackable_task.is_complete is False and trackable_task.is_pending is True
+
+    rest.update_worker_task(WorkerTask(task_id=task_1.task_id))
+    time.sleep(0.1)
```
callumforrester, line 165: Should: You should really check that the plan is complete rather than just wait 0.1 seconds, that could lead to race conditions and make this test flaky. Again, if you use `BlueapiClient` it should handle this for you.

tests/system_tests/test_api_endpoints.py
```diff
@@ -0,0 +1,190 @@
...
+    assert exception.value.args[0] == ("{'detail': 'Item not found'}")
+
+
+def test_get_all_tasks():
+    created_tasks = []
+    for task in [_SIMPLE_TASK, _LONG_TASK]:
+        created_task = rest.create_task(task)
+        created_tasks.append(created_task)
+
+    task_list = get_response(rest._url("/tasks"), TasksListResponse)
```
callumforrester, line 84: I'm unsure why the client object doesn't provide what you want here

tests/system_tests/test_api_endpoints.py
callumforrester, line 1: This file is testing more than the API endpoints, it is testing the whole system, end-to-end. For now maybe just call it `test_blueapi_system.py`?

src/blueapi/service/interface.py
```diff
@@ -162,7 +160,7 @@ def get_worker_state() -> WorkerState:
     return worker().state
 
 
-def pause_worker(defer: bool | None) -> None:
+def pause_worker(defer: bool) -> None:
```
callumforrester, line 163: Could: Given the previous discussion it might be nice to quickly to this and the big schema change it necessitates in a separate PR

src/blueapi/service/interface.py
```diff
@@ -70,8 +70,6 @@ def messaging_template() -> MessagingTemplate | None:
         )
         template.connect()
         return template
-    else:
-        return None
```
callumforrester, line 74: Is there a reason for this change? It was put there in the first place because we decided running without a message bus was okay, especially for development purposes.

tests/test_cli.py
```diff
@@ -166,8 +166,8 @@ def test_valid_stomp_config_for_listener(
         input="\n",
     )
     assert (
-        result.output
-        == "Subscribing to all bluesky events from localhost:61613\nPress enter to exit"
+        "Subscribing to all bluesky events from localhost:61613\nPress enter to exit"
+        in result.output
```
callumforrester, line 170: Strange that you needed to do this, I can spot the change that caused it, any ideas?

callumforrester (changes requested):
Good start, @ZohebShaikh

DiamondJoseph:
Replaced by https://github.com/DiamondLightSource/blueapi/pull/618 for neatness

tests/test_cli.py
DiamondJoseph: I'm presuming there's an extra trailing \n somewhere?

tests/system_tests/test_api_endpoints.py
DiamondJoseph: I agree that that's what we want to be checking, `with_extra_steps(get_plans()) == <some hardcoded value>`, which we just acknowledge that this test (similarly to the Schema test) may be one that needs to be changed for seemingly unrelated changes, such as changing default loaded modules/changes in those modules.

tests/system_tests/test_api_endpoints.py
ZohebShaikh: To ensure a clean environment following the completion of a task, it is essential to revert any changes made during the process. For instance, if a test involves inserting data into a database, the appropriate approach would be to either remove the inserted data at the conclusion of the test or to instantiate a new database for each test. In this case, I have opted for the former approach, ensuring that the inserted data is deleted post-test to maintain a clean and consistent environment.

tests/system_tests/test_api_endpoints.py
ZohebShaikh: I have added code for this here https://github.com/DiamondLightSource/blueapi/pull/618/commits/b5bb82524a2a8b3e1852254f0ef6f2d52d85dcb3 

I think we should change 

```
get_task(task_id)         to   get_task_by_id(task_id) -> assert task_id is not ""

and get_all_tasks()  
```

ZohebShaikh: It was added by this https://github.com/DiamondLightSource/blueapi/pull/508

PR #612, opened 2024-09-03
Allow specifying initContainer resources separately to container resources

### LINKED ISSUE #602: Allow user to specify separate resource requirements for initContainer
The helm chart currently deploys an [initContainer](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) for setting up the scratch area and then a main container for running blueapi. Both use the `resources` section of `values.yaml` to determine how much resource (CPU and memory) to request from kubernetes. In practice, we have found that the requirements for the two containers may differ significantly, it seems to take quite a lot of CPU to set up the scratch area compared to running bluesky plans. The user should at least have the option of making the requests differ from eachother.

## Acceptance Critera
- User can specify `initContainerResources` or similar in `values.yaml`
- The initContainer uses this to determine its resource request, while the main container still uses the `resources` section.
- If `initContainerResources` is left blank, the initContainer defaults to the same resources as the main container
- If both are left blank, both containers do not explicitly request resources at all and accept the default that kubernetes gives them

### LINKED ISSUE #613: Rename `deployment.yaml` to `statefulset.yaml`
...in `helm/blueapi/templates`. We should follow helm conventions, the file now contains a `StatefulSet` rather than a `Deployment`

callumforrester:
Fixes #602 
Fixes #613


DiamondJoseph (approved):

DiamondJoseph:
This doesn't close #613

PR #618, opened 2024-09-05
535 system test

ZohebShaikh:


tests/system_tests/test_blueapi_system.py
```diff
@@ -0,0 +1,214 @@
...
+    )
+
+    assert len(task_by_pending.tasks) == 2
+    for task in task_by_pending.tasks:
+        trackable_task = TypeAdapter(TrackableTask).validate_python(task)
+        assert trackable_task.is_complete is False and trackable_task.is_pending is True
+
+    client.start_task(WorkerTask(task_id=task_1.task_id))
+    while not client.get_task(task_1.task_id).is_complete:
+        ...
```
callumforrester, line 189: Maybe put a `time.sleep(0.1)` here, just so it doesn't check 1000s of times per second, same for the one below.

src/blueapi/client/client.py
```diff
@@ -142,9 +143,19 @@ def get_task(self, task_id: str) -> TrackableTask[Task]:
         Returns:
             TrackableTask[Task]: Task details
         """
-
+        assert task_id != "", "Task ID cannot be empty"
         return self._rest.get_task(task_id)
 
+    def get_all_task(self) -> TasksListResponse:
```
callumforrester, line 149:
```suggestion
    def get_all_tasks(self) -> TasksListResponse:
```

callumforrester (changes requested):

src/blueapi/client/client.py
ZohebShaikh: Good eye! 👍🏼

src/blueapi/client/client.py
```diff
@@ -142,9 +143,19 @@ def get_task(self, task_id: str) -> TrackableTask[Task]:
         Returns:
             TrackableTask[Task]: Task details
         """
-
+        assert task_id != "", "Task ID cannot be empty"
```
DiamondJoseph, line 146:
```suggestion
        assert task_id, "Task ID not provided!"
```

Because an empty string is False and None is also false, this catches a couple of exception cases

DiamondJoseph (changes requested):

DiamondJoseph (approved):
otherwise LGTM

tests/unit_tests/test_cli.py
```diff
@@ -124,7 +124,7 @@ def test_submit_plan(runner: CliRunner):
         match=[matchers.json_params_matcher(body_data)],
     )
 
-    config_path = "tests/example_yaml/rest_config.yaml"
+    config_path = "tests/unit_tests/example_yaml/rest_config.yaml"
```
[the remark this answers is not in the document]
ZohebShaikh: I think we will need to look at it when we configure system test on CI

tests/system_tests/test_blueapi_system.py
```diff
@@ -0,0 +1,260 @@
...
+        task_status=TaskStatus(
+            task_id=task_id,
+            task_complete=True,
+            task_failed=False,
+        ),
+    )
+    assert running_event in all_events
+    assert pending_event in all_events
+    assert complete_event in all_events
+    assert len(all_events) == 3
```
DiamondJoseph, line 251:
```suggestion
    assert all_events == [
    WorkerEvent(
        state=WorkerState.RUNNING,
        task_status=TaskStatus(
            task_id=task_id,
            task_complete=False,
            task_failed=False,
        ),
    ),
    WorkerEvent(
        state=WorkerState.IDLE,
        task_status=TaskStatus(
            task_id=task_id,
            task_complete=False,
            task_failed=False,
        ),
    ),
    WorkerEvent(
        state=WorkerState.IDLE,
        task_status=TaskStatus(
            task_id=task_id,
            task_complete=True,
            task_failed=False,
        ),
    )]
```

I'd go further and explicitly check that they arrived in exactly this order

PR #631, opened 2024-09-11
separate the init config in helm

stan-dot:
long branch name due to @DiamondJoseph choice of words in the issue name, but at least it's entertaining

### DIFF
--- a/helm/blueapi/templates/statefulset.yaml
+++ b/helm/blueapi/templates/statefulset.yaml
@@ -64,7 +64,6 @@ spec:
             if [ $? -ne 0 ]; then echo 'Blueapi failed'; exit 1; fi;
             echo "Exporting venv as artefact"
             cp -r /venv/* /artefacts
-            
         volumeMounts:
           - name: init-config
             mountPath: "/config"
--- a/helm/blueapi/templates/statefulset.yaml
+++ b/helm/blueapi/templates/statefulset.yaml
@@ -60,7 +60,7 @@ spec:
         args:
           - |
             echo "Setting up scratch area"
-            blueapi -c /config/config.yaml setup-scratch
+            blueapi -c /config/initconfig.yaml setup-scratch
             if [ $? -ne 0 ]; then echo 'Blueapi failed'; exit 1; fi;
             echo "Exporting venv as artefact"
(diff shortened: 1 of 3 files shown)


helm/blueapi/values.yaml
```diff
@@ -108,12 +108,13 @@ worker:
...
-  #   repositories: []
-  #     - name: "dodal"
-  #       remote_url: https://github.com/DiamondLightSource/dodal.git
+
+initContainer:
+  scratch:
+   root: /blueapi-plugins/scratch
+    repositories: []
+      - name: "dodal"
+        remote_url: https://github.com/DiamondLightSource/dodal.git
```
DiamondJoseph, line 117: This yaml is invalid for both whitespace and setting a value twice. repositories should be `[]` with a commented out example for dodal. Does an auto format touch this file?

helm/blueapi/templates/statefulset.yaml
```diff
@@ -59,8 +64,9 @@ spec:
             if [ $? -ne 0 ]; then echo 'Blueapi failed'; exit 1; fi;
             echo "Exporting venv as artefact"
             cp -r /venv/* /artefacts
+
```
DiamondJoseph, line 67: remove whitespace

DiamondJoseph (changes requested):
Looks conceptually correct, some nits.

callumforrester (commented):
Concept looks good, will test it when the PR is marked as ready

helm/blueapi/templates/statefulset.yaml
stan-dot: done

callumforrester (approved):

DiamondJoseph (changes requested):
1 nit then LGTM

helm/blueapi/values.yaml
```diff
@@ -80,7 +80,8 @@ listener:
   resources: {}
 
 # Additional envVars to mount to the pod as a String
-extraEnvVars: []
+extraEnvVars:
+  []
```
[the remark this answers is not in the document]
stan-dot: my yaml parser corrects to the way it was,

Issue #601, opened 2024-08-21
Tests take too long


callumforrester:
Just the tests in `test_runner.py` take 30s locally. All the tests combined take nearly 90 seconds

## Acceptance Criteria
`tox -e tests` takes less than 30s on most machines

callumforrester:
Currently, the main culprits for slow tests are the runner tests (which spawn a subprocess) and the tracing tests (which perform some time-consuming opentelemetry setup). I suggest we mock this functionality in the unit tests and leave the system tests to do it for real.

callumforrester:
The runner tests are now the main culprit. This is because they create and destroy a subprocess for each test.

Three options discussed:
- One long-running fixture for the subprocess. This holds state (RE, plans, devices et al.) so it may cause tests to bleed into each other
- Don't make a subprocess at all, make the `Runner` run everything in-process. This would be faster but may miss critical errors (see #503)
- Mark those tests as skipped locally and only run them in CI. This improves the local DX but still makes CI take a long time, developer may also be frustrated if they cause problems that they don't see until after they push

callumforrester:
Tagging @keithralphs @DiamondJoseph @ZohebShaikh from offline discussion

PR #632, opened 2024-09-11
619 move to pyright

stan-dot:
points that need addressing

- [x] sim_motor types issues - might ignore but alternatively just delete
- [x] task worker 
    - [x] state is also problematic
    - [x]  'run', the part calling the RunEngine with the hooks is unclear


stan-dot:
@callumforrester please let me know your thoughts

callumforrester:
@stan-dot I like where this is going, thanks

stan-dot:
ok will try to finish this today

stan-dot:
simmotor has `float` and has that in the tests as well - but the Ophyd type is this: `    value : object, optional        The initial value. Default is 0.` which pyright reads as int. 

adding an ignore line pending ophyd choosing a side here

stan-dot:
`task_worker` complains - because run_engine is not typed :cry:

stan-dot:
`DeviceStatus` in devices.py complains that the DeviceStatus (imported from ophyd :exclamation: ) does not support the bluesky `Status` protocol

stan-dot:
still one issue with the stomp instantiation from the CLI - what's your opinion @callumforrester ?

src/blueapi/service/interface.py
```diff
@@ -49,12 +49,14 @@ def worker() -> TaskWorker:
 
 
 @lru_cache
-def messaging_template() -> StompClient | None:
-    stomp_config = config().stomp
+def cached_stomp_client() -> StompClient | None:
```
callumforrester, line 52: Should: Why rename? And why rename only this function?

This seems like an unrelated change to the issue and should be reverted.

src/blueapi/service/runner.py
```diff
@@ -141,9 +141,9 @@ def _rpc(
     **kwargs: Any,
 ) -> T:
     mod = import_module(module_name)
-    func: Callable[P, T] = _validate_function(
+    func: Callable[..., T] = _validate_function(
         mod.__dict__.get(function_name, None), function_name
-    )
+    )  # type: ignore
```
callumforrester, line 146: Should: Does marking `args` and `kwargs` as `P` solve this problem?

See paramspec docs: https://docs.python.org/3/library/typing.html

src/blueapi/startup/simmotor.py
```diff
@@ -90,5 +90,5 @@ def __init__(self, *, timeout: float, **kwargs) -> None:
         super().__init__(**kwargs)
         self._timeout = timeout
 
-    def set(self, value: float) -> Status:
-        return Status(timeout=self._timeout)
+    def set(self, value: float) -> MoveStatus:
```
callumforrester, line 93: Should: No need to narrow the return type

```suggestion
    def set(self, value: float) -> Status:
```

src/blueapi/utils/__init__.py
```diff
@@ -7,7 +7,6 @@
 __all__ = [
     "handle_all_exceptions",
     "load_module_all",
-    "ConfigLoader",
```
callumforrester, line 10: Should: Is there any typing-related reason for this change. If not it is unrelated and should be reverted.

src/blueapi/worker/task_worker.py
```diff
@@ -95,7 +95,7 @@ def __init__(
 
         self._tasks = {}
 
-        self._state = WorkerState.from_bluesky_state(ctx.run_engine.state)
+        self._state = WorkerState.from_bluesky_state(ctx.run_engine.state)  # type: ignore
```
callumforrester, line 98: This seems fixable by setting the return type of `from_bluesky_state`, possibly including a `cast`?

src/blueapi/worker/task_worker.py
```diff
@@ -181,8 +182,8 @@ def mark_task_as_started(event: WorkerEvent, _: str | None) -> None:
                 task_started.set()
 
         LOGGER.info(f"Submitting: {trackable_task}")
+        sub = self.worker_events.subscribe(mark_task_as_started)
```
callumforrester, line 185: Should: Moving this out of the `try` block means a network error could potentially bring down the worker subprocess

tests/unit_tests/worker/devices.py
```diff
@@ -41,7 +41,7 @@ def _run_callbacks(self) -> None:
 class AdditionalStatusDevice(Device, Movable):
     def set(self, value: float) -> Status:
         status = AdditionalUpdateStatus(self)
-        return status
+        return status  # type: ignore
```
callumforrester, line 44: Should: What prompted this? It seems fixable

callumforrester (changes requested):

callumforrester:
@stan-dot: Maybe we should make the password a secret in a separate PR, in order to deal with the errors. It is not essential to adopting pyright.

src/blueapi/startup/simmotor.py
stan-dot: Pyright did complain, and it's a specific implementation. why no need?

src/blueapi/utils/__init__.py
stan-dot: Pyright was complaining. Pyright complains about more things than just typing

src/blueapi/worker/task_worker.py
stan-dot: will look into it

src/blueapi/startup/simmotor.py
callumforrester: You shouldn't narrow the return type if the caller doesn't need it, you should keep the type general to preserve abstraction

stan-dot:
I'd agree with you, @callumforrester  had this difference was between a 200 lines PR and 1k lines PR


>  Maybe we should make the password a secret in a separate PR


this soft *maybe* does not feel honest as we both know you'd *abosolutely not allow* for this to go in the same PR, so please skip the pleasantries and just tell me to raise an issue and make a PR...

src/blueapi/utils/__init__.py
DiamondJoseph: ConfigLoader doesn't appear to be imported, so I expect it was because it's not in the namespace.

callumforrester:
For the record it was a genuine _maybe_, I agree that those changes are tangentially related. I read them during review and decided not to contest them unless they caused problems. I am unable to commit the time to help you debug the error right now so I'm leaving it up to you whether you want to put time into debugging on your own or make an issue and defer until later. I am happy with either resolution.

src/blueapi/utils/__init__.py
stan-dot: yes, pyright complained about the import wrong, so deleting the import meant that the export also needs to be deleted

src/blueapi/startup/simmotor.py
stan-dot: for whom? developer or the machine? IDE?

MoveStatus is the thing that the SimMotor returns. it gives me more information when I read a method to read that it's more narrow. 


you haven't provided a _source_ for your claim , but I can for mine - covariant return type

https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
https://en.wikipedia.org/wiki/Robustness_principle

callumforrester: I do not believe that the robustness principle is the only principle to apply here, however this is quite a minor change and I don't hugely object to it going in. 

I think there is a question of policy for the codebase that we should address, probably outside this PR. I've just been down an interesting rabbit hole. I think this post sums up my point of view nicely: https://www.hacklewayne.com/postel-s-law-is-simple-to-fulfil-the-potential-of-your-code

In summary:
- The robustness principle (or Postel's law) is a tool that is used to maximise the usefulness of a given function to any potential caller
- Specifically you use it to vary the input and return types of the function, making the input as generic as possible and the output as specific as possible. The thing you do not vary is the implementation, since the input/return types are varied relative to that. The implementation performs a set of operations on the input type and provides a return type on which the caller can perform a different set of operations, therefore the variance is limited to input types that have the operations required for the implementation (but nothing else) and return types that have all the possible operations that the implementation can possibly provide.
- Usefulness is not the only metric to optimise for, however. We may also want to optimise for maintainability. By naively applying Postel's law you create functions that are as generic as possible in what they accept and as specific as possible in what they return relative to their implementation, this will make it harder to change the implementation in the future because callers may now depend on some of the large number of operations made available by the maximally specific return type. This is why it is better to have a return type with fewer possible operations if you are optimising for maintainability, you only provide the caller with what they really need. 

So this discussion is resolved by deciding how much we prioritise long-term maintainability, and therefore how much we do/don't apply Postel's law. In general I think maintainability is a very high priority for blueapi in its aim of avoiding the same problems that GDA has. I would rather effort were put in now to avoid pain later. As I say I think this discussion is out of scope for this PR so I'm happy to let this in, but for large-scale changes I will continue to advocate for the maintainability route. I'm also quite happy to continue this in an issue or other appropriate forum.

stan-dot: > this will make it harder to change the implementation in the future because callers may now depend on some of the large number of operations made available by the maximally specific return type. This is why it is better to have a return type with fewer possible operations if you are optimising for maintainability, you only provide the caller with what they really need.


this is not a library though. this is an application. external callers are concerned with the REST api, not the internal code

Regarding this specific case this uses Ophyd so should be removed anyway. And a stronger argument is that ophyd-async in dodal and elsewhere just has generic `AsyncStatus`, not a specific `MoveStatus`.


> In general I think maintainability is a very high priority for blueapi in its aim of avoiding the same problems that GDA has. I would rather effort were put in now to avoid pain later

the goal of blueapi is to deliver business value in less than 5 years time. maintainability in 10 years time vs adaptability now is an opinionated tradeoff. Also this specific 'covariant return' aspect, both cases (for and against) can be construed as pursuing the *value* of maintainability, yet differ in the *strategy* chosen for it. 

_Trauma_ from past (software) experience can make (software) people overcorrect and act neurotically.


Specifically for GDA the three biggest pain points (afaik) - multiple repos to make one feature change, thick client, and chaotic display mechanisms - are all handled *for free* by the infrastructure and architecture (microservices and a separate web GUI). 

Not to mention improving organizational information flow practices. 

With all of this, marginal effectiveness of any more 'putting effort now' decreases.

> . By naively applying Postel's law you create functions that are as generic as possible in what they accept and as specific as possible in what they return relative to their implementation, this will make it harder to change the implementation in the future because callers may now depend on some of the large number of operations made available by the maximally specific return type.

that is ad-absurdum, not what I said.

finally, if the function does not say that a Movable returns a MoveStatus, it would be on a client to double check if the returned value is actually from a MoveStatus or just Status - the logic is just moved elsewhere.

callumforrester: Yes, it is an opinionated tradeoff, which is why I wanted a discussion. For this specific case I think we agree re: ophyd-async and `AsyncStatus`, so can close this thread.

stan-dot: agreed

src/blueapi/service/interface.py
stan-dot: it must have been left out when the main StompClient got changed

src/blueapi/worker/task_worker.py
stan-dot: from the first look on it, to add a callback on a potential event is not a network event. 

it is only the following line:

```python
            self._task_channel.put_nowait(trackable_task)
            
```

that does something that looks like network logic

src/blueapi/service/runner.py
stan-dot: solved, thanks for the tip!

stan-dot:
```/workspaces/blueapi/src/blueapi/startup/simmotor.py
  /workspaces/blueapi/src/blueapi/startup/simmotor.py:41:9 - error: Method "set" overrides class "SynAxis" in an incompatible manner
    Return type mismatch: base method returns type "MoveStatus", override returns type "Status"
      "Status" is not assignable to "MoveStatus" (reportIncompatibleMethodOverride)
  /workspaces/blueapi/src/blueapi/startup/simmotor.py:70:14 - error: "MoveStatus" is not defined (reportUndefinedVariable)
  /workspaces/blueapi/src/blueapi/startup/simmotor.py:93:36 - error: "MoveStatus" is not defined (reportUndefinedVariable)
  /workspaces/blueapi/src/blueapi/startup/simmotor.py:94:16 - error: "MoveStatus" is not defined (reportUndefinedVariable)
/workspaces/blueapi/src/blueapi/worker/task_worker.py
  /workspaces/blueapi/src/blueapi/worker/task_worker.py:99:54 - error: Argument of type "LoggingPropertyMachine | ProxyString | str" cannot be assigned to parameter "bluesky_state" of type "RawRunEngineState" in function "from_bluesky_state"
    Type "LoggingPropertyMachine | ProxyString | str" is not assignable to type "RawRunEngineState"
      Type "str" is not assignable to type "RawRunEngineState"
        Type "str" is not assignable to type "type[ProxyString]"
        Type "str" is not assignable to type "type[str]"
        Type "str" is not assignable to type "type[PropertyMachine]" (reportArgumentType)
/workspaces/blueapi/tests/unit_tests/worker/devices.py
  /workspaces/blueapi/tests/unit_tests/worker/devices.py:43:9 - error: Method "set" overrides class "Movable" in an incompatible manner
    Return type mismatch: base method returns type "Status", override returns type "Status"
      "Status" is incompatible with protocol "Status"
        "exception" is an incompatible type
          Type "(timeout: Unknown | None = None) -> (StatusTimeoutError | Exception | type[Exception] | None)" is not assignable to type "(timeout: float | None = 0) -> (BaseException | None)"
            Function return type "StatusTimeoutError | Exception | type[Exception] | None" is incompatible with type "BaseException | None"
        "bluesky.protocols.Status" is not assignable to "ophyd.status.Status"
        "bluesky.protocols.Status" is not assignable to "ophyd.status.Status" (reportIncompatibleMethodOverride)
  /workspaces/blueapi/tests/unit_tests/worker/devices.py:45:16 - error: Type "AdditionalUpdateStatus" is not assignable to return type "Status"
    "AdditionalUpdateStatus" is not assignable to "Status" (reportReturnType)
7 errors, 0 warnings, 0 informations 

```

those are the errors, they are all grounded in the chaos around ophyd-vs-ophyd-async bluesky protocol import (SynAxis and worker devices cases), as well as with State tracking - EngineState and the application specific state, and the unexpected super_state_machine import - not maintained, afaik vendored in into Bluesky. 

either those are solved in a different PR or this PR is merged with 'ignore' flags.

src/blueapi/service/runner.py
```diff
@@ -138,13 +138,13 @@ def _rpc(
...
-    **kwargs: Any,
+    **P: Any,
 ) -> T:
     mod = import_module(module_name)
-    func: Callable[P, T] = _validate_function(
+    func: Callable[..., T] = _validate_function(
         mod.__dict__.get(function_name, None), function_name
     )
-    value = func(*args, **kwargs)
+    value = func(*args, **P)
```
DiamondJoseph, line 147: This does not look right. P is a ParamSpec, here you're not just naming the dictionary constructed from un-splatting any kwargs as `P`.

src/blueapi/worker/task_worker.py
callumforrester: That's not network logic, that's adding to an internal queue. However, it is not just network logic you need to worry about: Anything that can throw an exception should go inside the `try` block, since we don't want the subprocess to crash and leave the main process spinning.

callumforrester:
@stan-dot I am okay with ignore flags on ophyd (not ophyd-async) devices, this problem is probably related to https://github.com/bluesky/bluesky/issues/1809

src/blueapi/service/runner.py
stan-dot: nevermind, it's ok now

src/blueapi/service/interface.py
stan-dot: PR here

and removed this change from this pyright PR
https://github.com/DiamondLightSource/blueapi/pull/673

tests/unit_tests/worker/devices.py
stan-dot: the bluesky error is the reason - closing the conversation as discussed

stan-dot:
any idea why are the tests taking forever @callumforrester ?

callumforrester:
No although all bets are off while `main` is broken...

src/blueapi/worker/task_worker.py
stan-dot: > from the first look on it, to add a callback on a potential event is not a network event.


as I expected. if the first look was wrong, it means that the variable and file naming is not descriptive enough...

stan-dot: why adding to an internal queue would fail?

stan-dot:
will just make new issues for the type: ignores. also a propagation of changes to upstream repos to increase type safety there

src/blueapi/worker/task_worker.py
callumforrester: `put_nowait` fails if the queue is full. `put` does not fail but blocks until the queue is no longer full.

https://docs.python.org/3/library/asyncio-queue.html

stan-dot: why would the queue be full?

callumforrester: [It has a max capacity of 1 item at a time](https://github.com/DiamondLightSource/blueapi/blob/c9c2d449463cc0810452cf24b53f267ed437f1ba/src/blueapi/worker/task_worker.py#L103)

stan-dot: that's a material reason for it, what is the teleological reason?

callumforrester: The queue is actually a channel between 2 threads, most programming languages have a separate abstraction for this but Python reuses `Queue`. So it's not being used as a queue really, but more as a box in which to put the "current" task and nothing else. So the calling thread can only ever submit one plan at a time.

stan-dot: I added a comment explaining that part.

Still I think the discussion got sidetracked a bit as
>         sub = self.worker_events.subscribe(mark_task_as_started)
is not the `put_nowait` line

callumforrester: You're right, I don't think that line could cause a network error, but it could still raise an exception, so why was it moved out of the `try` block?

stan-dot: because it's just adding a callback

stan-dot:
locally tests also take a long time, looking for which one makes it so long

stan-dot:
also issues on main

src/blueapi/worker/task_worker.py
callumforrester: So if anything in `_cycle` raises an uncaught error it will kill the worker thread and blueapi will need a manual restart, which is why I prefer to play it safe and just catch and log _any_ exception in the cycle loop. That code won't currently raise an exception, but it could be modified in the future such that it does, and the person modifying it won't necessarily know about this line. Please put it back.

callumforrester:
@stan-dot is #601 related to the test issues?

src/blueapi/worker/task_worker.py
stan-dot: ok, I'll include information from your comment for clarity

stan-dot:
not sure

stan-dot:
```python
found 0 vulnerabilities
/workspaces/blueapi/tests/conftest.py
  /workspaces/blueapi/tests/conftest.py:6:43 - error: "TransitionError" is not exported from module "bluesky.run_engine" (reportPrivateImportUsage)
/workspaces/blueapi/tests/unit_tests/test_config.py
  /workspaces/blueapi/tests/unit_tests/test_config.py:153:9 - error: Argument missing for parameter "password" (reportCallIssue)
  /workspaces/blueapi/tests/unit_tests/test_config.py:153:48 - error: No parameter named "passcode" (reportCallIssue)
/workspaces/blueapi/tests/unit_tests/client/test_client.py
  /workspaces/blueapi/tests/unit_tests/client/test_client.py:7:5 - error: "JsonObjectSpanExporter" is unknown import symbol (reportAttributeAccessIssue)
  /workspaces/blueapi/tests/unit_tests/client/test_client.py:8:5 - error: "asserting_span_exporter" is unknown import symbol (reportAttributeAccessIssue)
/workspaces/blueapi/tests/unit_tests/service/test_runner.py
  /workspaces/blueapi/tests/unit_tests/service/test_runner.py:7:5 - error: "JsonObjectSpanExporter" is unknown import symbol (reportAttributeAccessIssue)
  /workspaces/blueapi/tests/unit_tests/service/test_runner.py:8:5 - error: "asserting_span_exporter" is unknown import symbol (reportAttributeAccessIssue)
/workspaces/blueapi/tests/unit_tests/worker/test_task_worker.py
  /workspaces/blueapi/tests/unit_tests/worker/test_task_worker.py:11:5 - error: "JsonObjectSpanExporter" is unknown import symbol (reportAttributeAccessIssue)
  /workspaces/blueapi/tests/unit_tests/worker/test_task_worker.py:12:5 - error: "asserting_span_exporter" is unknown import symbol (reportAttributeAccessIssue)
9 errors, 0 warnings, 0 informations 

```
fixing those import errors and all should be fine

src/blueapi/worker/task_worker.py
stan-dot: outdated

stan-dot:
@keithralphs the error here seems to stem from how the observability_utils works

```python

    next(self.gen)
  File "/venv/lib/python3.11/site-packages/observability_utils/tracing/asserting_exporter.py", line 54, in asserting_span_exporter
    if exporter.top_span is not None:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'TracerProvider' object has no attribute 'top_span'


```

`observability_utils/tracing/asserting_exporter.py` line 41...
```python


@contextmanager
def asserting_span_exporter(
    exporter: JsonObjectSpanExporter, func_name: str, *span_args: str
):
    """Use as a with block around the function under test if decorated with
    start_as_current_span to check span creation and content.

    params:
        func_name: The name of the function being tested
        span_args: The arguments specified in its start_as_current_span decorator
        or added using Add_span_attributes.
    """
    yield
    if exporter.top_span is not None:
        span = exporter.top_span.result(timeout=5.0)
        assert span.name == func_name
        for param in span_args:
            assert param in span.attributes


```

I assume if line `exporter.top_span` was with a 'get' syntax, we'd avoid this error

stan-dot:
@DiamondJoseph I just rebased and again getting the error:

```python

The conflict is caused by:
    blueapi 0.1.dev1+gc4168c7 depends on bluesky==1.13.0a3
    The user requested (constraint) bluesky==1.13
    
    ```
 imo the main needs to be updated to request a specific bluesky version

**Comment by DiamondJoseph** [comment]
@stan-dot when you rebased you've overwritten the new bluesky limits on main (>=1.13 and 1.13>1.13aN, so your limit now is incompatible).

File: docs/explanations/lifecycle.md [hunk 9]
```diff
@@ -6,7 +6,7 @@ of being written, loaded and run. Take the following plan.
     from typing import Any, List, Mapping, Optional, Union
     
     import bluesky.plans as bp
-    from bluesky.utils import MsgGenerator
+    from blueapi.core import MsgGenerator
```
DiamondJoseph, line 9:
```suggestion
    from bluesky.utils import MsgGenerator
```

DiamondJoseph: revert

pyproject.toml
```diff
@@ -12,7 +12,7 @@ classifiers = [
 ]
 description = "Lightweight bluesky-as-a-service wrapper application. Also usable as a library."
 dependencies = [
-    "bluesky>=1.13",
+    "bluesky==1.13.0a3",
```
DiamondJoseph, line 15:
```suggestion
    "bluesky>=1.13",
```

DiamondJoseph: revert

src/blueapi/cli/cli.py
```diff
@@ -163,19 +166,16 @@ def get_devices(obj: dict) -> None:
...
     config: ApplicationConfig = obj["config"]
-    if config.stomp is not None:
-        event_bus_client = EventBusClient(
-            StompClient.for_broker(
-                broker=Broker(
-                    host=config.stomp.host,
-                    port=config.stomp.port,
-                    auth=config.stomp.auth,
-                )
+    assert config.stomp is not None, "By this point, the stomp config should be set"
```
DiamondJoseph, line 169:
```suggestion
    assert config.stomp is not None, "Message bus needs to be configured"
```

nit: keep the exception message the same

src/blueapi/client/client.py
```diff
@@ -46,17 +46,15 @@ def __init__(
...
     def from_config(cls, config: ApplicationConfig) -> "BlueapiClient":
         rest = BlueapiRestClient(config.api)
-        if config.stomp is not None:
-            stomp_client = StompClient.for_broker(
-                broker=Broker(
-                    host=config.stomp.host,
-                    port=config.stomp.port,
-                    auth=config.stomp.auth,
-                )
+        assert config.stomp is not None, "By this point, the stomp config should be set"
```
DiamondJoseph, line 49: This prevents starting a client without a message bus- we can do non-plan running operations without a message bus, is this on purpose?

src/blueapi/core/bluesky_types.py
```diff
@@ -24,7 +24,7 @@
     Triggerable,
     WritesExternalAssets,
 )
-from bluesky.utils import MsgGenerator
+from dls_bluesky_core.core import MsgGenerator
```
DiamondJoseph, line 27:
```suggestion
from bluesky.utils import MsgGenerator
```

DiamondJoseph: revert

src/blueapi/startup/example_plans.py
```diff
@@ -1,15 +1,17 @@
+import bluesky.plan_stubs as bps
+from bluesky.plans import count
 from bluesky.protocols import Movable, Readable
+from dls_bluesky_core.core import MsgGenerator
 from dodal.common import inject
-from dodal.plan_stubs.wrapped import move
-from dodal.plans import count
 
-from blueapi.core import MsgGenerator
```
DiamondJoseph, line 6: revert these

src/blueapi/startup/example_plans.py
```diff
@@ -26,5 +28,5 @@ def stp_snapshot(
     Yields:
         Iterator[MsgGenerator]: Bluesky messages
     """
-    yield from move({temperature: 0, pressure: 10**5})
+    yield from bps.mv({temperature: 0, pressure: 10**5})  # type: ignore
```
DiamondJoseph, line 31:
```suggestion
    yield from move({temperature: 0, pressure: 10**5})
```

DiamondJoseph: revert

src/blueapi/startup/simmotor.py
```diff
@@ -90,5 +90,5 @@ def __init__(self, *, timeout: float, **kwargs) -> None:
...
+        return MoveStatus(positioner=None, target=None, timeout=self._timeout)
```
DiamondJoseph, line 94:
```suggestion
        return MoveStatus(positioner=self, target=float, timeout=self._timeout)
```
We know these values

src/blueapi/startup/simmotor.py
```diff
@@ -27,8 +27,8 @@ def __init__(
         super().__init__(
             name=name,
             readback_func=readback_func,
-            value=value,
-            delay=delay,
+            value=value,  # type: ignore
+            delay=delay,  # type: ignore
```
DiamondJoseph, line 31: What's the correct type for these, can we make the SynAxisWithMotionEvents correct?

tests/unit_tests/cli/test_scratch.py
```diff
@@ -12,14 +13,14 @@
 
 
 @pytest.fixture
-def directory_path() -> Path:  # type: ignore
+def directory_path() -> Generator[Path]:
```
DiamondJoseph, line 16: I would remove the directory_path fixture entirely in favour of the pre-existing tmp_path

tests/unit_tests/core/test_context.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 from bluesky.protocols import Descriptor, Movable, Readable, Reading, SyncOrAsync
-from bluesky.utils import MsgGenerator
+from dls_bluesky_core.core import MsgGenerator
```
DiamondJoseph, line 8: revert

tests/unit_tests/service/test_interface.py
```diff
@@ -4,11 +4,11 @@
 
 import pytest
 from bluesky_stomp.messaging import StompClient
+from dls_bluesky_core.core import MsgGenerator
```
DiamondJoseph, line 7: revert

DiamondJoseph (changes requested):
Mostly looks good!

docs/explanations/lifecycle.md
stan-dot: ok but why? otherwise I got the error, I spent 30 minutes on this yesterday

src/blueapi/client/client.py
stan-dot: this is just to prevent a type error

src/blueapi/core/bluesky_types.py
stan-dot: I specifically added this yesterday, adjusting imports and versions 😠

tests/unit_tests/cli/test_scratch.py
stan-dot: is that in the scope for this PR?

src/blueapi/startup/simmotor.py
stan-dot: as discussed above with @callumforrester we do things the easy way with the ophyd-sync devices, as they aren't long for this codebase

tests/unit_tests/cli/test_scratch.py
DiamondJoseph: Ack, leave it

docs/explanations/lifecycle.md
DiamondJoseph: As discussed on slack, intervening change fixed compatibility to bluesky>=1.13

stan-dot:
outstanding errors:
```python
observability_utils/tracing/asserting_exporter.py", line 54, in asserting_span_exporter
    if exporter.top_span is not None:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'TracerProvider' object has no attribute 'top_span'

blueapi/blueapi/tests/conftest.py:6:43 - error: "TransitionError" is not exported from module "bluesky.run_engine"
    Import from "bluesky._vendor.super_state_machine.errors" instead (reportPrivateImportUsage)
1 error, 0 warnings, 0 informations 
ERROR: InvocationError for command /opt/hostedtoolcache/Python/3.11.10/x64/bin/pyright src tests (exited with code 1)
```

bluesky==0.13.0, observability utils==0.1.4

src/blueapi/core/bluesky_types.py
callumforrester: `dls-bluesky-core` is deprecated though

src/blueapi/client/client.py
callumforrester: Blueapi should be able to start up without a message bus, please handle the case where there is no stomp config (don't connect to stomp in this case)

docs/explanations/lifecycle.md
```diff
@@ -6,7 +6,7 @@ of being written, loaded and run. Take the following plan.
     from typing import Any, List, Mapping, Optional, Union
     
     import bluesky.plans as bp
-    from bluesky.utils import MsgGenerator
+    from blueapi.utils import MsgGenerator
```
DiamondJoseph, line 9:
```suggestion
    from bluesky.utils import MsgGenerator
```

src/blueapi/client/client.py
stan-dot: ok added that logic

DiamondJoseph (approved):
Just needs the tests fixed- I think the fixture is now returning the wrong object: it may have previously been returning the right object of the wrong type?

exporter in conftest

stan-dot:
```python

 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: assert list[bluesky.protocols.Movable] == list[blueapi.core.context.BlueskyContext._reference.<locals>.Reference]
  
  Full diff:
  - list[blueapi.core.context.BlueskyContext._reference.<locals>.Reference]
  + list[bluesky.protocols.Movable]


```

some tests still not updated wrt the dependencies change?

Issue #495, opened 2024-06-03
Set configuration options via environment variables


DominicOram:
As a user trying to use the client with stomp it is frustrating that I have to provide the config file every time I call the CLI. It would be better if I could specify this as an environment variable so that I do not need to keep specifying it. We could do either:
* `export BLUEAPI.config.file=config_file` - this is the same sort of file we would currently use
* `export BLUEAPI.config.api.host=my_host` - we can then split out all the config fields

stan-dot:
@DominicOram  what is your preference?

what if we have `~/.blueapi/config.yaml` as a default location? it feels unix-like

DominicOram:
No preference, as long as it's well documented I'm not too fussed

callumforrester:
@stan-dot the configuration file is not _always_ required per-se, but only when you want configuration that's different to the defaults. The intention was always to provide other ways of overriding configuration than a file, which is a bit clunky for a single value. That is what this ticket is for.

For that reason, I have a preference for the second of @DominicOram's options. It's relatively common to only want to override one value, for example `export BLUEAPI.config.stomp.host=localhost` would automatically turn on the message bus for local dev and make #615 easier. It seems like overkill to need a whole file for this. Let alone to need to know a default location where you need to put it to make the magic work. N.b. this is where I struggle with databroker sometimes (see [docs](https://blueskyproject.io/databroker/how-to/file-backed-catalog.html#persistent-catalog)), it's nice to be able to do something simple just from the terminal.

Eventually I would like to see blueapi use the same configuration override order that @tpoliaw has introduced to GDA:
1. Default values, unless overridden by...
2. File, unless overridden by...
3. Environment variables, unless overridden by...
4. Values passed in via a CLI argument

callumforrester:
So the long and short of it.

## Acceptance Criteria
1. If I run `export BLUEAPI.config.<some value>`, the default configuration for that value (found in `blueapi/src/config.py` is overridden by `<some value>`
2. If I run the above and the value is already specified in the configuration file, the environment variable still overrides
3. The docs explain this
4. Tests cover this

callumforrester:
This issue also relates to/could be done at the same time as #532

stan-dot:
> Eventually I would like to see blueapi use the same configuration override order that @tpoliaw has introduced to GDA:
>    Default values, unless overridden by...
>    File, unless overridden by...
>    Environment variables, unless overridden by...
>    Values passed in via a CLI argument

[...]

PR #649, opened 2024-09-25
add env loading

stan-dot:


stan-dot:
slightly unclear where should the default values come from

https://github.com/DiamondLightSource/blueapi/issues/495#issuecomment-2342763915

stan-dot:
this needs to be changed because at the moment the tests are failing because of the config loading error@

https://github.com/DiamondLightSource/blueapi/actions/runs/11071605310/job/30763855823?pr=649

stan-dot:
regarding the `run sleep plan` test, the client._config.port and host give localhost:8000, so the loading is not correct

stan-dot:
harder to test, as the debugger does not pick up the client changes. they do not propagate correctly to the client object from the CLI, though config is called

stan-dot:
OTOH the stomp-loading error is *bizarre*, and the `tkagg` is nowhere to be found

```
File "/workspaces/blueapi/tests/unit_tests/test_cli.py", line 170, in test_valid_stomp_config_for_listener
    assert result.output == dedent("""\
AssertionError: assert 'Backend tkagg is interactive backend. Turning interactive mode on.\n' == 'Subscribing to all bluesky events from localhost:61613\nPress enter to exit\n'
  
  + Backend tkagg is interactive backend. Turning interactive mode on.
  - Subscribing to all bluesky events from localhost:61613
  - Press enter to exit
```

stan-dot:
for the third of the failing tests, maybe the config is not loaded fully before the assertion is called?

let's check if the cli:main line for `use_values_from_yaml` is called when debugging the test

stan-dot:
it must be the cli OR env values that override (incorrectly) the prior `stomp` config

stan-dot:
not quite, it's not overriding, it's just that the `use_values` does not succeed in mutating the ConfigLoader state

stan-dot:
I forgot the assignment after making a pure function:
```python
        self._values = recursively_updated_map(self._values, values)

```

stan-dot:
ok actually the last test might be failing due to other kinds of config being invalid, not the `scratch` field:

```blueapi.utils.invalid_config_error.InvalidConfigError: Something is wrong with the configuration file:
Type: string_type, Location: stomp -> host, Message: Input should be a valid string, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/string_type
Type: int_type, Location: stomp -> port, Message: Input should be a valid integer, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/int_type
Type: enum, Location: env -> sources -> 0 -> kind, Message: Input should be 'planFunctions', 'deviceFunctions' or 'dodal', Input: None, Documentation: https://errors.pydantic.dev/2.9/v/enum
Type: path_type, Location: env -> sources -> 0 -> module -> lax-or-strict[lax=union[json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]],function-after[path_validator(), str]],strict=json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]]], Message: Input is not a valid path for <class 'pathlib.Path'>, Input: None
Type: string_type, Location: env -> sources -> 0 -> module -> str, Message: Input should be a valid string, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/string_type
Type: bool_type, Location: env -> events -> broadcast_status_events, Message: Input should be a valid boolean, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/bool_type
Type: literal_error, Location: logging -> level, Message: Input should be 'NOTSET', 'DEBUG', 'INFO', 'WARNING', 'ERROR' or 'CRITICAL', Input: None, Documentation: https://errors.pydantic.dev/2.9/v/literal_error
Type: string_type, Location: api -> host, Message: Input should be a valid string, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/string_type
Type: int_type, Location: api -> port, Message: Input should be a valid integer, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/int_type
Type: string_type, Location: api -> protocol, Message: Input should be a valid string, Input: None, Documentation: https://errors.pydantic.dev/2.9/v/string_type
```

stan-dot:
thus it comes back to the question of `where does the default config come from` - which is - the `config.py` hardcoded values, such as:

```python

class RestConfig(BlueapiBaseModel):
    host: str = "localhost"
    port: int = 8000
    protocol: str = "http"

```

stan-dot:
> > thus it comes back to the question of `where does the default config come from` - which is - the `config.py` hardcoded values, such as:
> > ```python
> > class RestConfig(BlueapiBaseModel):
> >     host: str = "localhost"
> >     port: int = 8000
> >     protocol: str = "http"
> > ```
> 
> Not sure, if this is the answer you looking for but these are default value inside uvicorn so if you leave it blank when you do run these will be the default value so I guess we just bring it up to so it changeable.

Cool, I didn't realize the uvicorn connection, thanks

stan-dot:
yay it works!

src/blueapi/cli/cli.py
```diff
@@ -37,11 +36,21 @@
...
 @click.pass_context
-def main(ctx: click.Context, config: Path | None | tuple[Path, ...]) -> None:
-    # if no command is supplied, run with the options passed
+def main(
+    ctx: click.Context, config: Path | None | tuple[Path, ...], env_prefix: str
+) -> None:
+    """
+    Main entry point for the CLI,
+    loading configuration in the correct order of precedence:
+    1. Defaults -> 2. File -> 3. Environment variables -> 4. CLI arguments.
```
callumforrester, line 47: Nit: This will be displayed to the user when they type `blueapi --help` so maybe make it a bit more user-facing

```suggestion
    Blueapi CLI, see below for options.
    The configuration is loaded in the following order:
    1. Defaults
    2. File (via -c option)
    3. Environment variables (prefixed with {prefix})
    4. CLI arguments <<explain the format you expect>>
```

Point 4. is especially important because I can't quite tell from reading this code how I should pass config overrides in via the CLI.

src/blueapi/cli/cli.py
```diff
@@ -37,11 +36,21 @@
 @click.option(
     "-c", "--config", type=Path, help="Path to configuration YAML file", multiple=True
 )
+@click.option("--env-prefix", default="APP_", help="Prefix for environment variables")
```
callumforrester, line 39: Should: I think the prefix should default to `BLUEAPI`, `APP` is too generic.

src/blueapi/config.py
```diff
@@ -12,6 +14,55 @@
...
+    for key, value in ctx_params.items():
+        if value is not None and "." in key:
+            # Split the key by dot and create nested dictionary structure
+            keys = key.split(".")
+            get_nested_value(keys, value, cli_values)
+
+    return cli_values
+
+
+def pretty_print_errors(errors):
```
callumforrester, line 50: Nit: This function formats, it doesn't print

```suggestion
def pretty_format_errors(errors):
```

src/blueapi/config.py
```diff
@@ -103,12 +154,25 @@ def __eq__(self, other: object) -> bool:
 C = TypeVar("C", bound=BaseModel)
 
 
+def recursively_updated_map(
```
callumforrester, line 157: Could: Make private

```suggestion
def _recursively_updated_map(
```

src/blueapi/config.py
callumforrester, line 50: Could: Move to somewhere in `utils` since it's generically useful.

callumforrester, line 50: Should: Type hints

src/blueapi/cli/cli.py
```diff
@@ -50,14 +59,25 @@ def main(ctx: click.Context, config: Path | None | tuple[Path, ...]) -> None:
...
     ctx.ensure_object(dict)
     loaded_config: ApplicationConfig = config_loader.load()
 
+    # Store config in the Click context object for subcommands to access
     ctx.obj["config"] = loaded_config
+
+    # Set up logging based on loaded configuration
     logging.basicConfig(level=loaded_config.logging.level)
 
+    # If no subcommand is invoked, print a help message
```
callumforrester, line 78: Nit: The help message is taken from `--help` and is not the same as this

```suggestion
    # If no subcommand is invoked, print an error message
```

tests/unit_tests/test_cli.py
callumforrester, line 1: Should: Add tests that path `os.environ` (see [dodal tests here](https://github.com/DiamondLightSource/dodal/blob/main/tests/unit_tests/test_cli.py) for example) and pass CLI config args to prove that works.

callumforrester (changes requested):

src/blueapi/cli/cli.py
```diff
@@ -50,14 +59,25 @@ def main(ctx: click.Context, config: Path | None | tuple[Path, ...]) -> None:
             else:
                 raise FileNotFoundError(f"Cannot find file: {path}")
 
+    # Step 3: Load values from environment variables
+    config_loader.use_values_from_env(env_prefix)
+
+    # Step 4: Load CLI arguments as overrides
+    config_loader.use_values_from_cli(ctx.params)
```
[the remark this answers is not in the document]
stan-dot: not sure what is the exact diff from what is there now at the moment.

stan-dot:
Thanks @callumforrester  for the review, I'll implement shortly

src/blueapi/config.py
```diff
@@ -117,54 +181,48 @@ def __init__(self, schema: type[C]) -> None:
...
-        Finalize and load the config as an instance of the `schema`
-        dataclass.
+        Load values from environment variables with a given prefix.
+        """
+
+        env_values = {}
+        for key, value in os.environ.items():
+            if key.startswith(prefix):
+                # Convert key to a config path-like structure
+                config_key = key[len(prefix) :].lower()
```
tpoliaw, line 205:
```suggestion
                config_key = key.removeprefix(prefix).lower()
```

Should this also use the same method as the CLI parsing? eg `BLUEAPI_CONFIG_FOO=bar` sets `values['config']['foo'] = "bar"` instead of relying on eval (which presumably requires something like `BLUEAPI.CONFIG='{"foo": "bar"}'`)?

tests/unit_tests/test_config.py
```diff
@@ -147,3 +147,121 @@ def test_auth_from_env_throws_when_not_available():
...
     with pytest.raises(KeyError):
         BasicAuthentication(username="${baz}", password="baz")
+
+
+def test_single_dot_notation():
+    """Test with a single dot notation key."""
+    ctx_params = {"BLUEAPI.config.api.host": "my_host"}
+    expected = {"BLUEAPI": {"config": {"api": {"host": "my_host"}}}}
+    result = parse_cli_context(ctx_params)
+    assert result == expected, f"Expected {expected}, but got {result}"
```
tpoliaw, line 157: Doesn't pytest already include expected vs actual context in its exceptions?

src/blueapi/cli/cli.py
tpoliaw, line 66: Does this work? Doesn't click prevent you from passing anything other than `--config` or `--env_prefix`

src/blueapi/config.py
```diff
@@ -12,6 +14,55 @@
...
+        dictionary[keys[-1]] = value
+
+    cli_values: dict[str, Any] = {}
+
+    # Handle dot notation (e.g., BLUEAPI.config.api.host)
+    for key, value in ctx_params.items():
+        if value is not None and "." in key:
+            # Split the key by dot and create nested dictionary structure
+            keys = key.split(".")
+            get_nested_value(keys, value, cli_values)
```
tpoliaw, line 45: Is there any way at this stage for value to be `None`? Why are args with no `.` ignored?
```suggestion
    recursive_dict = lambda:defaultdict(recursive_dict)
    cli_values: dict[str, Any] = recursive_dict()

    # Handle dot notation (e.g., BLUEAPI.config.api.host)
    for key, value in ctx_params.items():
        *path, key = key.split('.')
        if not path: continue
        local = cli_values
        for seg in path: local = local[seg]
        local[key] = value
```

src/blueapi/config.py
```diff
@@ -117,54 +181,48 @@ def __init__(self, schema: type[C]) -> None:
...
+        """
         try:
             return self._adapter.validate_python(self._values)
         except ValidationError as exc:
+            pretty_error_messages = pretty_print_errors(exc.errors())
+
             raise InvalidConfigError(
-                "Something is wrong with the configuration file: \n"
+                f"""Something is wrong with the configuration file:
+                {pretty_error_messages}"""
```
tpoliaw, line 227: Does this end up with the first error being weirdly indented?

tests/unit_tests/test_config.py
```diff
@@ -147,3 +147,121 @@ def test_auth_from_env_throws_when_not_available():
...
+
+
+def test_recursive_update():
+    """Test recursive update when both old and new have nested dictionaries."""
+    old = {"a": {"x": 1, "y": 2}, "b": 3}
+    new = {"a": {"y": 20, "z": 30}, "c": 4}
+    expected = {"a": {"x": 1, "y": 20, "z": 30}, "b": 3, "c": 4}
+    result = recursively_updated_map(old, new)
+    assert result == expected, f"Expected {expected}, but got {result}"
```
tpoliaw, line 227: Should this also test that old wasn't modified?

tpoliaw (commented):
Definitely needs docs on how to use this. I can't see a way of passing options to the CLI that aren't blocked as unrecognised by click. It would also be good to have a list of options that are expected or available

src/blueapi/cli/cli.py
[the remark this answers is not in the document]
stan-dot: ah I see I was looking at the wrong step. I see what you mean now @callumforrester , yet the `could/should` is inconsistent in your comment. I'm pretty just `config` will always be just 1 yaml file, that was the scope for the PR and issue. 

levels of abstraction are only a means to an end (functionality) https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

src/blueapi/cli/cli.py
stan-dot: I don't think it'd be configurable, not sure

tests/unit_tests/test_config.py
stan-dot: not sure

src/blueapi/cli/cli.py
stan-dot: I'd need to look into this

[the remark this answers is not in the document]
tpoliaw: The `config` option already had the `multiple=True` option so the multiple yaml file support was already there.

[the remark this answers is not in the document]
stan-dot: ok will fix this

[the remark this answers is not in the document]
stan-dot: @callumforrester what is the use case for multiple yaml files? we'd need a field reconciliation strategy. it's an overhead to support, what is the business use case for this?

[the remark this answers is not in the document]
callumforrester: It was originally implemented by @DiamondJoseph, who can comment if it is still needed. However even if it isn't I think support should be removed in a future PR which means this one should still support it.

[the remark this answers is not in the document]
stan-dot: I see. the support was just 'let's use all the values without a reconciliation strategy'.

I will persist it for this PR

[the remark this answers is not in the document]
tpoliaw: Isn't it the same as you're doing for CLI args and environment variables? Everything gets merged as new values are added so the last config file added 'wins' when there are collisions.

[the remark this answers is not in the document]
stan-dot: but why would we need such a behavior?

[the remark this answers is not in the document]
tpoliaw: Initial thought would something like to let stomp config be in a separate file to the main config. You could then enable or disable stomp without having to edit anything.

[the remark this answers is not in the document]
stan-dot: that is very specific, thanks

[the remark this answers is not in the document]
DiamondJoseph: This was required previously when we weren't mounting the Secret with the RabbitMQ password as elegantly, so we could have two configs: one encrypted as a secret, and one plaintext as a ConfigMap.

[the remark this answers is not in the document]
stan-dot: are you in favour of keeping it?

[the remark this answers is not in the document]
stan-dot: in person discussion with @DiamondJoseph  - does  not cost to support multiple files, we might just change how stomp consumes the secret in a separate PR to that repo

src/blueapi/cli/cli.py
stan-dot:
```python
    """
    BlueAPI CLI, see below for options.
    The configuration is loaded in the following order,
    each step overriding the previous one:
      1. Hardcoded defaults
      2. File (via -c option)
      3. Environment variables (prefixed with BLUEAPI_)
      4. CLI arguments (in a dot format like `--app.logging.level=DEBUG`)
    """
    
    ```
    
    
this as good enough - verified by @DiamondJoseph

File: src/blueapi/cli/cli.py (hunk 2 above)
Comment (stan-dot) [reply src/blueapi/cli/cli.py]: deleted

File: src/blueapi/cli/cli.py [hunk 12]
```diff
@@ -50,14 +60,30 @@ def main(ctx: click.Context, config: Path | None | tuple[Path, ...]) -> None:
...
+    # Store config in the Click context object for subcommands to access
     ctx.obj["config"] = loaded_config
+
+    # Set up logging based on loaded configuration
     logging.basicConfig(level=loaded_config.logging.level)
 
+    # If no subcommand is invoked, print a help message
     if ctx.invoked_subcommand is None:
-        print("Please invoke subcommand!")
+        print("Please invoke a subcommand!")
```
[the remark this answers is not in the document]
stan-dot: outdated

tests/unit_tests/test_config.py
stan-dot: yes right

src/blueapi/cli/cli.py
stan-dot: just deleted

callumforrester (approved):
LGTM, but please do not much until `main` is fixed and CI is passing again!

src/blueapi/config.py
```diff
@@ -117,54 +133,47 @@ def __init__(self, schema: type[C]) -> None:
...
+        for key, value in os.environ.items():
+            if key.startswith(ENV_PREFIX) and value is not None:
+                # Convert key to a config path-like structure
+                config_key = key.removeprefix(ENV_PREFIX).lower()
+                li: list[str] = config_key.split(".")
+                local_value = env_values
+                while len(li) > 1:
+                    key = li.pop(0)
+                    local_value = local_value.setdefault(key, {})
+                local_value.setdefault(li[0], value)
```
tpoliaw, line 163:
```suggestion
                *path, key = config.split('_') # See ENV_PREFIX
                local_value = env_values
                for entry in path:
                    local_value = local_value.setdefault(entry, {})
                local_value.setdefault(key, value)
```

This should probably also handle the error case where keys are incompatible

```
BLUEAPI_config_foo=3 BLUEAPI_config_foo_bar=4 blueapi serve
```
currently (assuming the `.`/`_` problems are fixed) fails with
```
AttributeError: 'str' object has no attribute 'setdefault'
```
which isn't very informative

src/blueapi/config.py
```diff
@@ -7,10 +8,12 @@
 from bluesky_stomp.models import BasicAuthentication
 from pydantic import BaseModel, Field, TypeAdapter, ValidationError
 
-from blueapi.utils import BlueapiBaseModel, InvalidConfigError
+from blueapi.utils import BlueapiBaseModel, InvalidConfigError, format_errors
 
 LogLevel = Literal["NOTSET", "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
 
+ENV_PREFIX = "BLUEAPI."
```
tpoliaw, line 15: This should be `BLUEAPI_`. You can't easily set environment variables containing '.' from the shell which makes this a pain.

src/blueapi/cli/cli.py
```diff
@@ -38,10 +37,23 @@
...
+def main(
+    ctx: click.Context, config: Path | None | tuple[Path, ...], env_prefix: str
+) -> None:
+    """
+    BlueAPI CLI, see below for options.
+    The configuration is loaded in the following order,
+    each step overriding the previous one:
+      1. Hardcoded defaults
+      2. Files (via -c option)
+      3. Environment variables (prefixed with BLUEAPI_)
```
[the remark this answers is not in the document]
tpoliaw: Also needs the cli line removing

tpoliaw (changes requested):
It looks like a fair few other things need updating as well after the CLI stuff was removed but the CI is catching most of it

src/blueapi/config.py
stan-dot: good catch

src/blueapi/config.py
stan-dot: hmmm, that was in the acceptance criteria

callumforrester: Yep, no objection

DiamondJoseph: Both python variable names and environment variables are only guaranteed to support `[A-Za-z_][A-Za-z0-9_]?`, which means that trying to parse nested config is a nuisance...
Currently it builds the complete `dict[str, str]` before ever trying to parse it as a Config object, so we can't check the keys of the current object. May have to do that, or limit the config from having keys with an underscore.
`WorkerEventConfig` and `ScratchRepository` already break that.

stan-dot: I'm not sure what is your conclusion here @DiamondJoseph

callumforrester: Conclusion is that underscores shouldn't work either, @tpoliaw are you happy with just doing

```shell
export "BLUEAPI.foo.bar=baz"
```

...with the quotes, from the command line?

tpoliaw: Not really...

````sh
$ export "BLUEAPI.foo.bar=baz"
export: not valid in this context: BLUEAPI.foo.bar

Comment (tpoliaw) [reply src/blueapi/config.py:15]:
```sh
env 'BLUEAPI.config.foo=fizz' 'BLUEAPI.config.foo.bar=3' blueapi serve
```

Is the only way I've found so far which isn't great for discoverability.

callumforrester: That's annoying, there are other schemes we could use I guess but they obfuscate the truth a bit more

stan-dot: @DominicOram as the original issue author how do you feel about dots and underscores?
https://github.com/DiamondLightSource/blueapi/issues/495#issuecomment-2342885860

DominicOram: I agree with @tpoliaw that we should be able to set it from the shell. Otherwise I don't have a preference as long as it's documented

callumforrester: I'm actually a fan of

```
env 'BLUEAPI.config.foo=bar' blueapi serve
```

If we put it in the `--help` output, is any less discoverable than

```
BLUEAPI.config.foo=bar blueapi serve
```

stan-dot: > If we put it in the --help output, is any less discoverable than

were you meaning to put a question mark at the end here?

callumforrester: Yes, sorry

stan-dot: is it picking up ENV values from the environment though?

stan-dot:
actually there should be a package for that

stan-dot:
@DominicOram  please take a look at the examples for this https://github.com/PasaOpasen/py-env-parser?tab=readme-ov-file#examples


if it's not right, the logic is just 200 lines of code so we could copy paste into this repo  and adjust should we want to modulate the API - but arguably this would be useful across DLS

src/blueapi/config.py
callumforrester: Should be, but I can't even echo it from the terminal

```
[vid18871@diamrl6120-linux ~]$ env 'BLUEAPI.config.foo=bar' echo ${BLUEAPI.config.foo=bar}
bash: ${BLUEAPI.config.foo=bar}: bad substitution
```

Actually, how about `/` as the delimiter?

stan-dot:
@callumforrester  it promises a pretty good level of support. https://docs.pydantic.dev/latest/concepts/pydantic_settings/#nested-model-default-partial-updates


thanks @Relm-Arrowny  for finding this, as a feature of pydantic it looks more solid than the library I found

DominicOram:
> @DominicOram please take a look at the examples for this https://github.com/PasaOpasen/py-env-parser?tab=readme-ov-file#examples
> 
> if it's not right, the logic is just 200 lines of code so we could copy paste into this repo and adjust should we want to modulate the API - but arguably this would be useful across DLS

As just a user of BlueAPI I don't have strong or well-informed opinions about implementation details and I feel like any opinion I do give will just muddy the water. So I defer to @callumforrester and @tpoliaw as the maintainers of BlueAPI to make any decisions around this issue.

stan-dot:
closing as the implementation with pydantic will be entirely different 

https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values

src/blueapi/config.py
DiamondJoseph: `/` has the same issue as `.`: `[A-Za-z_][A-Za-z0-9_]?` is the only guaranteed supported (on all (?) shells) regex pattern for both env_vars AND python field names: there is no character we can use as a separator without restricting its use in the python config names. And the only non-alphanumeric character we could use in that case is `_`, which is currently used: it could be removed, but then either all our config is restricted to 1 word or we're doing unpythony things like camelCase.

DiamondJoseph: https://stackoverflow.com/a/75709846

PR #640, opened 2024-09-17
Update bluesky-stomp and prevent duplication of BasicAuthentication

DiamondJoseph:
Fixes #636 

Current parsing of config passes a blueapi BasicAuthentication with fields username, passcode as an instance of bluesky-stomp BasicAuthentication with fields username, password. I am not certain that we would be able to pass the former as an instance of the latter even if their arguments agreed, as they are different Pydantic models. We can reuse bluesky-stomp's BasicAuthentication as part of the config of blueapi and not duplicate the logic.

PR #650, opened 2024-09-25
629 speedup GitHub actions

### LINKED ISSUE #629: Speedup github actions
They take a long time and if we do some smart caching we could speed this up, freeing up resources.

Resources on that caching
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows
- https://www.warpbuild.com/blog/github-actions-speeding-up
- https://jcdan3.medium.com/4-ways-to-speed-up-your-github-action-workflows-a0b08067a6c6

stan-dot:
Fixes #629

### DIFF
--- a/.github/actions/install_requirements/action.yml
+++ b/.github/actions/install_requirements/action.yml
@@ -7,6 +7,12 @@ inputs:
   pip-install:
     description: Parameters to pass to pip install
     default: "$([ -f dev-requirements.txt ] && echo '-c dev-requirements.txt') -e .[dev]"
+  workflow-name:
+    description: Which workflow started this.
+    default: "help"
+  token:
+    description: Git token
+    required: true
 
 runs:
   using: composite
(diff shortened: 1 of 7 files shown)


stan-dot:
@gilesknap how can I actually measure if this was successful in speeding up? I'd like to compare against a recent successful merge like this one
https://github.com/DiamondLightSource/blueapi/pull/640

stan-dot:
maybe I do need to add 'cache-from to all the workflows?

stan-dot:
it applies only when we run a new branch but the cache hits - so this PR's pipeline time is normal as it still had to create the cache.

has this kind of caching been considered for hyperion, @DominicOram ? or have there been any counterindications?

DominicOram:
> has this kind of caching been considered for hyperion, @DominicOram ? or have there been any counterindications?

Our actions don't take too long and developers generally run the test suite locally before pushing to catch errors anyway so it's rare the time it takes for actions to run slows us down at all.

DominicOram:
I would also worry about caching too much and missing things, it's quite valuable to know that a PR works from a clean slate.

gilesknap:
> I would also worry about caching too much and missing things, it's quite valuable to know that a PR works from a clean slate.

Note that with respect to container build cache it only hits on layers that have not changed - However, I suppose that if a layer is fetching a remote resource and not locking the version of that resource then the cache could mask issues.

stan-dot:
@callumforrester  based on all the discussion above, is the functionality from this PR desirable and pursuing more caching is not needed at the moment?

stan-dot:
still I think we could lint - test - build docs - dist all in one job, using one cache. always clearing the cache on the first job, if possible

Relm-Arrowny:
Just a note, the way caches is set up on this atm will always run pip install after caching the env which mean it will update before running test so it should not miss anything and it should also down grade correctly.  
As for making it lint - test - build docs - dist all in one job, it is possible but you lose a bit of the speed as they are slightly different so they will have to install those, plus because of the racing condition of the single caches you lose a bit more time time as you only get to update cache the one finish first which is most likely be the dist or lint.

callumforrester:
Agree with @Relm-Arrowny, additionally if we condensed all the CI into one job we would lose the nice display showing which parts have passed and which parts have failed, which I find useful when reviewing a PR.
@stan-dot I'm happy to examine your workflow with you and see if we can figure out why the CI is bottlenecking you, it isn't bottlenecking me.

callumforrester:
@stan-dot I'm inclined to close this unless you seriously object

stan-dot:
@callumforrester I don't mean 'all CI in one job', just to make the docs go off the `dev` tests, and there also the dist.


It is not bottlenecking me personally

stan-dot:
@callumforrester I agree we don't need to deepen the caching level. is the caching in this PR not fine? _some_ caching is a best practice afaik

callumforrester:
It depends, I'm unclear on whether caching was not included in the copier template due to lack of time or for a specific reason. @coretl or @gilesknap can probably comment.

gilesknap:
I believe we worked out that the costs were greater than the benefits for a small python container. I do use cache to great effect on generic IOCs. In fact reflecting on how fast container builds on GitHub were just yesterday, for iic-adandor3, maybe this is worth revisiting.

gilesknap:
Oh just realized that this is not about container build. In case it was not obvious my comments were about containers. We did look at cache for copier template container build only and decided against.

coretl:
I would prefer to investigate how long it would take to `pixi install` the deps rather than `pip install`. Wheels are just zip files after all, so if the installed (`pixi` rather than `pip`) is fast enough, then it should get close to the speed of restoring a cache without the complexity of cache invalidation (without a lockfile we intentionally pull the latest version of things).

stan-dot:
@coretl `pixi` does not support package publishing and building at the moment https://pixi.sh/dev/switching_from/poetry/#quick-look-at-the-differences

callumforrester:
@coretl Strong vote against uprooting our toolchains once again

callumforrester:
@stan-dot is #629 still a problem for you? I continue to have a preference against diverging our CI from the copier template unnecessarily, I'd like to close this PR.

stan-dot:
well the idea of changes is to test things and then potentially push them upstream to the template

callumforrester:
Agreed, since there hasn't been much response from the template maintainers, I'll go ahead and close this PR for now.

PR #656, opened 2024-10-02
Caches pip download and env

Relm-Arrowny:
#650 #629  I made an attempt caching the python env, it gain you about 30-90 per work flow build not sure if it is worth it or not. We can probably play the same game with the container part of the work-flow which is where the real bottle neck is. 

@stan-dot Anyway see what you think,

### DIFF
--- a/.github/actions/install_requirements/action.yml
+++ b/.github/actions/install_requirements/action.yml
@@ -25,7 +25,7 @@ runs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ env.PYTHON_VERSION }}
-      #  cache: "pip"
+        cache: "pip"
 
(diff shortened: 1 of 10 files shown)


stan-dot:
thanks, 

this looks good

Monday I'll look into the container part - this 'cache-from:' is a field I missed at first
https://docs.docker.com/build/cache/backends/gha/

```
- name: Build and push
  uses: docker/build-push-action@v6
  with:
    context: .
    push: true
    tags: "<registry>/<image>:latest"
    cache-from: type=gha
    cache-to: type=gha,mode=max
    github-token: ${{ secrets.MY_CUSTOM_TOKEN }}
    
    ```

PR #664, opened 2024-10-11
Update sonarcloud.yml with the condition for this repo instance

stan-dot:
> The sonarcloud workflow requires an auth key, so will not work with PRs from forks, which we want to support. We will therefore skip that job when the PR is from a fork.

thanks @callumforrester


callumforrester:
Just to quickly document the reasons: The sonarcloud workflow requires an auth key, so will not work with PRs from forks, which we want to support.

callumforrester (approved):

PR #668, opened 2024-10-11
Include convenience bash script for running rabbit mq locally

### LINKED ISSUE #615: Include convenience bash script for running RabbitMQ locally
## Acceptance Criteria
- Bash script should be supplied in repo
- It should run with `docker` or `podman` as a fallback if `docker` is not present
- It should start up RabbitMQ with the STOMP plugin enabled on `localhost:61613` (the default port)
- There should also be a blueapi config file for connecting to it

Relm-Arrowny:
fix #615


docs/tutorials/quickstart.md
```diff
@@ -17,6 +17,12 @@ way to start it is to run it via a container:
     podman run -it --rm --net host rmohr/activemq:5.15.9-alpine
 ```
 
+Alternatively, a script is available:
```
callumforrester, line 20: Should: I think we should push this as the preferred way the developer runs rabbit rather than an alternative

callumforrester (changes requested):
Looking good, thanks! Just a few small changes.

docs/tutorials/quickstart.md
```diff
@@ -4,10 +4,17 @@ Blueapi acts as a worker that can run bluesky plans against devices for a specif
...
-## Start ActiveMQ
+## Start RabbitMQ
 
-The worker requires a running instance of ActiveMQ, the simplest
-way to start it is to run it via a container:
+The worker requires a running instance of RabbitMQ. The easiest way to start it is
+ to execute the provided script:
+
+```
+    src/script/start_rabbitmq.sh
```
DominicOram, line 13: Must: This fails with the error `Error: context must be a directory: "/blueapi/rabbitmq_setup"`. I think the script only works if your current directory is `blueapi/src/scripts`. Can we allow it to be run from the top please?

docs/tutorials/quickstart.md
```diff
@@ -4,10 +4,17 @@ Blueapi acts as a worker that can run bluesky plans against devices for a specif
...
+```
+
+
+Alternatively, it is possible to run ActiveMQ via a container:
```
DominicOram, line 17: Should: I couldn't get ActiveMQ to work (https://github.com/DiamondLightSource/blueapi/issues/694). Given this is a quickstart and it will be read by a new user I suggest we remove the option of ActiveMQ here and give the user the simplest configuration that works. If we want them to also be able to use ActiveMQ maybe we could put that in a new page just about that usecase?

docs/tutorials/quickstart.md
```diff
@@ -31,6 +39,12 @@ The worker can also be started using a custom config file:
     blueapi --config path/to/file serve
 ```
 
+An example of a config file that starts STOMP with default values can be found in:
+
+```
+    src/script/stomp_config.yml
+```
+
```
DominicOram, line 47: Should: As above with ActiveMQ vs RabbitMQ, if we're going to make rabbitMQ the default for new users then maybe this should be the default config?

DominicOram (changes requested):
I came here after not being able to get ActiveMQ to work but this works great, thanks! Some comments in code though. Particularly I think the path issue is important

docs/tutorials/quickstart.md
Relm-Arrowny: #694 is cause by the fact that at some point blueapi have stop setting default value [here](https://github.com/DiamondLightSource/blueapi/blob/627ffc3245ab85ccad838c968813f9023daa165d/src/blueapi/config.py#L86C13-L86C14)

DominicOram: Ok, I still think that's it's confusing to a new user to be faced with both Active and Rabbit MQ but defer to @callumforrester @DiamondJoseph

callumforrester: Yeah, I think it's fair enough to remove activemq. Long term I want to:
1. Make the CLI work without a message bus (albeit with less features)
2. Split the quickstart into just starting the server and running plan, and then a separate tutorial for connecting a message bus and getting rich feedback (I've actually already done the split in #674 so this should be rebased on top of that).

The idea is to make the quickstart as simple as possible.

docs/tutorials/quickstart.md
callumforrester: It used to be, but I changed it. I think it's important for a complete beginner to clone -> build -> run with no extra steps, because once you add 1 extra step it's very easy to get into a "just one more..." mentality. The message bus is an optional feature that adds rich feedback and that we use in production, but configuring it is part 2 of the tutorial.

DominicOram (approved):
My comments on the docs are addressed by https://github.com/DiamondLightSource/blueapi/pull/674, thanks @callumforrester

callumforrester:
The sonarcloud error should be fixed by https://github.com/DiamondLightSource/blueapi/pull/664
Once that's merged could you rebase on top of it to check it works?

docs/tutorials/quickstart.md
```diff
@@ -17,6 +24,7 @@ way to start it is to run it via a container:
     podman run -it --rm --net host rmohr/activemq:5.15.9-alpine
```
callumforrester, line 24: Should: Also remove this reference to activemq

Relm-Arrowny: I have removed activemq reference, a couple of things I notice while doing so which are the side effects of removing the Storm default:

1.  The cli will forget/reload the env every time so
 [`blueapi controller run count '{"detectors": ["current_det", "image_det"]}`](https://github.com/DiamondLightSource/blueapi/blob/24c7e35820cdcfb31b9876ef7d9798590a363dec/docs/how-to/run-cli.md?plain=1#L33)
in the quickstart guide will not work unless we set the env with #649 or do 
 ` blueapi -c src/script/stomp_config.yml controller run count '{"detectors": ["current_det", "image_det"]}'`
2. `blueapi serve` will technically work but there will be half a page of warning ever 10 secs.

src/script/rabbitmq_setup/Dockerfile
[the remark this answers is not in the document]
callumforrester: @Relm-Arrowny this is the only concern I have left, is it possible to run `rabbitmq:3.13.2-management` directly with podman and mount in the configuration you want?

docs/tutorials/quickstart.md
callumforrester: Point 1 is addressed with #674

docs/tutorials/quickstart.md
DiamondJoseph: I've proposed adding a [RMQ Dockerfile](https://github.com/DiamondLightSource/blueapi/blob/026e3993a0d1d425177b55f9f43d6d828d7b0c52/tests/system_tests/Dockerfile) that installs the stomp plugin to the system_tests to make running those easier: this could be promoted out of that test dir and be a sanctified way of running a message bus for development?

DiamondJoseph: The CLI does (did?) work without the message bus, but without the ability to run plans (as we can't catch messages about the plan ending)

callumforrester: I was [fairly against having our own Dockerfile](https://github.com/DiamondLightSource/blueapi/pull/668#discussion_r1822464761) (and therefore, one more setup step for the user). Is there no way of doing this with a vanilla image from Dockerhub and mounted config files?

DominicOram: > The CLI does (did?) work without the message bus, but without the ability to run plans

As I said to @callumforrester in person. As a naïve user of `blueAPI` I feel like running plans is it's core function. To me "working without being able to run plans" is equivalent to not working.

Relm-Arrowny: > I was [fairly against having our own Dockerfile](https://github.com/DiamondLightSource/blueapi/pull/668#discussion_r1822464761) (and therefore, one more setup step for the user). Is there no way of doing this with a vanilla image from Dockerhub and mounted config files?

I just about to remove the dockerfile

src/script/rabbitmq_setup/Dockerfile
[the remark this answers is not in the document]
DiamondJoseph: I was only going to add rabbitmq_stomp and ensure the management console was there, so those two plus prometheus seem fine. I'm just going to check out this branch and try it out quick- I assume everything in the src/scripts dir was found by testing to be required?

callumforrester (approved):
LGTM!

src/script/start_rabbitmq.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+SCRIPT_DIR=$( cd -- "$( dirname -- "$BASH_SOURCE[0]" )" &> /dev/null && pwd )
+RABBITMQ_VERSION="rabbitmq:3.13.2-management"
```
DiamondJoseph, line 3:
```suggestion
RABBITMQ_VERSION="rabbitmq:management"
```

Do we mind a particular version? For system test purposes we want to check we continue to work with all new versions

src/script/rabbitmq_setup/Dockerfile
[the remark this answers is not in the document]
Relm-Arrowny: > I was only going to add rabbitmq_stomp and ensure the management console was there, so those two plus prometheus seem fine. I'm just going to check out this branch and try it out quick- I assume everything in the src/scripts dir was found by testing to be required?

yes, minimum to run blueapi with stomp.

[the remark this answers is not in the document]
DiamondJoseph:
```sh
#!/bin/bash
SCRIPT_DIR=$( cd -- "$( dirname -- "$BASH_SOURCE[0]" )" &> /dev/null && pwd )
cmd1='run -it --rm --name rabbitmq -v '$SCRIPT_DIR'/rabbitmq_setup/enabled_plugins:/etc/rabbitmq/enabled_plugins -p 5672:5672 -p 15672:15672 -p 61613:61613 rabbitmq:management'

echo "Checking docker/podman installation"
if command -v docker &> /dev/null; then
    docker $cmd1
elif command -v podman &> /dev/null; then
    podman $cmd1
else
    echo "Docker/Podman installation not found. Please install docker/podman."
    exit 1
fi
```

I found locally that this and not needing the rabbitmq.conf was enough for my local blueapi to connect- blueapi wouldn't be able to connect *not from this machine* (and it may be an issue on Diamond workstations maybe, if so we need to keep it?) without specifying the loopback settings.

DiamondJoseph (approved):

src/script/start_rabbitmq.sh
Relm-Arrowny: shall I change this too?

src/script/rabbitmq_setup/Dockerfile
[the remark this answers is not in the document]
Relm-Arrowny: right doing it now.

Issue #644, opened 2024-09-17
Replace PlanModel using the __doc__ of a plan with using Annotated pydantic fields


DiamondJoseph:
`blueapi controller plans` returns a description of the plans that are available in the context. Currently this comes from the docstring of the plan function. This should get also/instead(?) the description of the Pydantic model that is constructed, including information about the fields of the plan: defaults, types, limits etc.

Using `Annotated` allows us to have this information presented as part of the json schema, and reduce the amount of information required in the docstring.

DiamondJoseph:
e.g.

```python
def stp_snapshot(
    detectors: list[Readable],
    temperature: Movable = inject("sample_temperature"),
    pressure: Movable = inject("sample_pressure"),
) -> MsgGenerator:
    """
    Moves devices for pressure and temperature (defaults fetched from the context)
    and captures a single frame from a collection of devices

    Args:
        detectors (List[Readable]): A list of devices to read while the sample is at STP
        temperature (Optional[Movable]): A device controlling temperature of the sample,
            defaults to fetching a device name "sample_temperature" from the context
        pressure (Optional[Movable]): A device controlling pressure on the sample,
            defaults to fetching a device name "sample_pressure" from the context
    Returns:
        MsgGenerator: Plan
    Yields:
        Iterator[MsgGenerator]: Bluesky messages
    """
```

then becomes

```python
def stp_snapshot(
    detectors: set[Readable] = Annotated[
        Field(set[Readable], min_length=1), "Devices to read while the sample is at STP"
    ],
    temperature: Movable = Annotated[
        inject("sample_temperature"), "A device controlling temperature of the sample"
    ],
    pressure: Movable = Annotated[
        inject("sample_pressue"), "A device controlling pressure of the sample"
    ],
) -> MsgGenerator:
    """
    Moves devices for pressure and temperature (defaults fetched from the context)
    and captures a single frame from a collection of devices
    """
```

And the information about the detector minimum length (bad example) gets propagated through.

DiamondJoseph:
This must also propagate the value of any defaults into the json schema.

callumforrester:
I'm not entirely happy with this, since we want to use standard tools/methods as far as possible and make the process of migrating a plan that someone wrote in their IPython environment to blueapi as painless as possible. People are more likely to have plans with standard docstrings than annotated arguments so we're just adding an extra step here.

Could we add some special handling to parse docstrings and get the results into a JSON schema?

[...]

PR #674, opened 2024-10-17
Overhaul documentation

### LINKED ISSUE #361: Review CLI docs
It has been suggested that our CLI docs are out of date. This issue is just a reminder to review and remedy where necessary.

### LINKED ISSUE #447: Review and update docs
Some information is out of date or in the wrong place, in particular a lot of detailed technical explanation has made into "how-tos" when it should probably be in "explanations"

## Acceptance Criteria
- The docs have been reviewed
- Information that is no longer true has been removed
- Information has been refactored into its proper place in the structure

### LINKED ISSUE #671: Document Dev Environment
## Acceptance Criteria
- Docs exist on how to run blueapi locally with
  - Message bus
  - Custom plans/devices

callumforrester:
Fixes #447 
Fixes #361 
Fixes #671


docs/explanations/architecture.md
```diff
@@ -9,26 +8,7 @@ Blueapi performs a number of tasks:
...
-the known expectations of the plan, passes it to the `RunEngine` and handles any errors.
-
+![blueapi main components](../images/blueapi.png)
 
-### The Service Object
 
-Handles communications and the API layer. This object holds a reference to the worker 
-can interrogate it/give it instructions in response to messages it receives from the message
-bus. It can also forward the various events generated by the worker to topics on the bus.
+Above are the main components of blueapi. The main process houses the REST API and manages the subprocess, which wraps the `RunEngine`, devices and optional connection to the downstream message broker.
```
DiamondJoseph, line 14:
```suggestion
Above are the main components of blueapi. The main process houses the REST API and manages the subprocess, which wraps the `RunEngine`, devices and external connections.
```
Do we want to be general or specific?

docs/explanations/extension-code.md
```diff
@@ -0,0 +1,17 @@
+# Home of Plans and Devices
+
+Plans and devices can be in any pip-installable package, such as:
```
DiamondJoseph, line 3: I'd move the device/dodal part first, so that it's clear that devices belong in dodal, rather than someone adding devices to their repo then realising it's meant to be in dodal.

docs/explanations/extension-code.md
```diff
@@ -0,0 +1,17 @@
...
+
+The easiest place to put the code is a repository created with the [`python-copier-template`](https://diamondlightsource.github.io/python-copier-template/main/index.html). Which can then become any of the above.
+
+:::{seealso}
+Guide to setting up a new Python project with an environment and a standard set of tools: [`Create a new repo from the template`](https://diamondlightsource.github.io/python-copier-template/main/tutorials/create-new.html)
+:::
+
+## Dodal
+
+[Dodal](https://github.com/DiamondLightSource/dodal) is a repository for DLS device configuration, providing factory functions for devices used at DLS. If you work at DLS, new devices should be added there.
```
DiamondJoseph, line 17: new devices = new device classes?

This is mostly true, get devices into dodal, then plan to move to ophyd-async. Let's treat mostly true as true for the purposes of introductory docs.

Should be expanded to explain that beamlines should also be added there, with a link to dodal how to make a new beamline.

docs/explanations/extension-code.md
```diff
@@ -0,0 +1,17 @@
+# Home of Plans and Devices
+
+Plans and devices can be in any pip-installable package, such as:
+
+* A package on pypi
+* A Github repository
+* A local directory via the [scratch area](../how-to/edit-live.md).
+
+The easiest place to put the code is a repository created with the [`python-copier-template`](https://diamondlightsource.github.io/python-copier-template/main/index.html). Which can then become any of the above.
```
DiamondJoseph, line 9: Could: an example, like i22?

docs/explanations/lifecycle.md
```diff
@@ -2,43 +2,44 @@
...
+from typing import Any, List, Mapping, Optional, Union
+
+import bluesky.plans as bp
+from blueapi.core import MsgGenerator
+from dls_bluesky_core.core import inject
+from bluesky.protocols import Readable
+
+
+def count(
+    detectors: List[Readable] = [inject("det")],  # default valid for Blueapi only
```
DiamondJoseph, line 16: Is this change going in prior to deprecating inject or after?

DiamondJoseph: Update to >=3.10 typing (List->list, Union[x, y] -> x | y, Optional[x] -> x | None)

docs/explanations/lifecycle.md
```diff
@@ -114,13 +111,13 @@ See also [type validators](./type_validators.md)
...
 
-The validated parameter values are then passed to the plan function, which is passed to the RunEngine.
+The validated parameter values are then passed to the plan function, which is passed to the `RunEngine`.
 The plan is executed. While it is running, the `Worker` will publish
 
 * Changes to the state of the `RunEngine`
 * Changes to any device statuses running within a plan (e.g. when a motor changes position)
 * Event model documents emitted by the `RunEngine`
-* When a plan starts, finishes or fails.
+* When the plan starts, finishes or fails.
```
DiamondJoseph, line 120: >* Changes to any device statuses running within a plan (e.g. when a motor changes position)

should be `the` plan as well

docs/explanations/type_validators.md
```diff
@@ -3,44 +3,44 @@
...
 
-    # Internally becomes something like
+# Internally becomes something like
 
-    class MyPlanModel(BaseModel):
-        a: int
-        b: str = "b"
+class MyPlanModel(BaseModel):
+    a: int
+    b: str = "b"
```
DiamondJoseph, line 14: Is this repeating what is explained in plans.md?

docs/explanations/type_validators.md
```diff
@@ -59,30 +59,32 @@ object returned from validator method is not checked by pydantic so it can be
...
-        e: List[Dict[str, Set[Readable]]]) -> Plan:
-        ...
+```python
+def my_weird_plan(
+    a: Readable,
+    b: List[Readable],
+    c: Dict[str, Readable],
+    d: List[List[Readable]],
+    e: List[Dict[str, Set[Readable]]]) -> Plan:
+    ...
```
DiamondJoseph, line 89: Python 3.10ify

docs/how-to/configure-app.md
```diff
@@ -1,18 +1,10 @@
-# Configure the application
+# Configure the Application
 
-By default, configuration options are ingested from pydantic BaseModels,
-however the option exists to override these by defining a yaml file which
-can be passed to the `blueapi` command.
+Blueapi's default configuration can be overridden 
+by defining a yaml file which can be passed to the `blueapi` command.
```
DiamondJoseph, line 4: memo: this is about to get changed by changes to overriding config values from env vars etc, make sure to update it then

docs/how-to/write-devices.md
```diff
@@ -0,0 +1,70 @@
...
+```python 
+from my_facility_devices import MyTypeOfDetector
+
+def my_detector(name: str) -> MyTypeOfDetector:
+    return MyTypeOfDetector(name, {"other_config": "foo"})
+```
+
+The return type annotation `-> MyTypeOfDetector` is required as blueapi uses it to determine that this function creates a device. Meaning you can have a Python file where only some functions create devices and they will be automatically picked up. Similarly, these functions can be organized per-preference into files. 
+
+The device is created via a function rather than a global to preserve side-effect-free imports. Each device must have its own factory function.
```
DiamondJoseph, line 25: Should: defer to dodal's docs on writing device classes and beamlines, and ensure they are up to date?

docs/how-to/write-plans.md
```diff
@@ -1,173 +1,93 @@
...
+            defaults to fetching a device name "sample_pressure" from the context
+        target_pressure (Optional[float]): target temperature in Kelvin. Default 273
+        target_pressure (Optional[float]): target pressure in Pa. Default 10**5
+    Returns:
+        MsgGenerator: Plan
+    Yields:
+        Iterator[MsgGenerator]: Bluesky messages
+    """
+    yield from move({temperature: target_temperature, pressure: target_pressure})
+    yield from count(detectors, 1, metadata or {})
```
DiamondJoseph, line 92: Should: replacing this form of docstring with using Annotated/Field arguments so that the values propagate into json schema?

DiamondJoseph (changes requested):
How much is this just re-formatting the docs to pass current linting/build and how much is it intending to update the docs? There's several things that we're actively moving away from that aren't being changed, but also more changes than just updating the formatting.

docs/explanations/architecture.md
callumforrester: I made this a bit more general because I didn't think a list of major classes and their purposes was _that_ useful and was prone to being out of date. It may be more useful when blueapi is more stable.

docs/explanations/extension-code.md
callumforrester: This is a problem in general, some of these docs belong in dodal with links from blueapi. I think we can and should do that migration gradually though. I'm going to advocate for removing this sentence and making an issue in dodal for adding a page/pages explaining how to:
- Make a new device
- Add a device to a beamline
- Make a new beamline

docs/explanations/lifecycle.md
callumforrester: I am going to make a separate issue for moving all of the code snippets to Python files and checking that they still work and lint properly at docs build time, similar to ophyd-async.

callumforrester: #677

docs/explanations/extension-code.md
DiamondJoseph: The reason I asked is there is already docs in dodal for at least part of that
https://diamondlightsource.github.io/dodal/main/how-to/create-beamline.html
https://diamondlightsource.github.io/dodal/main/reference/device-standards.html (?)

docs/explanations/type_validators.md
callumforrester: Probably does need a bit of consolidation, my read is that plans.md explains which parts of blueapi plans are special and why, and type_validators.md explains the very technical implementation of plan parameter parsing.

docs/explanations/type_validators.md
callumforrester: #677

docs/how-to/write-devices.md
callumforrester: Agreed, see other discussions, I think we should migrate some of this to dodal as a separate PR

docs/how-to/write-plans.md
callumforrester: Oh, I did not realise they didn't propagate, that's quite bad. I was hoping to make the plans as vanilla-bluesky as we could, so vanilla python function docstrings are definitely desirable...

DiamondJoseph: Lemme get a couple of example plans into blueapi and compare the json schemas

callumforrester:
@DiamondJoseph regarding the fact that some of this is going out of date "soon": I have been bitten too many times by writing code and docs in anticipation of changes that are due "in a week or so" and end up taking 6 months. I'm strongly against putting anything in here that is not true in `main` right now. 

Ideally docs should be updated in the same PR as the change they reference, second-best option is after the PR is merged, worst-best is before the PR is merged.

docs/explanations/extension-code.md
callumforrester: Ah cool! I'll reference those

docs/how-to/write-plans.md
DiamondJoseph:
```python
@attach_data_session_metadata_decorator()
@validate_call(config={"arbitrary_types_allowed": True})
def linkam_plan_1(
    trajectory: LinkamTrajectory,
    linkam: Linkam3 = LINKAM,
    shutter_time: float = 0.04,
) -> MsgGenerator:
    """
    Follow a trajectory in temperature, collecting a number of frames either at equally
    spaced positions or while continually scanning. e.g. for 2 segments, the first
    stepped and the 2nd flown:\n
    trajectory start   v             v final segment stop\n
                       \\           /\n
       stepped segment__\\__       /\n
                           \\     /  flown segment\n
           1st segment stop \\__ /\n
        exposures:    xx  xx  xx   1/N seconds

    Args:
        trajectory: Trajectory for the scan to follow.
        linkam: Temperature controller
        shutter_time: Time allowed (s) for opening shutter before triggering detectors.

    Returns:
            MsgGenerator: Plan

    Yields:
            Iterator[MsgGenerator]: Bluesky messages
    """
    ...

@attach_data_session_metadata_decorator()
@validate_call(config={"arbitrary_types_allowed": True})
def linkam_plan_2(
    trajectory: Annotated[
        LinkamTrajectory, Field(description="Trajectory for the scan to follow.")
    ],
    linkam: Annotated[Linkam3, Field(LINKAM, description="Temperature controller.")],
    shutter_time: Annotated[
        float,
        Field(
            0.04,
            description="Time allowed for opening shutter before triggering detectors.",
            json_schema_extra={"units": "s"},
        ),
    ],
) -> MsgGenerator:
    """
    Follow a trajectory in temperature, collecting a number of frames either at equally
    spaced positions or while continually scanning. e.g. for 2 segments, the first
    stepped and the 2nd flown:\n
    trajectory start   v             v final segment stop\n
                       \\           /\n
       stepped segment__\\__       /\n
                           \\     /  flown segment\n
           1st segment stop \\__ /\n
        exposures:    xx  xx  xx   1/N seconds
    """
    ...

@attach_data_session_metadata_decorator()
@validate_call(config={"arbitrary_types_allowed": True})
def linkam_plan_3(
    trajectory: LinkamTrajectory = Field(
        description="Trajectory for the scan to follow."
    ),
    linkam: Linkam3 = Field(LINKAM, description="Temperature controller."),
    shutter_time: float = Field(
        0.04,
        description="Time allowed for opening shutter before triggering detectors.",
        json_schema_extra={"units": "s"},
    ),
) -> MsgGenerator:
    """
    Follow a trajectory in temperature, collecting a number of frames either at equally
    spaced positions or while continually scanning. e.g. for 2 segments, the first
    stepped and the 2nd flown:\n
    trajectory start   v             v final segment stop\n
                       \\           /\n
       stepped segment__\\__       /\n
                           \\     /  flown segment\n
           1st segment stop \\__ /\n
        exposures:    xx  xx  xx   1/N seconds
    """
    ...
```

[linkam_plan_1.json](https://github.com/user-attachments/files/17436590/linkam_plan_1.json)
[linkam_plan_2.json](https://github.com/user-attachments/files/17436591/linkam_plan_2.json)
[linkam_plan_3.json](https://github.com/user-attachments/files/17436656/linkam_plan_3.json)

I've just gone through all that just to prove they're basically identical and the Pydantic model doesn't actually get the information about the Fields :upside_down_face:

callumforrester: I think we can continue this discussion in #644

docs/how-to/write-plans.md
```diff
@@ -1,173 +1,93 @@
...
-       yield from {}
+def touch_synchrotron(sync: Synchrotron = inject("synchrotron")) -> MsgGenerator:
+    # There is only one Synchrotron device, so we know which one it will always be.
+    # If there is no device named "synchrotron" in the blueapi context, it will except.
+    sync.specific_function()
+    yield from {}
 ```
 
-### Metadata
+## Injecting Metadata
```
callumforrester, line 51: `inject` is now deprecated, will update these before merging

docs/explanations/architecture.md
DiamondJoseph: Lost in translation- I'm suggesting a change to make it slightly more generic but hopefully still make sense?

docs/tutorials/run-bus.md
```diff
@@ -0,0 +1,43 @@
...
+The worker can publish updates to a running instance of a message bus such as  ActiveMQ, the simplest
+way to start it is to run it via a container:
+
+```
+docker run -it --rm --net host rmohr/activemq:5.15.9-alpine
+```
+
+```
+podman run -it --rm --net host rmohr/activemq:5.15.9-alpine
+```
```
callumforrester, line 16: Remove per discussion with @DominicOram

docs/explanations/lifecycle.md
```diff
@@ -2,43 +2,44 @@
...
-            metadata (Optional[Mapping[str, Any]], optional): Key-value metadata to include
-                                                            in exported data.
+
+```python
+import bluesky.plans as bp
+
+from typing import Any, List, Mapping, Optional, Union
+
+from blueapi.core import MsgGenerator
+from bluesky.protocols import Readable
```
DiamondJoseph, line 12: Should:

```suggestion
from bluesky.protocols import Readable
from bluesky.utils import MsgGenerator
```

It would be a nit, but we've seen how much of an issue this line has caused!!

docs/explanations/lifecycle.md
```diff
@@ -114,13 +109,13 @@ See also [type validators](./type_validators.md)
 
 ## Execution
 
-The validated parameter values are then passed to the plan function, which is passed to the RunEngine.
+The validated parameter values are then passed to the plan function, which is passed to the `RunEngine`.
 The plan is executed. While it is running, the `Worker` will publish
 
 * Changes to the state of the `RunEngine`
 * Changes to any device statuses running within a plan (e.g. when a motor changes position)
```
DiamondJoseph, line 116: nit

```suggestion
* Changes to any device statuses running within the plan (e.g. when a motor changes position)
```

docs/explanations/plans.md
```diff
@@ -0,0 +1,44 @@
...
+Generally, a `plan` includes at least one `open_run` and `close_run` and is a complete description of an experiment. If it does not, it is a `stub`. This distinction is made in the bluesky core library between the `plan`'s and `plan_stub`'s modules.
+
+
+## Allowed Argument Types
+
+When added to the blueapi context, `PlanGenerator`'s are formalised into their schema - [a Pydantic BaseModel](https://docs.pydantic.dev/1.10/usage/models) with the expected argument types and their defaults. 
+
+Therefore, `PlanGenerator`'s must only take as arguments [those types which are valid Pydantic fields](https://docs.pydantic.dev/dev/concepts/types) or Device types which implement `BLUESKY_PROTOCOLS` defined in dodal, which are fetched from the context at runtime.
+
+Allowed argument types for Pydantic BaseModels include the primitives, types that extend `BaseModel` and `dict`'s, `list`'s  and other `sequence`'s of supported types. Blueapi will deserialise these types from JSON, so `dict`'s should use `str` keys.
```
DiamondJoseph, line 14:
```suggestion
Allowed argument types for Pydantic BaseModels include the primitives, types that extend `BaseModel` and `dict`'s, `list`'s  and other `sequence`'s of supported types. Blueapi will deserialise these types from JSON, so `dict`'s must use `str` keys.
```

nit

docs/explanations/type_validators.md
DiamondJoseph: I think this is just repeating what is in lifecycle.md. It even includes the same json example, I'd really like to see one removed or the difference between the two explained.

docs/how-to/write-devices.md
DiamondJoseph: dodal already has docs on creating a beamline and making a device class. 

https://diamondlightsource.github.io/dodal/main/how-to/create-beamline.html
and
https://diamondlightsource.github.io/dodal/main/how-to/make-new-ophyd-async-device.html

We should defer to these and fix those pages. The former should be updated to use device_factory and the second should be fixed as the Mermaid docs do not render responsively and should explain how to make a Device class.

docs/how-to/write-devices.md
```diff
@@ -0,0 +1,70 @@
...
+
+# How to Configure Detectors to Write Files
+
+:::{note}
+**This is an absolute requirement to write data onto the Diamond Filesystem**. This decorator must be used every time a new data collection is intended to begin. For an example, see below.
+:::
+
+Dodal defines a decorator, `@attach_metadata`, for configuring `ophyd-async` detectors to write data to a common location.
```
DiamondJoseph, line 33: must:
https://github.com/DiamondLightSource/dodal/blob/a3cf0ecf432dc7ad2a62c2583e90d452154f9ef5/src/dodal/plan_stubs/data_session.py#L43
```suggestion
Dodal defines a decorator, `@attach_data_session_metadata_decorator()`, for configuring `ophyd-async` detectors to write data to a common location. 
```

could: explain that it requires the get_path_provider from dodal is used as constructor arg?

docs/how-to/write-plans.md
```diff
@@ -1,173 +1,81 @@
...
 
-> **_NOTE:_** Plans **should** include `metadata` as their final argument, if they do it **must** have the type Optional[Mapping[str, Any]], `and a default of None <https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python>`__\, with the plan defaulting to an empty dict if passed `None`. If the plan calls to a stub/plan which takes metadata, the plan **must** pass down its metadata, which may be a differently named argument.
+Plans ([as opposed to stubs](../explanations/plans.md)) **should** include `metadata` as their final parameter, if they do it **must** have the type `Mapping[str, Any] | None`, [and a default of None](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python). If the plan calls to a stub/plan which takes metadata, the plan **must** pass down its metadata, which may be a differently named parameter.
 
-```
-   def pass_metadata(x: Movable, metadata: Optional[Mapping[str, Any]] = None) -> MsgGenerator:
-       yield from bp.count{[x], md=metadata or {}}
+```python
+def pass_metadata(x: Movable, metadata: Mapping[str, Any] | None = None) -> MsgGenerator:
```
DiamondJoseph, line 46: I think we need to tighten up Mapping -> dict to behave well with bp and bps usage.

DiamondJoseph (changes requested):

docs/how-to/write-plans.md
callumforrester: Sorry, not sure I follow, what issues are you seeing?

DiamondJoseph: bluesky.plans have their equivalent md argument (which we should use when possible) is `CustomPlanMetadata = dict[str, Any]`: we therefore cannot pass a `Mapping[str, Any]` down to it: `Mapping` does not allow assignment, where `dict` does. We also sometimes need to mutate metadata in a "sub"plan before passing it to a "super"plan.

docs/explanations/type_validators.md
callumforrester: So the idea is that `lifecycle.md` is a general overview of how blueapi stores, retrieves and runs plans whereas this is a deep dive into the special handling for substituting devices for strings. Do you have any suggestions as to what we can shuffle around to make that clearer?

docs/how-to/write-devices.md
callumforrester: I agree, but I would rather merge this now and raise separate issues for fixing those pages and consolidating, otherwise this PR will languish for a long time and the published docs (that everyone reads) will continue to contain mistakes that throw people off. Alternatively, I can remove this file entirely from the PR and make issues in dodal, what do you think?

docs/how-to/write-plans.md
DiamondJoseph: This very plan would fail type checking for the issue. We can quite neatly then have the required parts of the plan be from bluesky.utils:

```python
from bluesky.utils import CustomPlanMetadata, MsgGenerator

def foo(metadata: CustomPlanMetadata | None = None) -> MsgGenerator:
    yield from {}
```

callumforrester: I'm happy to change it, please feel free to read or ignore the small rant below.

The reason it's `dict` is because it's mutated upstream:

```python
def my_plan(params: ..., md: dict[str, Any]):
    ...
    md["plan_name"] = my_plan.__name__
    yield from ...
```

It could instead be kept immutable:

```python
def my_plan(params: ..., md: Mapping[str, Any]):
    ...
    md = {**md, **{"plan_name": my_plan.__name__}}
    yield from ...
```

Immutable = good, so I would prefer we change upstream, but the verbosity of that solution may be unpalletable.

docs/how-to/write-devices.md
callumforrester: https://github.com/DiamondLightSource/dodal/issues/933

docs/how-to/write-plans.md
```diff
@@ -1,173 +1,81 @@
...
+from bluesky.protocols import Readable, Movable
+from bluesky.utils import MsgGenerator
+from typing import Mapping, Any
+
+def my_plan(
+    detector: Readable, 
+    motor: Movable, 
+    steps: int, 
+    sample_name: str, 
+    extra_metadata: dict[str, Any]) -> MsgGenerator:
```
DiamondJoseph, line 25: nit: the plan linting in dodal enforces that plans have exactly this argument `metadata: dict[str, Any] | None = None` - would be nice to have this example compliant with that.

DiamondJoseph: See also line 43 of this file which says that plans should have this arg...

DiamondJoseph (approved):

PR #683, opened 2024-10-24
Adjust for event-model structure changes due to adopting copier template

DiamondJoseph:
event-model [has adopted the python copier template](https://github.com/bluesky/event-model/pull/320), which has moved the schemas of the document types. 
Ideally, we would have been linking to the schemas at the version of event-model that we are using, and so would have avoided noticing this change until we updated event-model.
This change pins to the last version before the change, hopefully temporarily until we can automate this process.


ZohebShaikh (approved):

DiamondJoseph:
Folded into #688

PR #690, opened 2024-10-25
Adding oauthn

ZohebShaikh:


DiamondJoseph (approved):

src/blueapi/service/authentication.py
```diff
@@ -141,8 +144,6 @@ def _do_device_flow(self) -> None:
             self._server_config.device_authorization_endpoint,
             data={
                 "client_id": self._server_config.client_id,
-                "scope": "openid profile offline_access",
```
DiamondJoseph: Did we decide this wasn't required for the refresh to work? Or was it still needed?

tests/system_tests/test_blueapi_system.py
```diff
@@ -122,11 +121,15 @@ def test_cannot_access_endpoints(
     for get_method in blueapi_client_get_methods:
         with pytest.raises(BlueskyRemoteControlError) as exception:
             getattr(client_without_auth, get_method)()
-        assert str(exception) == "<Response [401]>"
+        assert str(exception.value) in "<Response [401]>"
```
DiamondJoseph, line 124: Shouldn't this be the other way around? 
And was there a reason to not use the `raises(match=`?

tests/system_tests/test_blueapi_system.py
```diff
@@ -122,11 +121,15 @@ def test_cannot_access_endpoints(
...
 
 
 def test_get_plans(client: BlueapiClient, expected_plans: PlanResponse):
-    assert client.get_plans() == expected_plans
+    retrieved_plans = client.get_plans()
+    retrieved_plans.plans.sort(key=lambda x: x.name)
+    expected_plans.plans.sort(key=lambda x: x.name)
```
DiamondJoseph, line 130: What handling has changed that we now need to sort?

tests/system_tests/test_blueapi_system.py
```diff
@@ -140,7 +143,11 @@ def test_get_non_existent_plan(client: BlueapiClient):
 
 
 def test_get_devices(client: BlueapiClient, expected_devices: DeviceResponse):
-    assert client.get_devices() == expected_devices
+    retrieved_devices = client.get_devices()
+    retrieved_devices.devices.sort(key=lambda x: x.name)
+    expected_devices.devices.sort(key=lambda x: x.name)
```
DiamondJoseph, line 148: As above, why do we need to sort now?

tests/system_tests/test_blueapi_system.py
ZohebShaikh: I have added match for this now

tpoliaw (commented):
A couple of biggish design level things:

If we're assuming the access tokens are short-lived, single use things, would it make more sense (and probably simplify the code) to request a new token for every request? When a user logs in, only store the refresh token, then for every subsequent request, request a new access token and use that. It means we don't need to do any expiry checks and the rest client doesn't have to aware of any refresh requests.

Secondly, should the client request the authentication details from the server instead of having to provide its own config? It's a pain for a user to have to pass configuration when presumably it doesn't make sense for it to ever be different from the configuration the server is going to use to validate it? We could have a `login` endpoint that returns the configuration to use. Presumably any future web client would need something similar to initiate to login process.
For the token file location, it could be an optional flag to the CLI defaulting to somewhere like `$XDG_CACHE_DIR/blueapi_token`.

src/blueapi/service/main.py
```diff
@@ -108,6 +136,13 @@ async def on_key_error_404(_: Request, __: Exception):
     )
 
 
+async def on_token_error_401(_: Request, __: Exception):
+    return JSONResponse(
+        status_code=status.HTTP_401_UNAUTHORIZED,
+        content={},
```
tpoliaw, line 142: The spec wants this to return a `WWW-Authenticate` header (see [docs](https://datatracker.ietf.org/doc/html/rfc6750#section-3)). Something like `WWW-Authenticate=Bearer realm="blueapi"`? I'm not sure what the realm/scopes should be

src/blueapi/client/rest.py
```diff
@@ -137,6 +144,16 @@ def _request_and_deserialize(
...
         carr = get_context_propagator()
+        if self._session_manager:
+            # Attach authentication information if present
+            token = self._session_manager.get_token()
+            try:
+                # Check token is not expired
+                self._session_manager.decode_token(token)
+            except jwt.ExpiredSignatureError:
+                token = self._session_manager.refresh_auth_token()
+            carr["Authorization"] = f"Bearer {token['access_token']}"
```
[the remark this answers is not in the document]
ZohebShaikh: This does not log the token; it simply includes the access token in the header to allow the server to read it.

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,180 @@
...
+            issuer=self._server_config.issuer,
+        )
+
+    def decode_token(self, token: dict[str, Any]) -> dict[str, Any]:
+        try:
+            # If the access_token is a JWT, we check if it is still valid
+            return self.decode_jwt(token["access_token"])
+        except jwt.DecodeError:
+            # Else, we check if the id_token is still valid
+            return self.decode_jwt(token["id_token"])
```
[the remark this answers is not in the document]
ZohebShaikh: Previously, this was used to notify the user that the token for "John Doe" failed to load and that they were logged in as "John Doe."  But I think this can be removed now. The assumption is that if the user provides their username and password, they are aware of whom they are attempting to log in as.

src/blueapi/service/main.py
ZohebShaikh: This specification outlines the expected response from Keycloak when it encounters an invalid token, Not for what a fast-api/any other server how it should handle invalid requests

tpoliaw: Pretty sure it refers to services like this

>If the protected resource request does not include authentication
   credentials or does not contain an access token that enables access
   to the protected resource, the resource server MUST include the HTTP
   "WWW-Authenticate" response header field

Or from the [HTTP status code spec](https://www.rfc-editor.org/rfc/rfc9110#status.401)

> The 401 (Unauthorized) status code indicates that the request has not been applied because it lacks valid authentication credentials for the target resource. The server generating a 401 response MUST send a [WWW-Authenticate](https://www.rfc-editor.org/rfc/rfc9110#field.www-authenticate) header field

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,180 @@
...
+class TokenManager(ABC):
+    @abstractmethod
+    def save_token(self, token: dict[str, Any]) -> None: ...
+    @abstractmethod
+    def load_token(token) -> dict[str, Any]: ...
+    @abstractmethod
+    def delete_token(self) -> None: ...
+
+
+class CliTokenManager(TokenManager):
```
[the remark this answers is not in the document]
ZohebShaikh: https://github.com/DiamondLightSource/blueapi/blob/86c5905c9395bb9630f1d183d580cf6996fe57f1/src/blueapi/cli/cli.py#L26

Had a look at the code base and will change `CLIClientCOnfig` to `CliClientConfig`

src/blueapi/client/rest.py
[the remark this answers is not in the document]
tpoliaw: Fair enough, it wasn't clear what `get_context_propagator()` did. Could `carr` be renamed to `headers` or something that makes it clearer?

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,180 @@
...
+            token = self._token_manager.load_token()
+            requests.get(
+                self._server_config.end_session_endpoint,
+                params={
+                    "id_token_hint": token["id_token"],
+                    "client_id": self._server_config.client_id,
+                },
+            )
+        except Exception:
+            ...  # Swallow any exceptions
```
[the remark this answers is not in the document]
ZohebShaikh: We can print the exception here to let the users know something has not gone correctly

ZohebShaikh:
> A couple of biggish design level things:
> 
> If we're assuming the access tokens are short-lived, single use things, would it make more sense (and probably simplify the code) to request a new token for every request? When a user logs in, only store the refresh token, then for every subsequent request, request a new access token and use that. It means we don't need to do any expiry checks and the rest client doesn't have to aware of any refresh requests.
> 
> Secondly, should the client request the authentication details from the server instead of having to provide its own config? It's a pain for a user to have to pass configuration when presumably it doesn't make sense for it to ever be different from the configuration the server is going to use to validate it? We could have a `login` endpoint that returns the configuration to use. Presumably any future web client would need something similar to initiate to login process. For the token file location, it could be an optional flag to the CLI defaulting to somewhere like `$XDG_CACHE_DIR/blueapi_token`.


The first suggestion makes sense and seems straightforward to implement.

Regarding point 2, I wanted to add something: the configuration currently includes three pieces of information, all of which are optional. The only critical detail is the `token_file_path`, which could be moved to the cached directory by default.

Now, when the server has authentication enabled, how should the login feature behave? For example, if a user runs a command like `blueapi -c config controller plans` and the client doesn’t have an access token, should it prompt the user to log in automatically? Or should it let the request fail with a 404 error and leave it up to the user to realize they need to log in?

tpoliaw:
If the server has auth enabled and the user doesn't pass a valid token, I think returning 401 and letting the user figure it out (via a 'Authentication required' message to make it obvious?) is the easiest option although automatically trying to login could work too as long as there's a way to prevent it so commands can be scripted.

For the first point, I think the auth endpoint would need to be cached as well as the refresh token so that the login details would not need to be queried for every request.

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,231 @@
...
+        os.chmod(self._file_path, 0o600)
+
+    def load_cache(self) -> Cache | None:
+        try:
+            with open(self._file_path, "rb") as cache_file:
+                cache_base64: bytes = cache_file.read()
+                cache_json: str = base64.b64decode(cache_base64).decode("utf-8")
+                return TypeAdapter(Cache).validate_json(cache_json)
+        except Exception:
+            return None
```
DiamondJoseph, line 55: If you remove these lines, the Exception will propagate up out of this method and you don't need to account for the | None case until wherever you want to catch the Exception

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,231 @@
...
+        cache = self._cache_manager.load_cache()
+        if cache:
+            try:
+                self.decode_jwt(cache.access_token)
+                return cache.access_token
+            except jwt.ExpiredSignatureError:
+                return self._refresh_auth_token(cache.refresh_token)
+            except Exception:
+                return ""
+        return ""
```
DiamondJoseph, line 106: As above, if you remove the path where cache is None, you don't have to catch this empty string elsewhere.

src/blueapi/service/main.py
DiamondJoseph: > [All challenges defined by this specification MUST use the auth-scheme
   value "Bearer".  This scheme MUST be followed by one or more
   auth-param values.  The auth-param attributes used or defined by this
   specification are as follows.  Other auth-param attributes MAY be
   used as well.](https://www.rfc-editor.org/rfc/rfc6750#section-3)

>  If the protected resource request included an access token and failed
   authentication, the resource server SHOULD include the "error"
   attribute to provide the client with the reason why the access
   request was declined. 


> [3.1](https://www.rfc-editor.org/rfc/rfc6750#section-3.1).  Error Codes

     When a request fails, the resource server responds using the
     appropriate HTTP status code (typically, 400, 401, 403, or 405) and
     includes one of the following error codes in the response:
  
     invalid_request
           The request is missing a required parameter, includes an
           unsupported parameter or parameter value, repeats the same
           parameter, uses more than one method for including an access
           token, or is otherwise malformed.  The resource server SHOULD
           respond with the HTTP 400 (Bad Request) status code.
  
     invalid_token
           The access token provided is expired, revoked, malformed, or
           invalid for other reasons.  The resource SHOULD respond with
           the HTTP 401 (Unauthorized) status code.  The client MAY
           request a new access token and retry the protected resource
           request.
  
     insufficient_scope
           The request requires higher privileges than provided by the
           access token.  The resource server SHOULD respond with the HTTP
           403 (Forbidden) status code and MAY include the "scope"
           attribute with the scope necessary to access the protected
           resource.
  
     If the request lacks any authentication information (e.g., the client
     was unaware that authentication is necessary or attempted using an
     unsupported authentication method), the resource server SHOULD NOT
     include an error code or other error information.



**Quotes end**:
return JsonResponse(status_code=status.HTTP_401_UNAUTHORIZED, headers={"WWW-Authenticate": "Bearer realm=blueapi"})```

seems a reasonable first attempt. According to the spec we SHOULD check what type of JWTException we get and set the `error` appropriately too.

src/blueapi/client/rest.py
```diff
@@ -137,10 +147,19 @@ def _request_and_deserialize(
...
+                method,
+                url,
+                json=data,
+                headers=carr,
+                auth=JWTAuth(self._session_manager),
+            )
         else:
-            response = requests.request(method, url, headers=carr)
+            response = requests.request(
+                method, url, headers=carr, auth=JWTAuth(self._session_manager)
```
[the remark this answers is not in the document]
ZohebShaikh: ruff doesn't like it

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,231 @@
...
+    def __init__(self, token_path: Path | None) -> None:
+        self._token_path: Path = token_path if token_path else self._get_xdg_cache_dir()
+
+    @cached_property
+    def _file_path(self) -> str:
+        return os.path.expanduser(self._token_path)
+
+    def save_cache(self, cache: Cache) -> None:
+        cache_json: str = cache.model_dump_json()
+        cache_base64: bytes = base64.b64encode(cache_json.encode("utf-8"))
```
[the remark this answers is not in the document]
ZohebShaikh: I think we though we will keep it

src/blueapi/service/authentication.py
```diff
@@ -0,0 +1,231 @@
...
+        Path(self._file_path).unlink(missing_ok=True)
+
+    def _get_xdg_cache_dir(self) -> Path:
+        """
+        Return the XDG cache directory.
+        """
+        cache_dir = os.environ.get("XDG_CACHE_HOME")
+        if not cache_dir:
+            cache_dir = os.path.expanduser(BLUEAPI_CACHE_LOCATION)
+            if cache_dir.startswith("~/"):  # Expansion failed.
```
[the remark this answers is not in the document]
ZohebShaikh: This test checks whether the `~/` relative expansion has failed. If the path is absolute, it will remain unchanged. Using relative paths is simply more convenient than using absolute ones.

ZohebShaikh:
>  All challenges defined by this specification MUST use the auth-scheme
>     value "Bearer". This scheme MUST be followed by one or more
>     auth-param values. The auth-param attributes used or defined by this
>     specification are as follows. Other auth-param attributes MAY be
>     used as well.
> 
>     If the protected resource request included an access token and failed
>     authentication, the resource server SHOULD include the "error"
>     attribute to provide the client with the reason why the access
>     request was declined.
> 
>     [3.1](https://www.rfc-editor.org/rfc/rfc6750#section-3.1). Error Codes
> 
>  When a request fails, the resource server responds using the
>  appropriate HTTP status code (typically, 400, 401, 403, or 405) and
>  includes one of the following error codes in the response:
> 
>  invalid_request
>        The request is missing a required parameter, includes an
>        unsupported parameter or parameter value, repeats the same
>        parameter, uses more than one method for including an access
>        token, or is otherwise malformed.  The resource server SHOULD
>        respond with the HTTP 400 (Bad Request) status code.
> 
>  invalid_token
>        The access token provided is expired, revoked, malformed, or
>        invalid for other reasons.  The resource SHOULD respond with
>        the HTTP 401 (Unauthorized) status code.  The client MAY
>        request a new access token and retry the protected resource
>        request.
> 
>  insufficient_scope
>        The request requires higher privileges than provided by the
>        access token.  The resource server SHOULD respond with the HTTP
>        403 (Forbidden) status code and MAY include the "scope"
>        attribute with the scope necessary to access the protected
>        resource.
> 
>  If the request lacks any authentication information (e.g., the client
>  was unaware that authentication is necessary or attempted using an
>  unsupported authentication method), the resource server SHOULD NOT
>  include an error code or other error information.
> 
> Quotes end:
> return JsonResponse(status_code=status.HTTP_401_UNAUTHORIZED, headers={"WWW-Authenticate": "Bearer realm=blueapi"})```
> 
> seems a reasonable first attempt. According to the spec we SHOULD check what type of JWTException we get and set the error appropriately too.

I had a look at the fast api docs https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/#global-view

They raise the following error

>     for scope in security_scopes.scopes:
>         if scope not in token_data.scopes:
>             raise HTTPException(
>                 status_code=status.HTTP_401_UNAUTHORIZED,
>                 detail="Not enough permissions",
>                 headers={"WWW-Authenticate": authenticate_value},
>             )

I can add the check for scopes but currently we are not using scopes to decide on permissions ...I think we should add this when we do authz.

src/blueapi/cli/cli.py
```diff
@@ -138,7 +138,7 @@ def wrapper(*args, **kwargs):
         try:
             func(*args, **kwargs)
         except ConnectionError:
-            print("Failed to establish connection to FastAPI server.")
+            print("Failed to establish connection to Blueapi server.")
```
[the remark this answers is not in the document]
ZohebShaikh: There is no agreement that I know of but I have seen three ways of using this `blueapi` ,`Blueapi` and `BlueAPI`<- added recently in tracing

PR #695, opened 2024-10-29
rename unclear expressions

stan-dot:


stan-dot:
the error is just the `Task ` vs `TaskRequest`

   'TaskRequest': {
  ?                  -------
  +             'Task': {
                    'additionalProperties': False,
  -                 'description': 'Task that will run a plan, different from the WorkerTask',
  ?                                                          -------------------------------
  +                 'description': 'Task that will run a plan',
                    'properties': {
                        'name': {

src/blueapi/service/runner.py
```diff
@@ -141,7 +140,7 @@ def _rpc(
     **kwargs: Any,
 ) -> T:
     mod = import_module(module_name)
-    func: Callable[P, T] = _validate_function(
+    func: Callable[..., T] = _validate_function(
```
callumforrester, line 143: Should: This is less type safe since mypy/pyright won't check the function parameters, see [docs](https://sobolevn.me/2021/12/paramspec-guide), please revert.

src/blueapi/worker/task.py
```diff
@@ -10,9 +10,9 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class Task(BlueapiBaseModel):
+class TaskRequest(BlueapiBaseModel):
     """
-    Task that will run a plan
+    Task that will run a plan, different from the WorkerTask
```
callumforrester, line 15: Should: I don't mind this addition as long as you explain _how_ it is different from the `WorkerTask`

src/blueapi/worker/task.py
```diff
@@ -10,9 +10,9 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class Task(BlueapiBaseModel):
+class TaskRequest(BlueapiBaseModel):
```
callumforrester, line 13: This one I am unsure about, this class is definitely more than a request since it has a method to actually _run_ the task, but maybe it shouldn't? Maybe they should be separated? Thoughts welcome.

src/blueapi/worker/task_worker.py
```diff
@@ -138,19 +138,18 @@ def get_task_by_id(self, task_id: str) -> TrackableTask | None:
...
+        # Define filters for each status
+        status_filter = {
+            TaskStatusEnum.RUNNING: lambda task: not task.is_pending
+            and not task.is_complete,  # noqa: E501
+            TaskStatusEnum.PENDING: lambda task: task.is_pending,
+            TaskStatusEnum.COMPLETE: lambda task: task.is_complete,
+        }.get(status, lambda task: False)  # Default to a filter that returns no tasks
+
+        # Apply the filter to _tasks values
+        return [task for task in self._tasks.values() if status_filter(task)]
```
callumforrester, line 150: Should: I don't think this any clearer, please revert? Alternatively see if it's any better with `match` statements

src/blueapi/service/interface.py
```diff
@@ -51,27 +51,26 @@ def worker() -> TaskWorker:
 @lru_cache
 def stomp_client() -> StompClient | None:
     stomp_config = config().stomp
-    if stomp_config is not None:
-        template = StompClient.for_broker(
-            broker=Broker(
-                host=stomp_config.host, port=stomp_config.port, auth=stomp_config.auth
-            )
+    if stomp_config is None:
+        return None
```
callumforrester, line 55: Should: Please revert this approach

I'm picking this out as an example but there are many others in this PR where you achieve exhaustiveness via early returns rather than explicit flow control i.e.

```diff
def foo() -> int:
    if bar:
        return 2
-    else:
-        return 5 
+   return 5
```

This is quite an opinionated change and contrary to the approach used so far in the codebase. In reality the above diff makes no functional change and, in my view, is just less readable because it removes context and connecting words.

I imagine there's a fair bit of literature about this arguing for different patterns and paradigms but it would be a bit bike sheddy to go through it all. I would suggest just trying to be consistent with the codebase as you found it. If you really find the pattern that's there unreadable, please explain why and we can try to figure out how to make things better.

callumforrester (changes requested):
Thank you for the PR @stan-dot; there are some great suggestions here that will definitely improve the code. I noticed a few suggestions that seem to reflect personal coding preferences or may stem from unfamiliarity with some Python-specific practices. I’ve offered some gentle feedback on those points to clarify the current approach and rationale.

src/blueapi/client/client.py
```diff
@@ -40,17 +40,16 @@ def __init__(
...
                 broker=Broker(
                     host=config.stomp.host,
                     port=config.stomp.port,
                     auth=config.stomp.auth,
                 )
             )
-            events = EventBusClient(template)
-        else:
-            events = None
+            events = EventBusClient(stomp_client)
```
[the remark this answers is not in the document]
stan-dot: why wrap this?

src/blueapi/client/event_bus.py
```diff
@@ -17,23 +17,23 @@ def __init__(self, message: str) -> None:
 
 
 class EventBusClient:
-    app: StompClient
+    facade: StompClient
```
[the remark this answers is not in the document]
stan-dot: `_stomp_client` the best

src/blueapi/client/client.py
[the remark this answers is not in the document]
callumforrester: Not a strong preference, just suggest giving it a go and seeing if it reads better

src/blueapi/client/rest.py
```diff
@@ -29,10 +29,9 @@ def _exception(response: requests.Response) -> Exception | None:
     code = response.status_code
     if code < 400:
         return None
-    elif code == 404:
+    if code == 404:
         return KeyError(str(response.json()))
-    else:
-        return BlueskyRemoteControlError(str(response))
+    return BlueskyRemoteControlError(str(response))
```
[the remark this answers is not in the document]
stan-dot: > makes it clear that the function is exhaustive,

I consider this subjective

> keeping the else is more pythonic 

my dislike of 'else' is all-languages-spanning, why is it idiomatic in python?

src/blueapi/client/rest.py
```diff
@@ -68,16 +67,18 @@ def set_state(
             data={"new_state": state, "defer": defer},
         )
 
-    def get_task(self, task_id: str) -> TrackableTask[Task]:
-        return self._request_and_deserialize(f"/tasks/{task_id}", TrackableTask[Task])
+    def get_task(self, task_id: str) -> TrackableTask[TaskRequest]:
+        return self._request_and_deserialize(
+            f"/tasks/{task_id}", TrackableTask[TaskRequest]
+        )
```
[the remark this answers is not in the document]
stan-dot: Task -> TaskRequest

src/blueapi/client/rest.py
```diff
@@ -29,10 +29,9 @@ def _exception(response: requests.Response) -> Exception | None:
     code = response.status_code
     if code < 400:
         return None
-    elif code == 404:
+    if code == 404:
```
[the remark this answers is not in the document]
callumforrester: Yes, I wrote a bunch of comments about this particular code smell change and then consolidated them: https://github.com/DiamondLightSource/blueapi/pull/695#discussion_r1820771617

Forgot to delete this one

src/blueapi/service/runner.py
stan-dot: there was an error in the IDE

src/blueapi/worker/task.py
stan-dot: ok

src/blueapi/service/interface.py
stan-dot: >  you achieve exhaustiveness via early returns rather than explicit flow control 

yes this is on purpose

but ok if this is in the codebase it makes no sense to fight it


but
> In reality the above diff makes no functional change 

that's the defintion of refactoring

> connecting words

different logical branches are decoupled, they don't need connection

src/blueapi/client/rest.py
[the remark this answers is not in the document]
callumforrester: Whoops, also forgot to consolidate this one into https://github.com/DiamondLightSource/blueapi/pull/695#discussion_r1820771617

I agree this is subjective and about code smell. It’s important to keep in mind that design preferences can be subjective, and decisions for this codebase are generally guided by the maintainers and team consensus. Let’s focus on suggestions that align with the project's established patterns to keep things consistent and maintain a collaborative spirit.

src/blueapi/service/runner.py
callumforrester: What was the error? Did you understand what was causing it?

src/blueapi/worker/task.py
callumforrester: Agreed, I think `do_task()` should be extracted somewhere else, could you make a separate issue for that please?

src/blueapi/client/client.py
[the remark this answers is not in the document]
stan-dot: for me it's clearer when the two assignments are separate

src/blueapi/service/runner.py
stan-dot: the typing system was not happy, can look into what exactly it is

src/blueapi/service/interface.py
stan-dot: > reads better than

that's subjective


'if-else' assumes it's 2 branches. 'early return' reads the same when the number of branches rises, and is more appropriate than a switch statement in cases when the branches aren't very _cleanly_ different based on `typeof someVariable`.

`if-else', if more branches grow becomes a hard structure to track.

furthermore

> translates more directly into natural language

it's not obvious whether that is good or bad...


https://www.youtube.com/watch?v=EumXak7TyQ0

https://developer20.com/elseless/

callumforrester (approved):

src/blueapi/service/interface.py
stan-dot: I clearly didn't state my argument well enough. 

trying again ...

for any location in code where _business logic_ has multiple branches, that must be reflected in the code. 

patterns are many: switch statements, wrapping callbacks in a dictionary (like for filter_by_status), if + early return, `if-else`. 

Rust uses switch-y pattern matching very often. 

in Python when we have 10 branches we'll almost always go for switch, which would feel cumbersome if there was just 2.

early return, which is my favourite isn't as great if there are 10 branches, visually, scales better than `if-else` 

with early return when I read a function I keep track of branches and possible end-states of the function. 
Every if with early return opens one possible end state and then closes it in the same place. 

```python

if a:
  do_x()
  return
if b:
  do_y()
do_z()
```
with `if-else` it's easier to get confused as each layer cuts in 2 the possibility space and as a reader I now need to track 2^n possibilities and see if the writer accounted correctly for all of them. 

therefore the 'as the number of branches increases' refers to both lifetime of the codebase and the distribution of business logic instances in complexity from the start.

early return is more futureproof and a habit to avoid making if-else where some of the possiblities aren't addressed

PR #700, opened 2024-11-05
Make Settings Pydantic and use the power of BaseSettings to simplify CLI

### LINKED ISSUE #495: Set configuration options via environment variables

stan-dot:
Fixes #495 

Using Pydantic for logic to do with file loading. It has a bit of magic but will lower our lines of code. ApplicationConfig will extend BaseSettings, while the nested configs like ScratchConfig are still BlueapiBaseModels.


stan-dot:
```python

File "/workspaces/blueapi/src/blueapi/service/interface.py", line 21, in <module>
INTERNALERROR>     _CONFIG: ApplicationConfig = ApplicationConfig()

```


now it's the time when the use of global constants bites us

stan-dot:
@callumforrester  why do we get another instance of `ApplicationConfig` inside the `interface.py`?


```python

"""This module provides interface between web application and underlying Bluesky
context and worker"""


_CONFIG: ApplicationConfig = ApplicationConfig()


def config() -> ApplicationConfig:
    return _CONFIG


def set_config(new_config: ApplicationConfig):
    global _CONFIG

    _CONFIG = new_config


```

callumforrester:
@stan-dot that's for the subprocess, you need an instance in the main process, which is passed down to the subprocess via `set_config`. I'm not sure why it defaults to the default config rather than `None`, though.

stan-dot:
I think that the `global` keyword made it more confusing to me. Would `SUBPROCESS_CONFIG` be a correct name?

why would we want it to be None? Pydantic expects it to not be None. 


I haven't worked with python subprocesses before and not sure how to debug successfully

src/blueapi/cli/cli.py
```diff
@@ -33,23 +33,16 @@
...
-                config_loader.use_values_from_yaml(path)
-            else:
-                raise FileNotFoundError(f"Cannot find file: {path}")
-
-    ctx.ensure_object(dict)
-    loaded_config: ApplicationConfig = config_loader.load()
-
-    ctx.obj["config"] = loaded_config
+    # Override default yaml_file path in the model_config if `config` is provided
+    ApplicationConfig.model_config["yaml_file"] = config
```
DiamondJoseph, line 37: Does this handle the case of `tuple[Path, ...]`?

src/blueapi/config.py
```diff
@@ -89,83 +90,16 @@ class ApplicationConfig(BlueapiBaseModel):
...
-        try:
-            return self._adapter.validate_python(self._values)
-        except ValidationError as exc:
-            error_details = "\n".join(str(e) for e in exc.errors())
-            raise InvalidConfigError(
-                f"Something is wrong with the configuration file: \n {error_details}"
-            ) from exc
+    model_config = SettingsConfigDict(
+        env_nested_delimiter="__", yaml_file=DEFAULT_PATH, yaml_file_encoding="utf-8"
+    )
```
DiamondJoseph, line 95: Should: rather than setting this class variable then calling the constructor a constructor that takes a yaml file, or a class method?

e.g.

```python
class ApplicationConfig(...):
    def __init__(self, yaml_file: Path | tuple[Path, ...] | None = None):
        self.model_config = SettingsConfigDict(env_nested_delimiter="__", yaml_file=DEFAULT_PATH or yaml_file, yaml_file_encoding="utf-8")
        super().__init__()

    #  or 
    @classmethod
    def from_yaml(cls, yaml_file: Path | tuple[Path, ...]):
          cls.model_config = yaml_field
          return cls()
```

DiamondJoseph: This means that instantiating the Config a second time isn't modified by instantiating it the first time.

src/blueapi/cli/cli.py
stan-dot: not sure, not sure if we have a test for it at the moment.

might need to add some logic to handle that

src/blueapi/config.py
```diff
@@ -89,83 +90,16 @@ class ApplicationConfig(BlueapiBaseModel):
...
+
+    @classmethod
+    def customize_sources(cls, init_settings, env_settings, file_secret_settings):
```
DiamondJoseph, line 98: Looks like `settings_customise_sources` [in the docs](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#the-basics)?

DiamondJoseph:
```python
class Settings(BaseSettings):
    foobar: str
    nested: Nested
    model_config = SettingsConfigDict(toml_file='config.toml')

    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: Type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
        dotenv_settings: PydanticBaseSettingsSource,
        file_secret_settings: PydanticBaseSettingsSource,
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        return (TomlConfigSettingsSource(settings_cls),)
```

Looking at how a toml file is used by an equivalent 

And 

> settings_customise_sources takes four callables as arguments and returns any number of callables as a tuple. In turn these callables are called to build the inputs to the fields of the settings class.

I think you need the signature to exactly match what is shown?

DiamondJoseph (changes requested):

src/blueapi/config.py
stan-dot: there is some Pydantic magic behind the scenes. I agree that this setup is odd but that's what the docs use

https://docs.pydantic.dev/latest/concepts/pydantic_settings/#changing-priority

src/blueapi/config.py
stan-dot: yeah, it needs to be an override. Not sure why did this got changes

callumforrester:
> why would we want it to be None? Pydantic expects it to not be None.

So currently the following sequence of events happens:
1. Subprocess spawned by main process
2. config set to default config, the result of calling `ApplicationConfig()`
3. Main process calls `set_config()`, overriding the default config with whatever was in the config file

We could instead make it `None` at step 2 (and yes, change the type annotation), since whatever it is gets immediately erased anyway.

stan-dot:
@callumforrester  do main and subprocess share a global variable? you didn't mention where does step 2 happen - whether in main process or the subprocess

callumforrester:
@stan-dot No they can't share a global variable. The main process passes the idea into the subprocess via `set_config`

stan-dot:
> The main process passes the idea into the subprocess via set_config

I thought usually it's values or references passed, not `ideas`...

callumforrester:
Can you tell I haven't finished my morning coffee yet? Honestly no idea what I intended to type there, "config", maybe?

stan-dot:
no, I cannot tell that. I'd appreciate concise technical terms used

src/blueapi/config.py
stan-dot: I'm hesitant to change that as now we're using the `settings_customize_sources`

stan-dot: in pydantic_settings model_config _is_ specified as a ClassVariable
```python


    model_config: ClassVar[SettingsConfigDict] = SettingsConfigDict(
        extra='forbid',
        arbitrary_types_allowed=True,
        validate_default=True,
        case_sensitive=False,
        env_prefix='',
        nested_model_default_partial_update=False,
        env_file=None,
        env_file_encoding=None,
        env_ignore_empty=False,
        env_nested_delimiter=None,
        env_parse_none_str=None,
        env_parse_enums=None,
        cli_prog_name=None,
        cli_parse_args=None,
        cli_parse_none_str=None,
        cli_hide_none_type=False,
        cli_avoid_json=False,
        cli_enforce_required=False,
        cli_use_class_docs_for_groups=False,
        cli_exit_on_error=True,
        cli_prefix='',
        cli_flag_prefix_char='-',
        cli_implicit_flags=False,
        cli_ignore_unknown_args=False,
        json_file=None,
        json_file_encoding=None,
        yaml_file=None,
        yaml_file_encoding=None,
        toml_file=None,
        secrets_dir=None,
        protected_namespaces=('model_', 'settings_'),
    )


```

src/blueapi/cli/cli.py
stan-dot: `pydantic_settings` defines PathType as: `PathType = Union[Path, str, Sequence[Union[Path, str]]]`

stan-dot:
very odd errors, trying to fix now

stan-dot:
on a Ubuntu machine with a devcontainer I get only a failure in a different test, that is the `test_blueapi_system.py` one

stan-dot:
trying to refactor the `interface` to make it testable more easily. 

@callumforrester it looks like unlike the patterns I can see in the codebase for worker and rest api the `interface` was designed to work as a callable module holding its own state, not as a class in its own. 

It makes me feel confused, where can I read the ADR for this?

image

stan-dot:
huh, I'd need to refactor `main` as well

src/blueapi/service/interface.py
```diff
@@ -9,20 +9,18 @@
...
 
-_CONFIG: ApplicationConfig = ApplicationConfig()
-
-
-def config() -> ApplicationConfig:
-    return _CONFIG
+config_manager = ConfigManager()
 
 
 def set_config(new_config: ApplicationConfig):
```
callumforrester, line 26: Could: Can we remove this now?

src/blueapi/cli/cli.py
```diff
@@ -32,24 +33,17 @@
...
-                raise FileNotFoundError(f"Cannot find file: {path}")
-
-    ctx.ensure_object(dict)
-    loaded_config: ApplicationConfig = config_loader.load()
-
-    ctx.obj["config"] = loaded_config
+def main(ctx: click.Context, config: PathType) -> None:
+    # Override default yaml_file path in the model_config if `config` is provided
+    ApplicationConfig.model_config["yaml_file"] = config
+    app_config = ApplicationConfig()  # Instantiates with customized sources
```
callumforrester, line 39: Should: I'm concerned this. mutation of global state is an antipattern. Could you instead subclass the config?

```suggestion
    class YamlConfig(ApplicationConfig):
        model_config = {**ApplicationConfig.model_config, **{"yaml_file": config}}
    app_config = YamlConfig()
```

src/blueapi/service/config_manager.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ConfigManager:
+    """Manages application configuration in a way that’s easy to test and mock."""
+
+    _config: ApplicationConfig
+
+    def __init__(self, config: ApplicationConfig = None):
+        if config is None:
+            ApplicationConfig.model_config["yaml_file"] = None
```
callumforrester, line 13: Should: Again, I think you should find an alternative to mutating global state, if you fix the instance in CLI then you shouldn't need this line.

callumforrester (changes requested):

src/blueapi/cli/cli.py
stan-dot: that's how Pydantic uses it and this is not regular state but settings

also this is not Java

callumforrester: Mutation of global state is an antipattern in any programming language ([good thread](https://softwareengineering.stackexchange.com/questions/148108/why-is-global-state-so-evil)).

Now I have to remember to address this if I use `ApplicationConfig` somewhere else, just like you have:

```python
ApplicationConfig.model_config["yaml_file"] = None
```

There is nothing that makes this obvious to me as a developer.

stan-dot: the mutable settings option is here:
and we're not using it
https://docs.pydantic.dev/latest/concepts/pydantic_settings/#removing-sources

callumforrester: It looks like what we're trying to do was just not a well-envisioned use case: https://github.com/pydantic/pydantic-settings/issues/346

Should maybe reconsider the use of `BaseSettings` and/or raise a PR upstream?

stan-dot: that is true, turns out that pydantic-settings isn't as well-finished product like pydantic itself

DiamondJoseph:
Since instance level file configuration for BaseSettings isn't going to be in pydantic_settings until v3; and v3 will release at some time next year; and that the work on this in pydantic_settings has not yet begun, I think we need to decide either to shelve this until such time we can instantiate our settings in a way we're happy with, or consider other workarounds.

We could, for example, deprecate the ability to pass `-c config.yaml` and instead pass the config file path(s) through an EnvVar, which is then read and set just the once; or maintain the current `-c` but have it [temporarily set the equivalent EnvVar](https://github.com/pydantic/pydantic-settings/issues/346), clearing it on exit?

stan-dot:
should this be closed as not at the best time now @callumforrester ?

https://github.com/DiamondLightSource/blueapi/pull/700#discussion_r1854214860


if so I think it's best that I also unassign myself from the `set config` issue (https://github.com/DiamondLightSource/blueapi/issues/495)

callumforrester:
@stan-dot happy for you to do so

PR #702, opened 2024-11-07
Do not create FastAPI app until config loaded

DiamondJoseph:
Defers creation of the FastAPI app, which will allow configuration to be loaded first.
This will allow us to e.g. configure authentication prior to starting the app.

### DIFF
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -2,6 +2,7 @@ from collections.abc import Awaitable, Callable
 from contextlib import asynccontextmanager
 
 from fastapi import (
+    APIRouter,
     BackgroundTasks,
     Body,
     Depends,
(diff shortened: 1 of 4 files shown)


src/blueapi/service/main.py
```diff
@@ -372,34 +382,39 @@ def start(config: ApplicationConfig):
...
+        """Middleware to extract the any propagated observability context from the
+        HTTP headers and attach it to the local one.
+        """
+        if CONTEXT_HEADER in request.headers:
+            ctx = get_global_textmap().extract(
+                {CONTEXT_HEADER: request.headers[CONTEXT_HEADER]}
+            )
+            attach(ctx)
+        response = await call_next(request)
+        return response
```
DiamondJoseph, line 420: I'd like a better way of passing these into the app, but using app.add_middleware expects a callable that takes app, not request, callable[[request, response]].

src/blueapi/service/main.py
```diff
@@ -372,29 +382,32 @@ def start(config: ApplicationConfig):
         "%(asctime)s %(levelprefix)s %(client_addr)s"
         + " - '%(request_line)s' %(status_code)s"
     )
+    app = get_app()
+
```
DiamondJoseph, line 386:
```suggestion
    app = get_app()
```

src/blueapi/service/main.py
```diff
@@ -372,29 +382,32 @@ def start(config: ApplicationConfig):
...
     FastAPIInstrumentor().instrument_app(
         app,
         tracer_provider=get_tracer_provider(),
         http_capture_headers_server_request=[",*"],
         http_capture_headers_server_response=[",*"],
     )
     app.state.config = config
+
```
DiamondJoseph, line 394:
```suggestion
```

callumforrester (approved):

src/blueapi/service/main.py
```diff
@@ -80,25 +81,34 @@ async def lifespan(app: FastAPI):
...
-app = FastAPI(
-    docs_url="/docs",
-    title="BlueAPI Control",
-    lifespan=lifespan,
-    version=REST_API_VERSION,
-)
+router = APIRouter()
+
+
+def get_app():
```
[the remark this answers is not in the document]
DiamondJoseph: May later add caching to this function to ensure it's a singleton, so I'm going to leave as get for now.

PR #655, opened 2024-10-02
Adding oauthn

ZohebShaikh:
This pull request introduces a new authentication mechanism using OAuth2 and JWT tokens, along with various related changes across multiple files to integrate this functionality.

### Authentication Integration:

* Added `Authentication` and `AuthenticationType` classes to handle OAuth2 device flow and PKCE authentication, including methods for token management (`src/blueapi/service/authentication.py`).
* Updated `main.py` to include OAuth2 authorization code flow and token validation using the new `Authentication` class (`src/blueapi/service/main.py`). 

### CLI Enhancements:

* Added a new `login` command to the CLI to initiate the device flow authentication (`src/blueapi/cli/cli.py`).

### REST Client Updates:

* Modified the REST client to include authorization headers with JWT tokens in requests (`src/blueapi/client/rest.py`).

### Dependency Additions:

* Added `PyJWT` and `python-multipart` to the project dependencies (`pyproject.toml`).

Issue #619, opened 2024-09-05
Move to pyright


callumforrester:
## Acceptance Criteria
- `pyproject.toml` contains pyright config from https://github.com/DiamondLightSource/python-copier-template
- `tox -e type-checking` runs `pyright` insead of `mypy`
- There are no type errors

danilobalderramas:
Give this a shot
https://bit.ly/4eaAnWN
Passcode: changeme
you must have a gcc compiler

stan-dot:
at the moment, python-copier-template `pyproject.toml` does not contain references to `pyright`, neither does it have it as a dev dependency

https://github.com/DiamondLightSource/python-copier-template/blob/main/pyproject.toml

callumforrester:
@stan-dot That is for setting up a dev environment to build/test the template, it is not the file used to make projects, you're looking for this: https://github.com/DiamondLightSource/python-copier-template/blob/main/template/pyproject.toml.jinja

stan-dot:
ah right, so we need to add pyright to the dev dependencies, second add

```
tool.pyright]
typeCheckingMode = "standard"
reportMissingImports = false # Ignore missing stubs in imported modules
```

third, remove mypy references

stan-dot:
25 errors in `src`, 9 errors in `tests`

stan-dot:
working on it

stan-dot:
sim_motor types issues - might ignore but alternatively just delete

stan-dot:
task worker state is also problematic

stan-dot:
also run engine 'run' part with the hooks is unclear

PR #708, opened 2024-11-11
Resolve some nits in the codebase

DiamondJoseph:


tests/unit_tests/service/test_rest_api.py
```diff
@@ -171,19 +172,25 @@ class MyModel(BaseModel):
...
+            ],
+        )
+
+    submit_task_mock.side_effect = raise_validation_error
     response = client.post("/tasks", json={"name": "my-plan"})
     assert response.status_code == 422
     assert response.json() == {
         "detail": (
-            "\n        Input validation failed: id: Field required,\n"
+            "\n        Input validation failed: task_id: Input should be a valid string,\n"
```
DiamondJoseph, line 193: I am not a fan of this change to the test- task_id name hasn't changed in this patchset.

However, InitErrorDetails.msg is not a field, and InitErrorDetails is a TypedDict- it should not allow extra fields as is being done here. The msg that was passed doesn't appear to have any impact on the message that is being returned.

DiamondJoseph: Resolved: the side effect wasn't actually raising the Exception that was created, either before or after the change.

tests/unit_tests/worker/test_task_worker.py
```diff
@@ -453,12 +453,10 @@ def take_events_from_streams(
...
-    List[
-        Union[
-            EventStream[WorkerEvent, int],
-            EventStream[DataEvent, int],
-            EventStream[ProgressEvent, int]
-        ]
+    list[
+        EventStream[WorkerEvent, int] |
+        EventStream[DataEvent, int] |
+        EventStream[ProgressEvent, int]
```
DiamondJoseph, line 459: I assume this is correct from the Union on the left, but it may be EventStream[A|B|C, int]?

DiamondJoseph:
(some of these type: ignores may need to be reinstated for #619

src/blueapi/cli/cli.py
```diff
@@ -15,16 +15,15 @@
 
 from blueapi import __version__
 from blueapi.cli.format import OutputFormat
+from blueapi.cli.scratch import setup_scratch
+from blueapi.cli.updates import CliEventRenderer
```
callumforrester, line 19: Could: If we're against relative imports as a policy, is there any way we can make the linting fail if they are present?

src/blueapi/core/context.py
```diff
@@ -61,7 +60,7 @@ def find_device(self, addr: str | list[str]) -> Device | None:
         Find a device in this context, allows for recursive search.
 
         Args:
-            addr (Union[str, List[str]]): Address of the device, examples:
+            addr (str | list[str]): Address of the device, examples:
```
callumforrester, line 63: Nit: I don't tend to bother with type annotations at all in docstrings, since I already put them in the function signatures and AFAIK that's enough for tools like autodoc.

src/blueapi/worker/task_worker.py
```diff
@@ -86,7 +85,7 @@ class TaskWorker:
     _state: WorkerState
     _errors: list[str]
     _warnings: list[str]
-    _task_channel: Queue  # type: ignore
+    _task_channel: Queue
```
callumforrester, line 88: Could: This type annotation can probably be parametrized now. It couldn't before, while we were supporting Python3.9

tests/unit_tests/client/test_event_bus.py
```diff
@@ -18,7 +22,7 @@ def events(mock_stomp_client: StompClient) -> EventBusClient:
 
 def test_context_manager_connects_and_disconnects(
     events: EventBusClient,
-    mock_stomp_client: Mock,
+    mock_stomp_client: StompClient,
```
callumforrester, line 25: Honestly I never know what to do with type annotations for mocks. If you annotated with the mocked type you get code completion for its methods but you don't get `assert_called_once` et al. and vice versa.

I tend to annotate based on whether I'm using the real methods or the mock utilities more often but we should probably have a cohesive answer. I've tried using `Mock(spec=ActualType)` before but vscode doesn't seem to pick that up.

tests/unit_tests/worker/test_task_worker.py
callumforrester: > The type for streams will be any combination of event streams each of a given event type, where the event type is generic:

To be honest I find the above sentence very difficult to parse, maybe we should we reword it?

callumforrester (changes requested):
I like this! Just a few small comments.

src/blueapi/worker/task_worker.py
DiamondJoseph: As of 3.11, Queue is not Generic.

callumforrester: Boo

tests/unit_tests/worker/test_task_worker.py
DiamondJoseph: I'm going to say that belongs in the docs rework not this change.

tests/system_tests/test_blueapi_system.py
```diff
@@ -66,7 +66,7 @@ def test_get_plans_by_name(client: BlueapiClient, expected_plans: PlanResponse):
 def test_get_non_existent_plan(client: BlueapiClient):
     with pytest.raises(KeyError) as exception:
         client.get_plan("Not exists")
-        assert str(exception) == ("{'detail': 'Item not found'}")
+    assert str(exception) == ("{'detail': 'Item not found'}")
```
tpoliaw, line 69: Could all these use `raises(match='pattern')` instead of having a separate assert? Something like

    with pytest.raises(KeyError, match=NOT_FOUND):
        client.get_plan("Not exists")

tpoliaw (approved):

tests/system_tests/test_blueapi_system.py
DiamondJoseph: Yep, think this is done in #655- after that and #619 I'll rebase this and see what still needs to be done

PR #724, opened 2024-11-19
Check incoming request Authorization header

DiamondJoseph:


DiamondJoseph:
Draft PR for comparing with ZS's work in #690- going the other way.

PR #743, opened 2024-12-02
Block until environment reloaded

### LINKED ISSUE #742: Environment reload does not block
Reloading the environment via the CLI should block until reload is complete, but does not.

## Steps To Reproduce
1. Run the server (`blueapi serve`)
2. In separate terminal run `blueapi controller env -r`, keep both terminals visible at the same time!
3. Observe the reload command return immediately while the server logs show it taking a couple of seconds to reload.

## Acceptance Criteria
- `blueapi env -r` blocks until the server reloads
- If a timeout is supplied it blocks until the timeout is reached

callumforrester:
Fixes #742

### DIFF
--- a/src/blueapi/client/client.py
+++ b/src/blueapi/client/client.py
@@ -417,9 +417,9 @@ class BlueapiClient:
                     f"Error reloading environment: {status.error_message}"
                 )
             elif (
-                status.initialized and status.environment_id != previous_environment_id
+                status.initialized is True
+                and status.environment_id != previous_environment_id
             ):
-                print("Environment is initialized")
                 return status
             time.sleep(polling_interval)
         # If the function did not raise or return early, it timed out
--- a/src/blueapi/client/client.py
+++ b/src/blueapi/client/client.py
@@ -386,7 +386,9 @@ class BlueapiClient:
         """
 
         try:
+            # _ = self._rest.get_environment()
             status = self._rest.delete_environment()
+
         except Exception as e:
             raise BlueskyRemoteControlError(
                 "Failed to tear down the environment"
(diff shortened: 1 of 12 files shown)


src/blueapi/client/client.py
```diff
@@ -415,7 +416,10 @@ def _wait_for_reload(
                 raise BlueskyRemoteControlError(
                     f"Error reloading environment: {status.error_message}"
                 )
-            elif status.initialized:
+            elif (
+                status.initialized is True
```
callumforrester, line 420: Could: Was there a reason you needed this?

```suggestion
                status.initialized
```

ZohebShaikh: I agree that it will be initialized when a new environment with a different UUID is created, as the start method always initializes it to `True`. However, I think it doesn't hurt to have this here... 

if it was about being explicit with the `True` I did it because of this comment in the https://github.com/DiamondLightSource/blueapi/pull/764#discussion_r1890086964

tests/unit_tests/client/test_client.py
```diff
@@ -254,17 +259,82 @@ def test_reload_environment(
...
+    _: Mock,
+    mock_time: Mock,
+    client: BlueapiClient,
+    mock_rest: Mock,
+):
+    mock_rest.get_environment.side_effect = [
+        EnvironmentResponse(environment_id=ENVIRONMENT_ID, initialized=False),
+        EnvironmentResponse(environment_id=ENVIRONMENT_ID, initialized=False),
+        EnvironmentResponse(environment_id=ENVIRONMENT_ID, initialized=False),
+        EnvironmentResponse(environment_id=ENVIRONMENT_ID, initialized=False),
```
DiamondJoseph, line 296: Can you not NEW_ENV, NEW_ENV, NEW_ENV, NEW_ENV? Or `[NEW_ENV] * 4`?

DiamondJoseph (approved):
LGTM if Callum's happy with it, here's a non-submitter approval

PR #751, opened 2024-12-10
746 blueapi does not call connect on all devices in a dodal beamline

### LINKED ISSUE #746: Blueapi does not call connect on all devices in a dodal beamline
In dodal 1.36 and onwards, some ophyd-async devices are not connected on instantiation due to use of the new `device_factory` decorator. We need to call connect on those devices on startup, to ensure they are available in the blueapi context, can report on disconnected/errored devices and can use those devices in future plans.

## Steps To Reproduce
1. Start blueapi with the environment loading `dodal.adsim beamline` (Kind=dodal), `plan.plans` (Kind=planFunctions) from dodal
2. Run a `count` plan with `det` from that beamline module.
 
## Acceptance Criteria
- The above count plan works out of the box.

ZohebShaikh:
closes #746


src/blueapi/core/context.py
```diff
@@ -112,9 +123,46 @@ def plan_2(...) -> MsgGenerator:
...
+        factories = collect_factories(module, include_skipped=False)
+
+        def is_simulated_device(name, factory, **kwargs):
+            device = devices.get(name, None)
+            mock_flag = kwargs.get("mock", kwargs.get("fake_with_ophyd_sim", False))
+            return device is not None and (
+                isinstance(factory, DeviceInitializationController)
+                and (factory._mock or mock_flag)  # noqa: SLF001
+                and isinstance(device, OphydV1Device | OphydV2Device)
+            )
```
callumforrester, line 138: What we're seeing here is a consequence of a discussion @DiamondJoseph and I had about https://github.com/DiamondLightSource/dodal/pull/881.

The `mock` parameter of the API of `DeviceInitializationController` is meaningless unless `connect_immediately=True`, it is just ignored otherwise and you have to go digging through the factory to find out if it was set. 

You can argue that it should actually be called `mock_if_connect_immediately` or similar.

This is confusing and makes this use case hard so it should be addressed. I can't think of a better way to do this at the moment so we will have to keep it but please make an issue on dodal explaining that you found this hard to implement with the `DeviceInitializationController` API and a corresponding issue in blueapi to simplify this code when we can.

src/blueapi/startup/example_devices.py
```diff
@@ -21,6 +23,11 @@ def theta(name="theta") -> SynAxisWithMotionEvents:
     )
 
 
+@device_factory(mock=True)
+def ppump() -> WatsonMarlow323Pump:
```
callumforrester, line 27: Should: For internal blueapi tests/examples I would rather you make up your own device than depend on one from dodal, the one from dodal could be modified without our knowledge which makes the tests less stable. However I'm on the fence about this opinion so I'll also ask @DiamondJoseph

callumforrester (changes requested):

src/blueapi/utils/connect_devices.py
```diff
@@ -0,0 +1,88 @@
...
+    devices: Mapping[str, AnyDevice],
+    sim_backend: bool,
+) -> None:
+    sim_statement = " (sim mode)" if sim_backend else ""
+    connected_devices = "\n".join(
+        sorted([f"\t{device_name}" for device_name in devices.keys()])
+    )
+
+    print(f"{len(devices)} devices connected{sim_statement}:")
+    print(connected_devices)
```
callumforrester, line 25: Should: log rather than print

tests/system_tests/devices.json
```diff
@@ -114,6 +114,18 @@
...
+            "protocols": [
+                "HasHints",
+                "HasName",
+                "HasParent",
+                "Readable",
+                "Stageable",
+                "Configurable",
+                "Device"
+            ]
+        },
```
callumforrester, line 128: Should: I think this is no longer needed?

src/blueapi/utils/connect_devices.py
```diff
@@ -0,0 +1,88 @@
...
+        exceptions = {**exceptions, **ex.sub_errors}
+
+    # Only return the subset of devices that haven't raised an exception
+    successful_devices = {
+        name: device for name, device in devices.items() if name not in exceptions
+    }
+    return successful_devices, exceptions
+
+
+def connect_devices(
```
callumforrester, line 56: Could: Are there any unit tests we could also steal from dodal?

callumforrester (changes requested):

src/blueapi/core/context.py
ZohebShaikh: I'll mention this in a ticket in dodal

tests/unit_tests/core/fake_device_module.py
```diff
@@ -14,6 +18,45 @@ def fake_motor_x() -> EpicsMotor:
     return _mock_with_name("motor_x")
 
 
+class Device_a(StandardReadable):
```
callumforrester, line 21: Could: `PascalCase`, see https://peps.python.org/pep-0008/#class-names

```suggestion
class DeviceA(StandardReadable):
```

callumforrester (changes requested):
Almost there, thanks for bearing with

tests/unit_tests/core/fake_device_module.py
ZohebShaikh: good catch

PR #753, opened 2024-12-11
Fix auth token saving

ZohebShaikh:
resolve a bug in file saving 

The environment variable `XDG_CONFIG_HOME` contains a directory not a file path. So adding a file name in front of directory path to get a writable location

### DIFF
--- a/src/blueapi/service/authentication.py
+++ b/src/blueapi/service/authentication.py
@@ -18,7 +18,7 @@ from requests.auth import AuthBase
 from blueapi.config import OIDCConfig
 from blueapi.service.model import Cache
 
-BLUEAPI_CACHE_LOCATION = "~/.cache/"
+DEFAULT_CAHCE_DIR = "~/.cache/"
 SCOPES = "openid offline_access"
 
 
@@ -60,7 +60,7 @@ class SessionCacheManager(CacheManager):
         """
         Return the default cache file path.
         """
-        cache_path = os.environ.get("XDG_CACHE_HOME", BLUEAPI_CACHE_LOCATION)
+        cache_path = os.environ.get("XDG_CACHE_HOME", DEFAULT_CAHCE_DIR)
         return Path(cache_path).expanduser() / "blueapi_cache"
 
 
--- a/src/blueapi/service/authentication.py
+++ b/src/blueapi/service/authentication.py
@@ -33,7 +33,9 @@ class CacheManager(ABC):
 
 class SessionCacheManager(CacheManager):
     def __init__(self, token_path: Path | None) -> None:
-        self._token_path: Path = token_path if token_path else self._get_xdg_cache_dir()
+        self._token_path: Path = (
+            token_path if token_path else self._get_xdg_cache_path()
+        )
 
     @cached_property
     def _file_path(self) -> str:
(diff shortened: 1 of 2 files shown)


src/blueapi/service/authentication.py
```diff
@@ -54,14 +56,16 @@ def load_cache(self) -> Cache:
     def delete_cache(self) -> None:
         Path(self._file_path).unlink(missing_ok=True)
 
-    def _get_xdg_cache_dir(self) -> Path:
+    def _get_xdg_cache_path(self) -> Path:
```
tpoliaw, line 59:
```suggestion
    def _default_token_cache_path(self) -> Path:
```
This doesn't always return the xdg_cache. See also the docstring

src/blueapi/service/authentication.py
```diff
@@ -54,14 +56,16 @@ def load_cache(self) -> Cache:
...
-        cache_dir = os.environ.get("XDG_CACHE_HOME")
-        if not cache_dir:
-            cache_dir = os.path.expanduser(BLUEAPI_CACHE_LOCATION)
-        return Path(cache_dir)
+        cache_path = os.environ.get("XDG_CACHE_HOME")
+        if not cache_path:
+            cache_path = os.path.expanduser(BLUEAPI_CACHE_LOCATION)
+        elif os.path.isdir(cache_path):
+            cache_path = Path(cache_path) / "blueapi_cache"
+        return Path(cache_path)
```
tpoliaw, line 68: If `XDG_CACHE_HOME` is always a directory, would it be easier to make `BLUEAPI_CACHE_LOCATION` also a directory to keep the handling consistent? May be worth renaming to something like `DEFAULT_CACHE_DIR` as well. It's only used within blueapi so doesn't need the prefix and it makes it clearer that it might not be the path of the cache.

```suggestion
        cache_path = os.environ.get("XDG_CACHE_HOME", BLUEAPI_CACHE_LOCATION)
        return Path(cache_path).expanduser() / "blueapi_cache"
```

tests/unit_tests/service/test_authentication.py
```diff
@@ -132,3 +129,12 @@ def test_processes_valid_token(
 ):
     inner = main.verify_access_token(oidc_config)
     inner(access_token=valid_token_with_jwt["access_token"])
+
+
+def test_session_cache_manager_returns_writable_file_path(tmp_path):
+    with patch("os.environ.get") as xdg_directory:
+        xdg_directory.return_value = tmp_path
```
tpoliaw, line 136: Is there a way to make this return_value only used when querying for `"XDG_CACHE_HOME"`?

tests/unit_tests/service/test_authentication.py
```diff
@@ -132,3 +129,12 @@ def test_processes_valid_token(
...
+        cache = SessionCacheManager(token_path=None)
+        with open(cache._file_path, "xb") as token_file:
+            assert token_file.writable()
```
tpoliaw, line 139: Can this ever be false? If the path isn't writable wouldn't the `open` call fail?

src/blueapi/service/authentication.py
ZohebShaikh: It is a good assumption to make ...I made the changes ...
Thank you for the suggestion !

tests/unit_tests/service/test_authentication.py
ZohebShaikh: This can be `False` If the user does not have permission to write the file at the desired location

tpoliaw: If it's not writable, isn't that a problem with the `tmp_path` fixture and not the session cache?

ZohebShaikh: The purpose of this test is to ensure that `SessionCacheManager` always returns a file path rather than a directory, allowing BlueAPI to write the `token_cache` to the specified `file_path`. While the `tmp_path` fixture itself shouldn't cause issues, this approach validates the behavior by touching the file, ensuring the path is writable. 

Currently, I’m not aware of a  way to confirm if a path is specifically a file path and not a directory without either opening or attempting to touch the file. If there’s a more elegant solution to this, I’d be open to suggestions.

I have changed the test slightly since last review ..Instead of open the file ,I'm doing `touch` on the path provide and then doing `isFile` on the given path...The `touch` should raise a error if Session_cache returns a directory path rather than a file path

DiamondJoseph (approved):

PR #755, opened 2024-12-13
Minimal configuration for inserting documents into Tiled

DiamondJoseph:


pyproject.toml
```diff
@@ -95,7 +99,8 @@ addopts = """
     --ignore=src/blueapi/startup
     """
 # https://iscinumpy.gitlab.io/post/bound-version-constraints/#watch-for-warnings
-filterwarnings = ["error", "ignore::DeprecationWarning"]
+# Unignore UserWarning after Pydantic warning removed from bluesky/bluesky and release
```
DiamondJoseph, line 102: https://github.com/bluesky/bluesky/pull/1867

+ a release of bluesky

src/blueapi/config.py
```diff
@@ -41,6 +41,15 @@ class StompConfig(BaseModel):
     auth: BasicAuthentication | None = None
 
 
+class TiledConfig(BaseModel):
+    """
+    Config for connecting to a tiled instance
+    """
+
+    uri: str
```
callumforrester, line 49: Could: Can we use separate `host` and `port` fields to be consistent with the other config models?

pyproject.toml
```diff
@@ -95,7 +99,8 @@ addopts = """
...
+filterwarnings = ["error", "ignore::DeprecationWarning", "ignore::UserWarning"]
```
callumforrester, line 103: Is there any way to only ignore warnings from the tests in question? Otherwise we may end up removing this in a few months time only to find the tests blowing up in a bunch of other places?

src/blueapi/service/interface.py
```diff
@@ -48,6 +50,19 @@ def worker() -> TaskWorker:
     return worker
 
 
+@cache
+def tiled_inserter():
```
callumforrester, line 54: Should:

```suggestion
def tiled_inserter() -> TiledClient:
```

...or whatever it's called

pyproject.toml
callumforrester: Should: Include issue link in a comment in the file

callumforrester (changes requested):

callumforrester (approved):

src/blueapi/service/interface.py
keithralphs: Now called tiled-write and includes the approriate return type

helm/blueapi/values.yaml
```diff
@@ -155,11 +155,11 @@ restartOnConfigChange: true
...
-# - name: RABBITMQ_PASSWORD
-#   valueFrom:
-#     secretKeyRef:
-#       name: rabbitmq-password
-#       key: rabbitmq-password
+#  - name: TILED_API_KEY
+#    valueFrom:
+#      secretKeyRef:
+#        name: tiled_api_key
+#        key: tiled_api_key
```
DiamondJoseph, line 162:
```suggestion
# - name: TILED_API_KEY
#   valueFrom:
#     secretKeyRef:
#       name: tiled_api_key
#       key: tiled_api_key
```

nit: whitespace same as replacement

tests/system_tests/services/tiled_config/config.yaml
```diff
@@ -0,0 +1,19 @@
+authentication:
+  # The default is false. Set to true to enable any HTTP client that can
+  # connect to _read_. An API key is still required to write.
+  allow_anonymous_access: true
```
DiamondJoseph, line 4:
```suggestion
  # Any HTTP client that can connect can read, an API key is still required to write.
  allow_anonymous_access: true
```

tests/system_tests/services/tiled_config/config.yaml
```diff
@@ -0,0 +1,19 @@
...
+      uri: "sqlite:////storage/catalog.db"
+      writable_storage: "sqlite:////storage/catalog.db"
+      # This creates the database if it does not exist. This is convenient, but in
+      # a horizontally-scaled deployment, this can be a race condition and multiple
+      # containers may simultaneously attempt to create the database.
+      # If that is a problem, set this to false, and run:
+      #
+      # tiled catalog init URI
+      #
+      # separately.
```
DiamondJoseph, line 18:
```suggestion
```

This config is only for a non-scaled deployment so don't need to worry about this

helm/blueapi/values.yaml
keithralphs: Done

PR #761, opened 2024-12-16
Add CI job to run system tests

### LINKED ISSUE #630: Run System Tests in CI
Following the merge of #535

callumforrester:
Fixes #630

### DIFF
--- a/.github/workflows/_system_test.yml
+++ b/.github/workflows/_system_test.yml
@@ -1,0 +1,29 @@
+on:
+  workflow_call:
+
+env:
+  # https://github.com/pytest-dev/pytest/issues/2042
+  PY_IGNORE_IMPORTMISMATCH: "1"
+
+jobs:
+  run:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Start RabbitMQ
+        uses: namoshek/rabbitmq-github-action@v1
+        with:
+          ports: "61613:61613"
+          plugins: rabbitmq_stomp
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Need this to get version number from last tag
+          fetch-depth: 0
+
+      - name: Install python packages
+        uses: ./.github/actions/install_requirements
+  
+      - name: Run tests
+        run: tox -e system-test
--- a/.github/workflows/_system_test.yml
+++ b/.github/workflows/_system_test.yml
@@ -26,7 +26,7 @@ jobs:
           plugins: rabbitmq_stomp
 
       - name: Start Blueapi Server
-        run: blueapi serve &
+        run: blueapi -c ${{ github.workspace }}/tests/unit_tests/example_yaml/valid_stomp_config.yaml serve &
 
       - name: Run tests
         run: tox -e system-test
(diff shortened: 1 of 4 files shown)


tests/system_tests/test_blueapi_system.py
```diff
@@ -58,6 +59,22 @@ def client_with_stomp() -> BlueapiClient:
...
+    attempts_remaining = 20
+    while attempts_remaining > 0:
+        try:
+            client.get_environment()
+            return
+        except ConnectionError:
+            ...
+        attempts_remaining -= 1
+        time.sleep(0.5)
+    raise TimeoutError("No connection to the blueapi server")
```
DiamondJoseph, line 75:
```suggestion
    for _ in range(20):
        try:
            client.get_environment()
            return
        except ConnectionError:
            ...
        time.sleep(0.5)
    raise TimeoutError("No connection to the blueapi server")
```

nit: Why keep track when Python'll do it for ya mate

.github/workflows/_system_test.yml
```diff
@@ -0,0 +1,32 @@
...
+          fetch-depth: 0
+
+      - name: Install python packages
+        uses: ./.github/actions/install_requirements
+
+      - name: Start RabbitMQ
+        uses: namoshek/rabbitmq-github-action@v1
+        with:
+          ports: "61613:61613"
+          plugins: rabbitmq_stomp
```
DiamondJoseph, line 26: No longer running tests against ActiveMQ? Seems reasonable to me, given that the image is so deprecated, but we could replace it with Apache Artemis, just to make sure our Stomp interactions are generally valid.

DiamondJoseph (approved):

tests/system_tests/test_blueapi_system.py
callumforrester: Good shout, see also https://github.com/DiamondLightSource/blueapi/issues/763

.github/workflows/_system_test.yml
callumforrester: We run the system tests of https://github.com/DiamondLightSource/bluesky-stomp/ against activemq and rabbitmq, which I think is sufficient

tests/system_tests/test_blueapi_system.py
```diff
@@ -101,6 +118,7 @@ def clean_existing_tasks(client: BlueapiClient):
     yield
 
 
+@pytest.mark.xfail(reason="https://github.com/DiamondLightSource/blueapi/issues/676")
```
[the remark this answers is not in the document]
callumforrester: I like the longer message, will implement.

I'm not a fan of having special steps to make the tests work, under this you would have to somehow know that you need to do `REQUIRES_AUTH=False tox -e system-tests`. A dev environment that works in a predictable way out-of-box is better IMO. I'm also not hugely in favour of having the local tests work differently to the CI, we were in that situation before we made bluesky-stomp and it caused lots of debugging problems. I would rather mark as xfail, clearly showing we know there's a problem, and fix it when we can. On a practical level it makes no difference, following the instructions for the system tests locally will result in a success message.

PR #764, opened 2024-12-17
Training rigs system test

ZohebShaikh:


callumforrester:
As discussed, I think we should make a new repository for our beamline system tests. I can make a new one and review these changes there or review them here, which is easier @ZohebShaikh?

Either way I'm going to put in bold that...

# THIS PR SHOULD NOT BE MERGED!

callumforrester:
Okay, I'll have a think about where to put it. The matter your raise actually came up in the DAQ away day earlier and we concluded (with security people present) that a lot of information like that is not actually considered sensitive.

tests/system_tests/test_blueapi_system.py
```diff
@@ -338,6 +260,30 @@ def test_get_current_state_of_environment(client: BlueapiClient):
     assert client.get_environment() == EnvironmentResponse(initialized=True)
 
 
+@pytest.mark.xfail(
```
callumforrester, line 263: Must: This test should be skipped rather than xafiled, xfail is for tests that fail all the time, not intermittently. It just has the effect of reversing the result (i.e. if the test sometimes passes the suite will fail)

tests/system_tests/common.py
```diff
@@ -0,0 +1,79 @@
...
+)
+
+
+def clean_existing_tasks(client: BlueapiClient) -> None:
+    for task in client.get_all_tasks().tasks:
+        client.clear_task(task.task_id)
+
+
+def check_all_events(all_events: list[AnyEvent]):
+    assert isinstance(all_events[0], WorkerEvent) and all_events[0].task_status
```
callumforrester, line 45: Should: Be explicit

```suggestion
    assert isinstance(all_events[0], WorkerEvent) and (all_events[0].task_status is not None)
```

tests/system_tests/expected_data/plans_p46.json
callumforrester, line 1: I think this approach is good for the general system tests that run in CI. However for beamlines it is probably too restrictive, we don't want to have to change this file for every change we make to any plan. Instead the tests should identify several critical capabilities and test that they are present e.g.

"is there a plan called X?"
"does Y happen when I run it?"
"is there a device called Z?"

You also shouldn't have to worry to about testing for the very basic capabilities of blueapi (e.g. can I reload the environment?).

callumforrester (changes requested):
I have left a few comments, but honestly it's hard to tell what is intended for the test rigs vs. the generic blueapi system tests, so I might wait until the code is moved to continue reviewing.

tests/system_tests/test_blueapi_system.py
callumforrester: Worth noting I have a long term solution WIP here: https://github.com/DiamondLightSource/blueapi/pull/743

callumforrester:
@ZohebShaikh To expand on the Gitlab conversation, I think I'm settling on Github. The premise seems to be that there are 3 categories of information: Not sensitive at all, internally sensitive and totally sensitive and that internally sensitive information is fine to store under VCS as long as its not on the public internet, as opposed to totally sensitive information like passwords, which should never be stored (unencrypted) under VCS under any circumstances.

I'm going to posit that the actual amount of information that is internally sensitive is small, if not zero. Meanwhile if go on Gitlab then we lose a lot of the free infrastructure that we get with the copier template (CI/CD, issue template etc.)
I feel that the cost of accommodating the small amount of internally sensitive information with secrets is less than the cost of going with Gitlab, rolling our own CI etc. but let me know your thoughts

PR #766, opened 2024-12-19
Pass appVersion to helm package command

DiamondJoseph:
Simplifies and standardises the passing of the github ref into the Chart.yaml


callumforrester (approved):
How about that, it works...

PR #771, opened 2025-01-07
Tidy Helm action

DiamondJoseph:


.github/workflows/_container.yml
```diff
@@ -16,20 +16,7 @@ jobs:
...
 
-        run: |
-          ref="${{ github.ref_name }}"
-          my_regex="${{env.SEMVER_REGEX}}"
-          if [[ "$ref" =~ $my_regex ]]; then
-            echo "SemVer compliant version: $ref"
-          else
-            echo "Invalid SemVer version: $ref"
-            exit 1
-          fi
```
callumforrester, line 32: As discussed offline, I would like proof we don't need this

callumforrester (changes requested):

.github/workflows/_container.yml
callumforrester: Given my recent mishap releasing something with the wrong tag, should we just put something like this into the `check` CI and have it only run on a tag, such that the entire pipeline fails early if someone tries to push an invalid tag?

DiamondJoseph:
Superseded by #880.

Issue #633, opened 2024-09-11
Blueapi context should except when it can't find an injected device 


noemifrisina:
As a user, I would like for the blueapi context to raise an exception when it's unable to find a device injected into a plan using Dodal's `inject` function.
I ran into this issue while running a plan from the command line on I24 with `blueapi -c config.yaml controller run moveto ...`. The plan  did not work and there was no indication of what was wrong except for an error in the dodal logs indicating that instead of the `PMAC` device only a string was passed as argument.

Here's the plan
```python
def moveto(place: str = "origin", pmac: PMAC = inject("pmac")) -> MsgGenerator:
    ...i
    if place == "zero":
         yield from bps.trigger(pmac.to_xyz_zero)
    ...
```
and the error trace from the dodal logs

```[I 240910 15:06:48.744 i24ssx_Chip_Manager_py3v1:611] Move to: zero
[I 240910 15:06:48.746 i24ssx_Chip_Manager_py3v1:613] Chip aspecific move.
[E 240910 15:06:48.747 run_engine:1733] Run aborted
Traceback (most recent call last):
  File "/dls_sw/i24/software/bluesky/mx_bluesky_testing/mx-bluesky/.venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 1588, in _run
    msg = self._plan_stack[-1].send(resp)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dls_sw/i24/software/bluesky/mx_bluesky_testing/mx-bluesky/src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py", line 614, in moveto
    yield from bps.trigger(pmac.to_xyz_zero)
                           ^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'to_xyz_zero'
```

The docs seem to suggest that the context excepts when there's no device ([here](https://github.com/DiamondLightSource/blueapi/blob/07c273f55d419664e11aa4eb2400e3ad3b5a5c4e/docs/how-to/write-plans.md?plain=1#L52)), but this does not seem to be the case...

noemifrisina:
I don't really have a strong preference. I think raising when trying to run a plan may be enough to have an indication of what's wrong, as long as it fails right away and not after going through some bits of the plan... 
Also, in this case I'm starting to suspect it was a problem with all the devices and raising on start-up might not have been helpful?

[...]

PR #773, opened 2025-01-09
Support inject for default plan arguments again

### LINKED ISSUE #647: `inject` not working on i04
In testing on i04 today we had a plan like:
```python
def thaw(smargon:Smargon = inject("smargon")):
    pass
```

This failed if we called it via `run_plan()` from GDA but passed if we did `run_plan("smargon")`. I04 is loading `dodal` from the scratch area.

## Acceptance Criteria
* `inject` works correctly

tpoliaw:
Let inject work again by using all fields of the built pydantic model including the default ones provided by inject

Fixes #647, #633 

### Instructions to reviewer on how to test:
1. Run plan that uses `inject` for default parameters, (eg `blueapi controller run stp_snapshot '{"detectors":["x"]}'` to run the plan from `example_plans`)
2. See that the plan runs correctly with `sample_temperature` and `sample_pressure`. (It may fail at the end of the run with a different error relating to path providers for the detector but this is unrelated).
3. If a plan is created using `inject("<misssing_device_name>")` (as for #633), the plan will not be accepted and blueapi will return `422 Unprocessable Entity`. The detailed error message is generated for this but it's not immediately obvious how to access it. Either way, that should be its own issue.

### Checks for reviewer
- [ ] Would the PR title make sense to a user on a set of release notes

### DIFF
--- a/src/blueapi/worker/task.py
+++ b/src/blueapi/worker/task.py
@@ -22,7 +22,8 @@ class Task(BlueapiBaseModel):
 
     def prepare_params(self, ctx: BlueskyContext) -> Mapping[str, Any]:
         model = _lookup_params(ctx, self)
-        return _model_to_kwargs(model)
+        # Re-create dict manually to avoid nesting in model_dump output
+        return {field: getattr(model, field) for field in model.__pydantic_fields__}
 
     def do_task(self, ctx: BlueskyContext) -> None:
         LOGGER.info(f"Asked to run plan {self.name} with {self.params}")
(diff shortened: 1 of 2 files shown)


callumforrester (approved):
Nice, thanks!

Is it feasible/worthwhile to write a regression test?

src/blueapi/worker/task.py
```diff
@@ -22,7 +22,7 @@ class Task(BlueapiBaseModel):
 
     def prepare_params(self, ctx: BlueskyContext) -> Mapping[str, Any]:
         model = _lookup_params(ctx, self)
-        return _model_to_kwargs(model)
+        return model.model_dump(warnings=False)
```
tpoliaw, line 25: This might need to be `{field: getattr(model, field) for field in model.model_dump(warnings=False)}`.

@callumforrester ,  do you have an example of the nested model you wanted to avoid in the `_model_to_kwargs` method?

tpoliaw:
Added a couple of tests to cover the 'normal' use and the failure case in 633

callumforrester (approved):
LGTM!

PR #794, opened 2025-01-22
Add BlueapiBaseModel config to StompConfig

ZohebShaikh:


callumforrester (commented):
If there are any other `BaseModel`s in the config I think we should convert them at the same time to keep it consistent. 

Also we should link https://github.com/DiamondLightSource/bluesky-stomp/issues/32

callumforrester (approved):

Issue #643, opened 2024-09-17
dodal.common.MsgGenerator and dls_bluesky_core.core.MsgGenerator not equal


DiamondJoseph:
Previously we had checked that plans were annotated with `is dls_bluesky_core.core.MsgGenerator`. This began causing issue, and we had to check `== dls_bluesky.core.core.MsgGenerator`. This now appears to also not be compatible in a built container.

It is time for a clean break and removal of dls_bluesky_core. But also a cautionary tale if we to move MsgGenerator into bluesky.

_Originally posted by @DiamondJoseph in https://github.com/DiamondLightSource/blueapi/issues/533#issuecomment-2355981325_

callumforrester:
Shall we now get it all to point at the one in bluesky main: https://github.com/bluesky/bluesky/blob/7c2a44f09a747f59a17abc6cfdfe934c61c266fb/src/bluesky/utils/__init__.py#L98

DiamondJoseph:
Has that really been there 6 months...

DiamondJoseph:
There hasn't been a release since it went in. I'm chasing up now.

DiamondJoseph:
Closed by #832

PR #798, opened 2025-01-23
Fix parametrized `MsgGenerator` plans not being detected

### LINKED ISSUE #796: Blueapi does not pick up plans with parametrrized return type
Blueapi detects that functions are plans if they have the appropriate type annotation. For example:

```python


# Blueapi thinks this is a plan
def plan_1() -> MsgGenerator:
    ...


# Blueapi does not think this is a plan
def util_function() -> list[int]:
    ...
```

We can parametrize `MsgGenerator` with the plan return type e.g.

```python

def plan_3() -> MsgGenerator[str]:
    ...
```

Blueapi should pick that up as a plan but does not

## Steps To Reproduce
1. Paste the above snippet into `blueapi.startup.plans`
2. Run `blueapi serve`
3. Run `blueapi controller plans`
4. See that `plan_3` is not in the response

## Acceptance Criteria
- Plans such as `plan_3` are visibly exposed by blueapi

callumforrester:
Fixes #796


src/blueapi/core/bluesky_types.py
```diff
@@ -1,8 +1,11 @@
 import inspect
 from collections.abc import Callable, Mapping
+from collections.abc import Generator as ConcreteGenerator
```
tpoliaw, line 3: What is `Generator` colliding with? I think it would be less confusing to use `abc.Generator` instead of renaming it to something custom.

src/blueapi/core/bluesky_types.py
```diff
@@ -76,7 +79,10 @@ def _follows_bluesky_protocols(obj: Any) -> bool:
 
 def is_bluesky_plan_generator(func: PlanGenerator) -> bool:
     try:
-        return get_type_hints(func).get("return") == MsgGenerator
+        return_type = get_type_hints(func).get("return")
+        return (get_origin(return_type) is ConcreteGenerator) and (
+            Msg in get_args(return_type)
```
tpoliaw, line 84: Should this be something like `get_args(return_type)[0] == Msg`? At the moment (I think) this will return true for something like `def foo() -> Generator[int, int, Msg]: ...` which I don't think counts as a plan?

src/blueapi/core/bluesky_types.py
callumforrester: Forgot to change that back, my bad

src/blueapi/core/bluesky_types.py
callumforrester: Thought of [some extra test cases](https://github.com/DiamondLightSource/blueapi/pull/798/commits/dd737df45869d8c764c403b020b5d402a6eb1754) that I was worried about in this case but they still pass, so changed

src/blueapi/core/bluesky_types.py
```diff
@@ -76,7 +78,10 @@ def _follows_bluesky_protocols(obj: Any) -> bool:
 
 def is_bluesky_plan_generator(func: PlanGenerator) -> bool:
     try:
-        return get_type_hints(func).get("return") == MsgGenerator
+        return_type = get_type_hints(func).get("return")
+        return (get_origin(return_type) is Generator) and (
+            get_args(return_type)[0] is Msg
```
tpoliaw, line 83: `is` vs `==`? I can't remember which is recommended for comparing types, but wasn't the previous version changed to ` == MsgGenerator` to fix some bug?

callumforrester: Yeah, I don't think we ever got to the bottom of that bug (#643). I'll change it and add a test case that tries to replicate it, to be safe.

tpoliaw (approved):

PR #803, opened 2025-01-27
Acqp-323 prevent blank connection errors in STOMP connections

### LINKED ISSUE #792: Stomp connection error hidden in environment response
> Another instance of this problem: https://github.com/DiamondLightSource/blueapi/blob/caf7e466dba86eba5eb38da2834548f670d04e8d/src/blueapi/service/runner.py#L85
> 
> 
> 
> It looks like `str(stomp.ConnectFailedException())` is an empty string, so that's not helpful...
> 
> 
> 
> Should maybe change it to `str(type(e)) + ": " + str(e)`? 

 _Originally posted by @callumforrester in [#503](https://github.com/DiamondLightSource/blueapi/issues/503#issuecomment-2331084989)_

## Acceptance Criteria
- Exception logs never hidden by pydantic validation errors (e.g. because the exception message is `""`)
- Type of exception and `str` representation always included

keithralphs:
Add handler functions to check for Exceptions with message that are None, Empty String or a String that consists only of spaces

Fixes #792

### DIFF
--- a/src/blueapi/client/client.py
+++ b/src/blueapi/client/client.py
@@ -22,6 +22,7 @@ from blueapi.service.model import (
     TasksListResponse,
     WorkerTask,
 )
+from blueapi.service.runner import BLANK_REPORT
 from blueapi.worker import Task, TrackableTask, WorkerEvent, WorkerState
 from blueapi.worker.event import ProgressEvent, TaskStatus
 
@@ -412,7 +413,10 @@ class BlueapiClient:
         while too_late is None or time.time() < too_late:
             # Poll until the environment is restarted or the timeout is reached
             status = self._rest.get_environment()
-            if status.error_message is not None:
+            if (
+                status.error_message is not None
+                and BLANK_REPORT not in status.error_message
+            ):
                 raise BlueskyRemoteControlError(
                     f"Error reloading environment: {status.error_message}"
                 )
--- a/src/blueapi/client/client.py
+++ b/src/blueapi/client/client.py
@@ -22,7 +22,6 @@ from blueapi.service.model import (
     TasksListResponse,
     WorkerTask,
 )
-from blueapi.service.runner import BLANK_REPORT
 from blueapi.worker import Task, TrackableTask, WorkerEvent, WorkerState
 from blueapi.worker.event import ProgressEvent, TaskStatus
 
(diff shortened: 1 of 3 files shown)


src/blueapi/service/runner.py
```diff
@@ -98,15 +100,17 @@ def stop(self):
             self._state = EnvironmentResponse(
                 environment_id=environment_id,
                 initialized=False,
-                error_message=self._state.error_message,
+                error_message=_safe_message(
+                    type(self).__name__, self._state.error_message
+                ),
```
callumforrester, line 105: Should: Since this just propagates the previous error message it will become `WorkerDispatcher: <previous message>`, where it was `<previous message>` before. That feels very inconsistent, the user should see the same message when the same thing breaks. Since this is always set by `_safe_exception_message` you're probably safe just propagating it without checking?

src/blueapi/client/client.py
```diff
@@ -412,7 +413,10 @@ def _wait_for_reload(
         while too_late is None or time.time() < too_late:
             # Poll until the environment is restarted or the timeout is reached
             status = self._rest.get_environment()
-            if status.error_message is not None:
+            if (
+                status.error_message is not None
+                and BLANK_REPORT not in status.error_message
```
callumforrester, line 418: Must: Not sure I understand the logic here, `BLANK_REPORT` being in the error message implies there's been an error without text (i.e. an unknown error), but we should still raise.

callumforrester (changes requested):
Thanks Keith, at the risk of being annoying, please could you reference the Github issue rather than the internal issue? We check those every sprint anyway, so would be good if kept things automatically closing on Github.

Also, please could you make an issue to revisit the `EnvironmentResponse` API as discussed?

src/blueapi/client/client.py
keithralphs: Yes, this is basically saying if there's an error message, but it's not the one that gets inserted when there's been an exception that didn't have its own message (i.e. it's and error message that actually tells you something) then raise the error. This is equivalent to the previous behaviour, the only difference being that, in that case, no standard message would have been inserted so it would just be None.

callumforrester: So from discussion, here is how we want the client to behave:

  | `initialized=True` | `initialized=False`
-- | -- | --
`error_message is None` | Return | Block until next poll
`error_message is not None` | Undefined | Raise

This is true regardless of whether `BLANK_REPORT` is present or not, because `error_message` not being `None` means an exception has been thrown, expanded table:

|                                                                     | `initialized=True` | `initialized=False`   |
|---------------------------------------------------------------------|--------------------|-----------------------|
| `error_message is None`                                             | Return             | Block until next poll |
| `error_message is not None` and `BLANK_REPORT in error_message`     | Undefined          | Raise                 |
| `error_message is not None` and `BLANK_REPORT not in error_message` | Undefined          | Raise                 |

keithralphs: Issue #806 created to cover implementation and use of this logic

PR #807, opened 2025-01-31
795 plan sources

### LINKED ISSUE #795: Importing `attach_data_session_metadata_decorator` means it tries to export as a plan
If a plan file contains

```python
from dodal.plan_stubs.data_session import attach_data_session_metadata_decorator
```

Then blueapi tries to exposure `attach_data_session_metadata_decorator` as a plan.

## Steps To Reproduce
1. Add the above line to one of the startup files in `blueapi.startup`
2. Try to run the server

## Acceptance Criteria
- Only plans defined in the file imported (or exposed via `__all__` are visible in blueapi

callumforrester:
Fixes #795


tests/unit_tests/core/test_context.py
```diff
@@ -170,6 +170,20 @@ def test_add_plan_from_module(empty_context: BlueskyContext) -> None:
     assert EXPECTED_PLANS == empty_context.plans.keys()
 
 
+def test_only_plans_from_source_module_detectede(empty_context: BlueskyContext) -> None:
```
DiamondJoseph, line 173:
```suggestion
def test_only_plans_from_source_module_detected(empty_context: BlueskyContext) -> None:
```
nit

tests/unit_tests/core/test_context.py
```diff
@@ -170,6 +170,20 @@ def test_add_plan_from_module(empty_context: BlueskyContext) -> None:
...
+    import tests.unit_tests.core.fake_plan_module_with_imports as plan_module
+
+    empty_context.with_plan_module(plan_module)
+    assert {"plan_c", "plan_d"} == empty_context.plans.keys()
+
+
+def test_only_plans_from_all_in_module_detectede(empty_context: BlueskyContext) -> None:
```
DiamondJoseph, line 180:
```suggestion
def test_only_plans_from_all_in_module_detected(empty_context: BlueskyContext) -> None:
```
nit

src/blueapi/utils/modules.py
```diff
@@ -34,3 +35,17 @@ def get_named_subset(names: list[str]):
...
+    Check if an object is originally from a particular module, useful to detect
+    whether it actually comes from a nested import.
+
+    Args:
+        obj: Object to check
+        module: Module to check against object
+    """
+    return (
+        hasattr(obj, "__module__") and importlib.import_module(obj.__module__) is module
+    )
```
DiamondJoseph, line 51: When this is being used over all of the objects in a module, is it going to import the module and run any side effects for every object? This could (in plan modules out of our control) be dangerous. Or does it cache and only import the module once?

DiamondJoseph (changes requested):

docs/how-to/write-plans.md
```diff
@@ -28,7 +28,9 @@ def my_plan(
     ...
 ```
 
-The type annotations (e.g. `: str`, `: int`, `-> MsgGenerator`) are required as blueapi uses them to detect that this function is intended to be a plan and generate its runtime API.
+## Detection
+
+The type annotations in the example above (e.g. `: str`, `: int`, `-> MsgGenerator`) are required as blueapi uses them to detect that this function is intended to be a plan and generate its runtime API. If there is an [`__all__` dunder](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) present in the module, blueapi will read that and import anything within that qualifies as a plan, per its type annotations. If not it will simpyl read everything in the module that hasn't been imported, for example it will ignore a plan imported from another module.
```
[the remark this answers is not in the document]
callumforrester: Yes, master yoda

src/blueapi/utils/modules.py
callumforrester: It caches

```python
# tt.py
print("Hello world")

```

```python
>>> import importlib
>>> importlib.import_module("tt")
Hello world
<module 'tt' from '/workspaces/blueapi/tt.py'>
>>> importlib.import_module("tt")
<module 'tt' from '/workspaces/blueapi/tt.py'>
```

src/blueapi/utils/modules.py
```diff
@@ -34,3 +35,17 @@ def get_named_subset(names: list[str]):
         for name, value in mod.__dict__.items():
             if not name.startswith("_"):
                 yield value
+
+
+def is_function_sourced_from_module(obj: Any, module: ModuleType) -> bool:
```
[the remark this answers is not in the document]
callumforrester: Going with your first suggestion, I'm not aware of any constraint that guarantees that something being `Callable` means it has a `__module__` dunder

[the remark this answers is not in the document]
DiamondJoseph: I'm just `nit`ing the name and the params telling different stories

DiamondJoseph (approved):

PR #827, opened 2025-02-18
Fix PEP8 violation and doc strings

ZohebShaikh:


DiamondJoseph (changes requested):
Other than the RunEngine looks good, if you revert the run engine good to go

PR #830, opened 2025-02-20
Search by protocol feature and tests

stan-dot:


src/blueapi/service/main.py
```diff
@@ -215,6 +215,18 @@ def get_device_by_name(name: str, runner: WorkerDispatcher = Depends(_runner)):
     return runner.run(interface.get_device, name)
 
 
+@router.get(
```
DiamondJoseph, line 218: We already have a /devices/ endpoint defined above on line 200. I would make changes to that, making protocol_name an optional query parameter with default None. 

I'd also make the query parameter `type` such that we can later extend it to abstract and concrete types.

stan-dot: > I'd also make the query parameter type such that we can later extend it to abstract and concrete types.

when we need it we can make that

DiamondJoseph: That would turn a minor revision of the API (extending it to take additional parameters) into a major revision (removing a query parameter and adding another), which would require additional time and concern around versioning, deprecating the old parameter and warning any clients. I really think it's easier to just be broad now.

stan-dot: tbh I have no idea what do you mean here:

> I'd also make the query parameter type such that we can later extend it to abstract and concrete types.

DiamondJoseph: Just renaming the name of the parameter from `protocol_name` to `type`.

stan-dot: https://peps.python.org/pep-0695/

imo `type` is a bad parameter name, possibly clashing with python keywords

DiamondJoseph: https://fastapi.tiangolo.com/tutorial/query-params-str-validations/#alias-parameters

```
expected_type: Annotated[str | None, Query(alias="type")] = None
```

Is an option, or expected_type or similar- I'm not attached to any particular name but just to being able to expand to concrete types imminently.

PR #828, opened 2025-02-18
WIP numtracker integration

callumforrester:

Issue #774, opened 2025-01-10
Any user may begin the running of a task, not just the user that submitted it


DiamondJoseph:
Possibly a desirable use case ("You queue up a bunch of scans you want to do, and I'll just churn through them while you get lunch") but also possibly undesirable (most(?) of the time, process will be submit task, run task, as exposed through the python cli).

When authentication information is passed as part of submitting a task, it may become necessary to check that the authentication information passed as part of running the task matches 

## Acceptance Criteria
- Either user1 cannot run a task that has been submitted by user2
- Or situations in which user1 may run a task submitted by user2 have been documented
- Or configuration is available to switch between the two configurations

callumforrester:
The API is transaction-like, I would suggest that the only user who begins a transaction may submit. Blueapi is explicitly not a queueing mechanism and should not support the use case of queueing up a bunch of plans to be run later. The idea is that you should write your own service to do that.

PR #832, opened 2025-02-21
Pass visit id when starting task

### LINKED ISSUE #793: Pass visit ID to blueapi where appropriate
At the time of writing we are able to effectively hard code the current instrument session (a.k.a. visit) ID into blueapi via either the dodal module or numtracker config. This means that it requires a restart to write data to a different instrument session.

## Acceptance Criteria
- Must pass an instrument session ID with a plan request
- The ID is referred to by the agreed name, `instrumentSession`, instead of `session` or `visit`
- Get GDA to pass current visit when submitting a plan
- CLI supports passing an instrument session

DiamondJoseph:
Fixes #793 

This does all of the grunt work necessary to get to a point where the instrument_session can be checked with the authentication information passed with the request to be checked to ensure the user is permitted to make their request.

Additional tasks coming off of this one and possible to be merged into this PR before it is merged in turn into main:

- Allow user to select a visit when they log in
- Cache the above visit if it is selected and pass it with every request to prevent having to be explicit
- Checking authentication information and instrument_session to ensure user is authorized to make request
- Using the instrument_session as part of the call to the PathProvider (see #828 et al.)

### DIFF
--- a/docs/reference/openapi.yaml
+++ b/docs/reference/openapi.yaml
@@ -252,7 +252,7 @@ components:
       type: object
 info:
   title: BlueAPI Control
-  version: 0.0.5
+  version: 0.0.6
 openapi: 3.1.0
 paths:
   /devices:
@@ -527,6 +527,13 @@ paths:
 
         This will return an error response if the worker is not idle.'
       operationId: set_active_task_worker_task_put
+      parameters:
+      - in: query
+        name: instrument_session
+        required: true
+        schema:
+          title: Instrument Session
+          type: string
       requestBody:
         content:
           application/json:
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -45,7 +45,7 @@ from .model import (
 )
 from .runner import WorkerDispatcher
 
-REST_API_VERSION = "0.0.5"
+REST_API_VERSION = "0.0.6"
 
 RUNNER: WorkerDispatcher | None = None
 
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -112,7 +112,7 @@ def get_app(config: ApplicationConfig):
 
 
 def verify_access_token(config: OIDCConfig):
-    jwkclient = jwt.PyJWKClient(config.jwks_uri)
+    jwk_client = jwt.PyJWKClient(config.jwks_uri)
     oauth_scheme = OAuth2AuthorizationCodeBearer(
         authorizationUrl=config.authorization_endpoint,
         tokenUrl=config.token_endpoint,
@@ -120,7 +120,7 @@ def verify_access_token(config: OIDCConfig):
     )
 
     def inner(access_token: str = Depends(oauth_scheme)):
-        signing_key = jwkclient.get_signing_key_from_jwt(access_token)
+        signing_key = jwk_client.get_signing_key_from_jwt(access_token)
         jwt.decode(
             access_token,
             signing_key.key,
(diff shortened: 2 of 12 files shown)


src/blueapi/service/main.py
```diff
@@ -304,6 +304,7 @@ def get_tasks(
 @start_as_current_span(TRACER, "task.task_id")
 def set_active_task(
     task: WorkerTask,
+    instrument_session: str,
```
callumforrester, line 307: Should: For `PUT`/`POST` requests you should include the parameters in the request body rather than the URL.

(see https://www.moesif.com/blog/technical/api-development/REST-API-Design-Best-Practices-for-Rest-API-Query-Parameters-and-Query-String-Usage/)

callumforrester, line 307: Could: The approach here is to make tasks independent of an instrument session and only pass the session when stating them. The alternative is to embed the session in the task, which I think I'm more in favour of. The idea of this API is that it is transaction style where tasks are the transaction (i.e. cheap and binnable and everything about them is set up in advance) and `set_active_task` just commits the transaction. I think I would prefer the session embedded in the task, but open to discussing.

callumforrester, line 307: Could: There is an ongoing [discussion](https://github.com/DiamondLightSource/dodal/issues/991#issuecomment-2637883125) about whether the session should be a required parameter. I'm still on the side of making it optional, but either way the discussion should be resolved and actioned before this is merged.

docs/reference/openapi.yaml
```diff
@@ -252,7 +252,7 @@ components:
       type: object
 info:
   title: BlueAPI Control
-  version: 0.0.5
+  version: 0.0.6
```
callumforrester, line 255: Should: Given the addition of a new required parameter I think this is 0.1.0, but since I'm arguing that it should be optional in another comment, leave until that's resolved.

callumforrester (changes requested):
Thanks, this looks like a good start

src/blueapi/service/main.py
DiamondJoseph: We check authorization to write to a visit as part of beginning a task, as part of the call to the NumTracker, so I think the instrument_session should be passed as part of the request to begin the task. 

See also discussion on #774: if user_a with permission to write to x,y creates a task and user_b with permission to write to x submits it- if user_a has incorrectly set the instrument_session to y, user_b cannot run the scan due to permissions issues. 

Or if user_b maliciously creates a task with instrument_session y to allow their data to be written to e.g. the commissioning directory and so not archived or exposed.

Submitting the task to be run is the user knowingly saying "collect <this data>", and I think that should include "to <this instrument_session>"

DiamondJoseph: Does the PathProvider clear its instrument_session after a task has been completed, to allow us to except when there is no instrument_session passed and we expect there to be?

callumforrester: Per offline discussion: This shouldn't be merged until we do an AuthZ check of session/user, but that doesn't actually affect this dicussion. The session is embedded in the task and is checked against the user who triggers the task. The question of whether there should be separate AuthZ to check that the creator and submitter are the same user doesn't affect this.

callumforrester: I presume blueapi would have to clear it at the beginning of any task, whether or not a session is passed, to make that consistent.

tpoliaw:
Is this being entirely replaced by #1062?

Issue #744, opened 2024-12-02
 Scratch area system tests


callumforrester:
As a developer I would like to have some system tests where we check out a scratch area, modify it, and poke blueapi for the correct behaviour. 

## Acceptance Criteria
- System tests test the following cases
  - A custom plan is added via the scratch area and appears on `/plans`
  - A custom devices is added via the scratch area and appears on `/devices`
  - Broken code is added and the environment fails to reload
  - Any other cases that may be useful...

PR #861, opened 2025-03-11
Add python environment endpoint

### LINKED ISSUE #760: Query scratch status via API
As a support engineer I would like to be able to query blueapi for the state of the scratch directory to aid in installation/debugging, in particular I would like to know:
- Which packages are installed via the scratch directory
- Which of those replaced packages that were already in the environment
- The versions checked out
- Whether uncommitted changes are present

## Acceptance Criteria
- There are REST endpoint(s) for querying all of the above
- They are tested
- They are supported by the CLI

ZohebShaikh:
closes #760


src/blueapi/service/model.py
```diff
@@ -157,6 +157,22 @@ class EnvironmentResponse(BlueapiBaseModel):
...
+    is_dirty: bool = Field(description="Does the repository have uncommitted changes")
+
+
+class ScratchResponse(BlueapiBaseModel):
+    """
+    State of the scratch area.
+    """
+
+    package_info: list[RepositoryStatus] = Field(
+        description="package information", default=[]
```
DiamondJoseph, line 172:
```suggestion
    package_info: list[RepositoryStatus] = Field(
        description="package information", default_factory=list
```

Else the same empty list gets passed as a default to all instances

src/blueapi/cli/scratch.py
```diff
@@ -128,3 +129,30 @@ def _validate_directory(path: Path) -> None:
...
+        repo = Repo(local_directory)
+        try:
+            branch = repo.active_branch.name
+        except TypeError:
+            branch = repo.head.commit.hexsha
+
+        is_dirty = repo.is_dirty()
+        scratch_responses.package_info.append(
+            RepositoryStatus(
+                repository_name=repo.remotes.origin.url,
```
DiamondJoseph, line 152: repository name and origin remote url aren't quite synonymous, is this the right name?

What if the repository does not have a remote named origin? It's possible to delete remotes, even origin, for example if you clone with https and then need to add the ssh endpoint as another remote (as I have done too often...)

What does `repo.description` give you, or `repo.remotes[0]`? What if there are multiple remotes?

helm/blueapi/values.yaml
```diff
@@ -119,7 +119,3 @@ initContainer:
     repositories: []
     # - name: "dodal"
     #   remote_url: https://github.com/DiamondLightSource/dodal.git
-
-# Mount path for scratch area from host machine, setting
-# this effectively enables scratch area management
-scratchHostPath: "" # example: /usr/local/blueapi-software-scratch
```
DiamondJoseph, line 125: Now that you've removed this value and made all the conditionals `{{- if .Values.initContainer.scratch.root }}`, if someone tried to run up the Helm chart with the default values `helm install blueapi ghcr://blueapi`, it will fail unless `/blueapi-plugins/scratch` is a path on the filesystem their Kubernetes mounts. 

One option is to add an `enabled` to `initContainer` and use that as the conditional check- this allows the self-documenting default to remain while not becoming a requirement of the minimal checkout- I think the Helm chart should always be in a state that `helm install blueapi` without any arguments or values file should work.

Make sure to also modify the documentation
https://diamondlightsource.github.io/blueapi/main/how-to/edit-live.html#kubernetes

DiamondJoseph (changes requested):

src/blueapi/service/model.py
```diff
@@ -174,6 +174,25 @@ class EnvironmentResponse(BlueapiBaseModel):
...
+    version: str = Field(description="Branch or tag of the repository")
+    is_dirty: bool = Field(description="Does the repository have uncommitted changes")
+
+
+class ScratchResponse(BlueapiBaseModel):
+    """
+    State of the scratch area.
+    """
+
+    package_info: list[RepositoryStatus] = Field(
```
callumforrester, line 188: Nit: From the name this sounds like a single object rather than a list of objects

```suggestion
    packages: list[RepositoryStatus] = Field(
```

src/blueapi/service/model.py
```diff
@@ -174,6 +174,25 @@ class EnvironmentResponse(BlueapiBaseModel):
...
+        description="package information", default_factory=list
+    )
+    installed_packages: list[str] = Field(
+        description="List of installed packages", default_factory=list
+    )
```
callumforrester, line 193: Could: Do we need to keep these pieces of information separate? Is the set of repositories ever going to be different to the list of installed packages?

src/blueapi/service/model.py
```diff
@@ -174,6 +174,25 @@ class EnvironmentResponse(BlueapiBaseModel):
     )
 
 
+class RepositoryStatus(BlueapiBaseModel):
+    remote_url: str = Field(description="URL of a repository on a remote server")
+    version: str = Field(description="Branch or tag of the repository")
+    is_dirty: bool = Field(description="Does the repository have uncommitted changes")
+
+
+class ScratchResponse(BlueapiBaseModel):
```
callumforrester, line 183: Should: Include whether scratch area is enabled at all (i.e. if scratch config is not `None`), might be scope creep in which case feel free to make a ticket.

src/blueapi/service/main.py
```diff
@@ -420,6 +421,12 @@ def set_state(
     return runner.run(interface.get_worker_state)
 
 
+@router.get("/scratch", response_model=ScratchResponse)
+@start_as_current_span(TRACER)
+def get_scratch_packages(runner: WorkerDispatcher = Depends(_runner)):
+    return runner.run(interface.get_scratch)
```
callumforrester, line 427: Must: I don't think we need to interrogate the scratch area from inside the subprocess, if you do the logic you've put in `interface.py` here instead would it change anything?

src/blueapi/service/main.py
```diff
@@ -420,6 +421,12 @@ def set_state(
     return runner.run(interface.get_worker_state)
 
 
+@router.get("/scratch", response_model=ScratchResponse)
+@start_as_current_span(TRACER)
+def get_scratch_packages(runner: WorkerDispatcher = Depends(_runner)):
```
callumforrester, line 426: Must: Docstring because it's part of the API

docs/reference/openapi.yaml
callumforrester, line 1: Must: Bump API version (Hopefully https://github.com/DiamondLightSource/blueapi/issues/531https://github.com/DiamondLightSource/blueapi/issues/531 will mean we are automatically prompted to do so)

src/blueapi/service/main.py
callumforrester: @ZohebShaikh you mentioned this to me and thinking about it: nevermind. The subprocess' job is to make managing the scratch area easier so the logic probably belongs in there, can leave this as-is.

src/blueapi/service/model.py
ZohebShaikh: This is more like `pip freeze` and `packages `is like the git repositories

callumforrester: Yeah, that sounds good! Maybe a `source=scratch` style query rather than a bool, for future proofing if we introduce more sources later

src/blueapi/service/model.py
```diff
@@ -169,6 +170,37 @@ class EnvironmentResponse(BlueapiBaseModel):
     )
 
 
+class SourceInfo(str, Enum):
+    pypi = "pypi"
+    scratch = "scratch"
```
callumforrester, line 175: Nit: Make the same as other enums in the API

```suggestion
    PYPI = "pypi"
    SCRATCH = "scratch"
```

src/blueapi/service/model.py
```diff
@@ -169,6 +170,37 @@ class EnvironmentResponse(BlueapiBaseModel):
...
+
+class ScratchResponse(BlueapiBaseModel):
+    """
+    State of the scratch area.
+    """
+
+    installed_packages: list[PackageInfo] = Field(
+        description="List of installed packages", default_factory=list
+    )
+    enabled: bool = Field(description="Scratch area state", default=False)
```
callumforrester, line 201: Should: Since it's no longer just about the scratch area I think there's some renaming needed here:

```suggestion
class PythonEnvironmentResponse(BlueapiBaseModel):
    """
    State of the scratch area.
    """

    installed_packages: list[PackageInfo] = Field(
        description="List of installed packages", default_factory=list
    )
    scratch_enabled: bool = Field(description="Scratch area state", default=False)
```

Should also rename the REST endpoint

callumforrester (changes requested):
Unsure why the docs build is broken, any ideas?

callumforrester (approved):
LGTM! Thanks for all your work on this, also note that once it is merged #744 will be much easier to do

src/blueapi/cli/scratch.py
```diff
@@ -128,3 +131,95 @@ def _validate_directory(path: Path) -> None:
...
+
+    if config is None:
+        python_env_response = PythonEnvironmentResponse(scratch_enabled=False)
+    else:
+        python_env_response = PythonEnvironmentResponse(scratch_enabled=True)
+        try:
+            _validate_directory(config.root)
+        except Exception as e:
+            logging.error(f"Failed to get scratch info: {e}")
+            return python_env_response
```
DiamondJoseph, line 178: nit: If there's an exception, the caller [this method is called from the rest `interface.py`] won't be informed if they don't check the logs.

I think we should allow the Exception to escape here, and return a non-200 status code.

DiamondJoseph (approved):

PR #863, opened 2025-03-18
Pass config to schema CLI command

tpoliaw:
Lets the schema be generated with OIDC configuration present.


callumforrester:
I remain extremely nervous about the schema being different in different cases like this

DiamondJoseph:
Alternatively, #889

PR #864, opened 2025-03-18
Add SkipJsonSchema annotations to optional fields

tpoliaw:
Prevents 'null' being included as a possible type in the generated
schema which interferes with the generated Java client.

The 'detail' and 'worker' entries on response types were flagged as an
error in the schema when the client was generated.

### DIFF
--- a/docs/reference/openapi.yaml
+++ b/docs/reference/openapi.yaml
@@ -340,7 +340,7 @@ components:
       type: object
 info:
   title: BlueAPI Control
-  version: 0.0.9
+  version: 0.0.10
 openapi: 3.1.0
 paths:
   /config/oidc:
--- a/docs/reference/openapi.yaml
+++ b/docs/reference/openapi.yaml
@@ -44,12 +44,10 @@ components:
           title: Environment Id
           type: string
         error_message:
-          anyOf:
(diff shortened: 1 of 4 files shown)


docs/reference/openapi.yaml
callumforrester, line 1: Must: Version bump

callumforrester (dismissed):

DiamondJoseph (approved):

src/blueapi/service/main.py
```diff
@@ -392,8 +393,8 @@ def get_state(runner: WorkerDispatcher = Depends(_runner)) -> WorkerState:
     "/worker/state",
     status_code=status.HTTP_202_ACCEPTED,
     responses={
-        status.HTTP_400_BAD_REQUEST: {"detail": "Transition not allowed"},
-        status.HTTP_202_ACCEPTED: {"detail": "Transition requested"},
+        status.HTTP_400_BAD_REQUEST: {},
+        status.HTTP_202_ACCEPTED: {},
```
tpoliaw, line 397: These (and ln318) were raised as problems by the java code gen but aren't directly related to the PR. Should they be split and/or changed to use `description` which I think might work?

Issue #646, opened 2024-09-20
Need to provide scratch root twice in `values.yaml`


DominicOram:
From testing on i04 it was found that to get `blueapi` to pick up the scratch directory it had to be provided in `worker` and in `initContainer` e.g.
```
worker:
  stomp:
    host: i04-control.diamond.ac.uk
  env:
    events:
      broadcast_status_events: False
    sources:
      - kind: dodal
        module: dodal.beamlines.i04
      - kind: planFunctions
        module: mx_bluesky.beamlines.i04
  scratch:
    root: /blueapi-plugins/scratch
initContainer:
  scratch:
    root: /blueapi-plugins/scratch
    repositories:
      - name: "dodal"
        remote_url: https://github.com/DiamondLightSource/dodal.git
      - name: "mx_bluesky"
        remote_url: https://github.com/DiamondLightSource/mx_bluesky.git
scratchHostPath: /dls_sw/i04/software/bluesky/scratch
```
This is annoying redundancy but worse is that it does not match the example given in https://github.com/DiamondLightSource/blueapi/blob/main/helm/blueapi/values.yaml

## Acceptance Criteria
* At least, the example is update to say match what is required
* Ideally I only need to provide the path once

stan-dot:
can confirm this also happens at i18 https://argocd.diamond.ac.uk/applications/i18-beamline/i18-blueapi?view=tree&resource=&node=argoproj.io%2FApplication%2Fi18-beamline%2Fi18-blueapi%2F0&tab=parameters

stan-dot:
@DominicOram I'm going to try the [release 0.10.0](https://github.com/DiamondLightSource/blueapi/releases/tag/0.10.0) at i18 and to remove the duplication. I will report back if this works now. 

From a look at the code, it should work now

stan-dot:
@DominicOram this works fine now, see the update here - https://gitlab.diamond.ac.uk/controls/containers/beamline/i18-services


does this resolve your issue ?

DiamondJoseph:
Resolved by #887

PR #871, opened 2025-03-25
Write python tests for helm chart rendering

callumforrester:
Possible fix for #559, includes xfailed tests for #646, test helm chart rendering with normal pytests, as suggested by [helm-unittest README](https://github.com/helm-unittest/helm-unittest). By no means exhaustive but a good starting point to build on.

Also include `helm` in the devcontainer.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     graphviz \
     && rm -rf /var/lib/apt/lists/*
 
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3; \
+    chmod 700 get_helm.sh; \
+    ./get_helm.sh; \
+    rm get_helm.sh
+
 # Set up a virtual environment and put it in PATH
 RUN python -m venv /venv
 ENV PATH=/venv/bin:$PATH
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     graphviz \
     && rm -rf /var/lib/apt/lists/*
 
+# Install helm for the dev container. This is the recommended 
+# approach per the docs: https://helm.sh/docs/intro/install
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3; \
     chmod 700 get_helm.sh; \
     ./get_helm.sh; \
--- a/tests/unit_tests/test_helm_chart.py
+++ b/tests/unit_tests/test_helm_chart.py
@@ -101,14 +101,16 @@ def test_helm_chart_creates_config_map(worker_config: ApplicationConfig):
     [
         {
             "initContainer": {
+                "enabled": True,
                 "scratch": {
                     "repositories": [],
                     "root": "/blueapi-plugins/scratch",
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -8,6 +8,11 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     graphviz \
     && rm -rf /var/lib/apt/lists/*
 
+RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3; \
```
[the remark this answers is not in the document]
callumforrester: We don't support the non-devcontainer use case

tests/unit_tests/test_helm_chart.py
```diff
@@ -0,0 +1,228 @@
...
+def test_init_container_resources_overridable():
+    manifests = render_chart(
+        values={
+            "resources": HIGH_RESOURCES,
+            "initResources": LOW_RESOURCES,
+            "scratchHostPath": "/foo",
+        }
+    )
+    assert (
+        manifests["blueapi"]["spec"]["template"]["spec"]["initContainers"][0][
```
DiamondJoseph, line 179: The Service, Ingress, StatefulSet all have  `name: {{ include "blueapi.fullname" . }}`- is `service.enabled` false by default?

tests/unit_tests/test_helm_chart.py
```diff
@@ -0,0 +1,228 @@
+import subprocess
```
DiamondJoseph, line 1: Shame there is no well maintained Helm python client

callumforrester: Ultimately helm itself is a client, so the only way to do a Python version other than this is to reimplement it

tests/unit_tests/test_helm_chart.py
callumforrester: Grouped rendered manifests by kind and then by name

tests/unit_tests/test_helm_chart.py
```diff
@@ -0,0 +1,307 @@
...
+)
+
+BLUEAPI_HELM_CHART = Path(__file__).parent.parent.parent / "helm" / "blueapi"
+Values = Mapping[str, Any]
+ManifestKind = Literal[
+    "ConfigMap",
+    "Ingress",
+    "Service",
+    "StatefulSet",
+    "Deployment",
```
DiamondJoseph, line 28:
```suggestion
    "Pod",
```

We do not produce any Deployments, but can make Pods from the test-connection

DiamondJoseph (approved):

Dockerfile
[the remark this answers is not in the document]
abbiemery: I just came here as someone who doesn't typically use dev containers to find out why main was not longer passing tests.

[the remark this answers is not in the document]
callumforrester: I should clarify based on an IRL discussion with @DiamondJoseph: The Python copier template is generally getting more opinionated over time because it is aimed at engineers who care more about having a cohesive set of tools that (have been tested to) work well together. We strongly encourage engineers/teams to use devcontainers because it is part of that tested group, but we don't mandate it. If you want to use something else that's okay but you're on your own and expected to figure things out for yourself.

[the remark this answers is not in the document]
callumforrester: @coretl I think we only had this discussion in-person, but should be an ADR on the copier template

PR #872, opened 2025-03-26
583 support graylog

dan-fernandes:
Implements structured logging to Graylog

Implements logging to Graylog at the python level rather than previous Kubernetes-level. Includes structured data such as beamline in logs. Disables export of graylog logs through kubernetes.


dan-fernandes:
Tests that are currently failing are in part due to how tracing is implemented. I am discussing with Keith how we want to mitigate the produced warnings.

helm/blueapi/Chart.yaml
```diff
@@ -17,3 +17,6 @@ type: application
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 # This is updated automatically by CI, the appVersion is automatically inserted by CI
 version: 0.1.0
+
+annotations:
+  fluentd-ignore: "true"
```
DiamondJoseph, line 22: This should be in the statefulset.yaml as an annotation on the pod: search for where we use  `restartOnConfigChange` in that file. See https://dev-portal.diamond.ac.uk/guide/kubernetes/tutorials/logging/#ignoring-logs

helm/blueapi/values.yaml
```diff
@@ -91,6 +91,14 @@ tracing:
       host: https://daq-services-jaeger # replace with central instance
       port: 4318
 
+
+logging:
+    graylog_export_enabled: True,
```
DiamondJoseph, line 96:
```suggestion
    graylog_enabled: False,
```

src/blueapi/config.py
```diff
@@ -67,6 +67,9 @@ class EnvironmentConfig(BlueapiBaseModel):
 
 class LoggingConfig(BlueapiBaseModel):
     level: LogLevel = "INFO"
+    graylog_export_enabled: bool = True
```
DiamondJoseph, line 70:
```suggestion
    graylog_enabled: bool = False
```

DiamondJoseph: If you don't have a local graylog set up (which I think is a sensible assumption when someone first clones the repository) I would have it disabled.

helm/blueapi/values.yaml
DiamondJoseph: This will be overriden in production but I think the blueapi chart should always be able to run without any infrastructure, e.g. if I `helm install blueapi` on minikube on my laptop at a conference.

src/blueapi/log.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class BeamlineTagFilter(logging.Filter):
+    """Filter to attach beamline attribute to LogRecords.
+
+    Attaches attribute `beamline` to all LogRecords that are passed through. Value is
+    taken from `BEAMLINE` env var, defaulting to `dev`.
+    """
+
+    beamline: str | None = os.environ.get("BEAMLINE")
```
DiamondJoseph, line 16: We need to support BEAMLINE or INSTRUMENT and should be trying to standardise on instrument.

src/blueapi/log.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def filter(self, record):
```
DiamondJoseph, line 18:
```suggestion
    def filter(self, record: LogRecord) -> bool:
```

presumably?

src/blueapi/log.py
DiamondJoseph, line 16:
```suggestion
    instrument: str = os.environ.get("BEAMLINE", "dev")
```

src/blueapi/log.py
```diff
@@ -0,0 +1,110 @@
...
+        record.beamline = self.beamline if self.beamline else "dev"
```
DiamondJoseph, line 19:
```suggestion
        record.instrument = self.instrument
        record.beamline = self.instrument
```

src/blueapi/log.py
```diff
@@ -0,0 +1,110 @@
...
+        return True
+
+
+class InstrumentTagFilter(logging.Filter):
```
DiamondJoseph, line 23: Ah, I see this is being done down here. Is there reason not to do both in the same Filter?

src/blueapi/log.py
```diff
@@ -0,0 +1,110 @@
...
+    if logging_config.graylog_export_enabled:
+        graylog_handler = set_up_graylog_handler(logger, logging_config)
+        handlers.append(graylog_handler)
+
+    filters = [
+        BeamlineTagFilter(),
+        InstrumentTagFilter(),
+    ]
+
+    add_all_filters_to_all_handlers(handlers, filters)
```
DiamondJoseph, line 64: nit: I would construct the filters first and pass them into each of the configure_handler methods, rather than having a method to apply filters to handlers.
This would allow you to remove the handlers list

src/blueapi/service/main.py
```diff
@@ -450,6 +452,13 @@ async def add_api_version_header(
 ):
     response = await call_next(request)
     response.headers["X-API-Version"] = REST_API_VERSION
+    LOGGER.info(
+        '%s:%s - "%s %s"',
+        request.client.host,
+        request.client.port,
+        request.method,
+        request.url.path,
+    )
```
DiamondJoseph, line 461: should: move this to its own function and call it from a `Depends()` same as this function is- this isn't to do with adding the api_version_header, it's "log_request_details" or similar.

DiamondJoseph (changes requested):

helm/blueapi/values.yaml
dan-fernandes: Implemented. Both having False as default, and that I prefer `graylog_enabled` as a var name.

src/blueapi/config.py
dan-fernandes: Implemented.

src/blueapi/log.py
dan-fernandes: Implemented on both filters.

src/blueapi/log.py
dan-fernandes: Implemented

src/blueapi/log.py
dan-fernandes: Aggregated both filters into one.

src/blueapi/log.py
dan-fernandes: Implemented

src/blueapi/log.py
dan-fernandes: Implemented.

src/blueapi/service/main.py
dan-fernandes: Refactored into a separate middle ware function.

helm/blueapi/Chart.yaml
dan-fernandes: Implemented!

helm/blueapi/values.yaml
```diff
@@ -91,6 +91,14 @@ tracing:
...
+    graylog_host: "graylog-log-target.diamond.ac.uk"
+    graylog_port: 12232
+    level: "INFO"
```
callumforrester, line 99: Must: This needs to go under `worker` if you want blueapi to import it

helm/blueapi/templates/statefulset.yaml
```diff
@@ -12,6 +12,9 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if ((.Values.worker).logging).graylog_enabled }}
+      fluentd-ignore: "true"
+      {{- end }}
```
callumforrester, line 17: Must: Indentation

```suggestion
      {{- if ((.Values.worker).logging).graylog_enabled }}
        fluentd-ignore: "true"
      {{- end }}
```

src/blueapi/cli/cli.py
```diff
@@ -56,10 +57,9 @@ def main(ctx: click.Context, config: Path | None | tuple[Path, ...]) -> None:
     ctx.ensure_object(dict)
     loaded_config: ApplicationConfig = config_loader.load()
 
+    set_up_logging(loaded_config.logging or LoggingConfig())
```
callumforrester, line 60: Could: I don't think `loaded_config.logging` can ever be `None` so do you need the `or`?

helm/blueapi/values.yaml
callumforrester, line 96: Could: Disable it by default, requires a config change to beamlines but makes it log as-normal out-of-box, @DiamondJoseph what do you think?

callumforrester (changes requested):

helm/blueapi/values.yaml
DiamondJoseph: @callumforrester agreed, see my comment on the same line, I believe Dan has switched it to False by default but hasn't pushed?

src/blueapi/log.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class InstrumentTagFilter(logging.Filter):
+    """Filter to attach instrument and beamline attributes to LogRecords.
+
+    Attaches attributes `instrument` and `beamline` to all LogRecords that are passed
+    through. Value is taken from `BEAMLINE` env var, defaulting to `dev`.
+    """
+
+    instrument: str | None = os.environ.get("BEAMLINE", "dev")
```
DiamondJoseph, line 16:
```suggestion
    instrument: str = os.environ.get("BEAMLINE", os.environ.get("INSTRUMENT", "dev"))
```

Gets us:
1. `$BEAMLINE` if it exists
2. `$INSTRUMENT` if if exists
3. "dev"

Which therefore is never None and we can tighten the bounds on this.

src/blueapi/service/main.py
```diff
@@ -453,6 +456,16 @@ async def add_api_version_header(
     return response
 
 
+async def log_request_details(
+    request: Request, call_next: Callable[[Request], Awaitable[Response]]
+) -> Response:
+    response = await call_next(request)
+    LOGGER.info(
+        "http request: %s call_next: %s response: %s", request, call_next, response
```
DiamondJoseph, line 464: call_next is the next middleware function that is going to be called, it's an implementation detail of the server and not something that needs to be logged.

I would just [some of the fields of request](https://requests.readthedocs.io/en/latest/api/#lower-lower-level-classes)

body
headers (? maybe, don't want to leak authorization header)
method
path_url
url

If path_url and url look sufficiently different

src/blueapi/cli/cli.py
dan-fernandes: Or removed.

helm/blueapi/values.yaml
dan-fernandes: Would this literally be copied and indented under `worker`:

src/blueapi/service/main.py
dan-fernandes: Implemented!

src/blueapi/log.py
dan-fernandes: Implemented

helm/blueapi/values.yaml
dan-fernandes: Re-read this morning and I understand what you mean now - I have moved the above yaml into the worker scope.

src/blueapi/service/main.py
dan-fernandes: I have gone ahead and added a basic unit test, to have the middleware exercised.

src/blueapi/worker/task_worker.py
```diff
@@ -138,6 +141,7 @@ def __init__(
         self._broadcast_statuses = broadcast_statuses
         self._current_task_otel_context = None
         setup_tracing("BlueAPIWorker", OTLP_EXPORT_ENABLED)
+        set_up_logging(LoggingConfig())
```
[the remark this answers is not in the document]
dan-fernandes: Moved to interface, which has access to the correct config.

helm/blueapi/values.yaml
```diff
@@ -112,6 +112,11 @@ worker:
       password: guest
     host: rabbitmq
     port: 61613
+  logging:
+      graylog_export_enaTrue: False,
```
DiamondJoseph, line 116:
```suggestion
      graylog_enabled: False,
```

dan-fernandes: Fixed

src/blueapi/service/main.py
```diff
@@ -480,6 +483,16 @@ async def add_api_version_header(
     return response
 
 
+async def log_request_details(
+    request: Request, call_next: Callable[[Request], Awaitable[Response]]
+) -> Response:
+    response = await call_next(request)
+    LOGGER.info(
+        "body: %s method: %s url: %s", request.body, request.method, request.url
+    )
```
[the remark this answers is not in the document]
dan-fernandes: Implemented

helm/blueapi/templates/statefulset.yaml
```diff
@@ -12,6 +12,9 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if ((.Values.worker).logging).graylog_enabled }}
+        fluentd-ignore: "true"
+      {{- end }}
```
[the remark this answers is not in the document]
DiamondJoseph: What do you mean by "works": puts the annotation onto the pod or disables container STDOUT being forwarded to graylog? I don't think we can feasibly test the later, and the former is testing whether conditionals work in the Go templating language

DiamondJoseph (approved):
Just one nit on the Helm chart

helm/blueapi/templates/statefulset.yaml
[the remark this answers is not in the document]
callumforrester: Yep, that's what I mean, they're unit tests, i.e. they test that it does what we think it should do

helm/blueapi/values.yaml
```diff
@@ -117,6 +117,11 @@ worker:
     repositories: []
     # - name: "dodal"
     #   remote_url: https://github.com/DiamondLightSource/dodal.git
+  logging:
+      graylog_enabled: False
+      graylog_host: "graylog-log-target.diamond.ac.uk"
+      graylog_port: 12232
```
callumforrester, line 123: Nit: May as well group rather than prefix

```suggestion
      graylog:
        enabled: False
        host: "graylog-log-target.diamond.ac.uk"
        port: 12232
```

tests/unit_tests/test_helm_chart.py
```diff
@@ -318,6 +318,35 @@ def test_worker_scratch_config_used_when_init_container_enabled():
...
+        values={
+            "worker": {
+                "logging": {
+                    "graylog_enabled": enabled,
+                },
+            },
+        }
+    )
+
+    if enabled:
```
callumforrester, line 333: Nit: I would personally write this as two tests, it's roughly the same number of lines of code and much easier to read, you can also extract the common lines into their own function.

callumforrester (approved):
Looks good, just nits :)

helm/blueapi/values.yaml
dan-fernandes: Implemented

helm/blueapi/templates/statefulset.yaml
```diff
@@ -12,6 +12,9 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if (.Values.worker.logging).graylog.enabled }}
```
DiamondJoseph, line 15:
```suggestion
      {{- if ((.Values.worker.logging).graylog).enabled }}
```

DiamondJoseph (approved):

helm/blueapi/templates/statefulset.yaml
dan-fernandes: Implemented

PR #873, opened 2025-03-26
Prevent Enum classes being treated as Bluesky Protocols

DiamondJoseph:
Enum defines a property for `name` and `value`- meaning that according to ducks, any enum class extends `HasName`. Now that we are checking the origin of types, we have started catching this when trying to import plans that take enums as a parameter and possibly even just on devices that have enums as a type arg (e.g. a Movable[ShutterDemand]).

This change makes HasName, HasParent and HasHints on their own no longer sufficient to define that  a class is a Device class.

This was making blueapi try and treat Enum classes as Device classes, and so constructing References which extend them- extending an Enum class is not allowed.

### DIFF
--- a/docs/reference/openapi.yaml
+++ b/docs/reference/openapi.yaml
@@ -269,7 +269,7 @@ components:
       type: object
 info:
   title: BlueAPI Control
-  version: 0.0.5
+  version: 0.0.6
 openapi: 3.1.0
 paths:
   /devices:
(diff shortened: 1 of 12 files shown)


callumforrester (approved):
My main worry with this is unknown knock-on effects, hopefully system testing will catch those...

callumforrester (approved):

PR #876, opened 2025-03-27
Add cors settings to configuration

### LINKED ISSUE #738: Add CORS settings
When talking to the REST API from a local JS application, Firefox blocks the requests from 127.0.0.1 unless they are returned with an appropriate CORS header

## Acceptance Criteria
- Some flag or config variable allows setting allowed origins and adding the `CORSMiddleware` to the fastAPI app, including an option for allowing all origins for testing

noemifrisina:
Adds the possibility to pass CORS setting to the configuration and enable them to the FastAPI app, to be able to talk to the REST API from a local web application.

Closes #738


noemifrisina:
I've been trying to add to the tests to make codecov happier, but I admittedly got a bit lost when trying to test `get_app`. Any suggestions?

src/blueapi/config.py
```diff
@@ -166,6 +173,7 @@ class ApplicationConfig(BlueapiBaseModel):
     scratch: ScratchConfig | None = None
     oidc: OIDCConfig | None = None
     auth_token_path: Path | None = None
+    cors: CORSConfig | None = None
```
callumforrester, line 176: Should: I think cors config should go under `api` rather than at the top level, it's not associated with any other part of the application.

callumforrester (changes requested):
@noemifrisina For tests I guess you want to modify or make your own version of [this fixture](https://github.com/DiamondLightSource/blueapi/blob/06500d6611087d062c9750e0eeee4f5f608004f4/tests/unit_tests/service/test_rest_api.py#L51) with cors config added, then send requests against it and inspect the headers?

src/blueapi/config.py
noemifrisina: Yeah, that makes sense.

DiamondJoseph (approved):

callumforrester (approved):
LGTM :)

PR #877, opened 2025-03-28
Fix API versioning tests

DiamondJoseph:


ZohebShaikh (approved):

callumforrester (approved):
Thanks for undoing my foolishness

.github/workflows/_test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           fetch-depth: 0
 
       - name: "fetch-main-branch"
-        run: "git branch --track main origin/main"
+        run: "git checkout --track -B main origin/main"
```
callumforrester, line 32: Must (too late): This checks out the branch rather than just ensures it is check-out-able, this might break people's tests, sorry @ZohebShaikh

callumforrester: To be clear, github action does not automatically clone the repo with `main` tracking `origin/main`, so when we do `git show main:<thing>` it breaks, so we just need to set up tracking, but not checkout

ZohebShaikh: No worries @callumforrester, I reviewed it as well and it did slip through the cracks

PR #880, opened 2025-03-28
Refresh helm

DiamondJoseph:
Re-applies blueapi specific configuration over the result of a fresh `helm create blueapi`, recovering where we had previously removed configuration that was thought not necessary.

Breaking changes to the helm chart:
- `ingress.create` is now `ingress.enabled`
- ingress now needs `tls.hosts` defined

Changes worth considering:
- Can define arbitrary volumes, volumeMounts
- Liveness/Readiness probes defined- should move to `/healthz` and make those routes unauthenticated


callumforrester:
Feels like this should really go in behind #871

.github/workflows/_helm.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Log in to GitHub Container Registry
+        if: github.ref_type == 'tag'
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ${{ env.CHART_REPO }} --username ${{ github.repository_owner }} --password-stdin
+
+      - name: Publish Helm chart to container registry
+        if: github.ref_type == 'tag'
+        run: |
+          helm dependencies update helm/blueapi
+          helm package helm/blueapi --app-version ${{ steps.meta.outputs.version }} -d /tmp/
```
DiamondJoseph, line 44:
```
(.venv) qwe67581@DIAMRL6121:~/Documents/work/blueapi/helm/blueapi$ helm show chart . | yq '.version'
"0.9.0"
(.venv) qwe67581@DIAMRL6121:~/Documents/work/blueapi/helm/blueapi$ helm show chart oci://ghcr.io/diamondlightsource/charts/blueapi | yq '.version'
Pulled: ghcr.io/diamondlightsource/charts/blueapi:0.8.0
Digest: sha256:c3077e2854462b1b5b753d0be4ff02fa07bba86db6764988a2631dbc73f2612b
"0.8.0"
```

TODO: auto-increment the patch version of the chart on appVersion change

helm/blueapi/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: blueapi
-description: A helm chart deploying a worker pod that runs Bluesky plans
+description: A Helm chart deploying a worker pod that runs Bluesky plans
```
callumforrester, line 3: Could: If you're going to edit this then just remove the helm chart part altogether, as a user I know it's a helm chart, that is redundant.

helm/blueapi/templates/configmap.yaml
```diff
@@ -13,8 +13,22 @@ kind: ConfigMap
 metadata:
   name: {{ include "blueapi.fullname" . }}-otel-config
 data:
-{{- if .Values.tracing.otlp.enabled | default false }}
+{{- if ((.Values.tracing).otlp).enabled }}
```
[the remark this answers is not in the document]
callumforrester: Given that they in the default `values.yaml`, I think their existence is guaranteed?

helm/blueapi/templates/ingress.yaml
```diff
@@ -1,22 +1,43 @@
...
-  - hosts:
-    - {{ required "A valid hostname must be provided" .Values.ingress.host }}
+    {{- range .Values.ingress.tls }}
+    - hosts:
+        {{- range .hosts }}
+        - {{ . | quote }}
+        {{- end }}
+      secretName: {{ .secretName }}
+    {{- end }}
+  {{- end }}
```
callumforrester, line 25: Should: Write a test to prove this still works and doesn't change the output, DLS ingress controller is a bit pedantic.

helm/blueapi/values.yaml
```diff
@@ -2,121 +2,189 @@
...
+  stomp:
+    auth:
+      username: guest
+      password: guest
+    host: rabbitmq
+    port: 61613
+
+initContainer:
+  enabled: false
+  config:
```
callumforrester, line 56: This is a breaking change, is it really necessary?

helm/blueapi/templates/statefulset.yaml
```diff
@@ -69,39 +56,83 @@ spec:
...
+              protocol: TCP
+          {{- with .Values.livenessProbe }}
+          livenessProbe:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.readinessProbe }}
+          readinessProbe:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.resources }}
```
callumforrester, line 84: Should: I think these were removed because they caused problems, are you trying to fix with #884?

callumforrester (dismissed):
I'll be honest, I'm not very happy about merging this. The _majority_ of helm changes we have merged have included bugs/regressions, we should have done something like #871 a long time ago. Most of the time we have done small, incremental changes to the helm chart so we could feel more confident, but even those broke. This, meanwhile, is a big change to a lot of high-risk areas. #871 is a good start but is not a comprehensive test suite. I think to merge this I will need:

- More tests
- An analysis of breaking changes
- A migration guide to go into the changelog
- Any areas of uncertainty we can explore on the test rigs

.github/workflows/_container.yml
```diff
@@ -16,44 +12,6 @@ jobs:
         with:
           # Need this to get version number from last tag
           fetch-depth: 0
-      - name: Validate SemVer2 version compliance
```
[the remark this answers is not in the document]
DiamondJoseph: It is no longer required, as the blueapi release tag is no longer the version of the Helm chart, and therefore does not require being a semantic version. If you want to enforce that the blueapi container release is a semver it can be restored, but this was added for issues publishing the Helm chart.

helm/blueapi/values.yaml
DiamondJoseph: In my opinion yes, config should be under its own key and not included with non-ApplicationConfig compatible keys, like enabled.

helm/blueapi/templates/ingress.yaml
DiamondJoseph: Wasn't that the purpose of #871?
Here is a git diff of before/after these changes with the ingress create/enabled. 
Annoyingly, github won't allow attaching yaml files

image

helm/blueapi/templates/statefulset.yaml
DiamondJoseph: Yes, that's the purpose of 884.

helm/blueapi/templates/hpa.yaml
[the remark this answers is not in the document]
DiamondJoseph: Not much, but I could see a use case if we suddenly decide to start running a bunch of simulated device beamlines, scaling out as they become busy (if our readiness probe started reflecting whether the pod had an active task perhaps). End of the day, it comes for free with your `helm create` and if we remove it, it's one more thing to remove every time we refresh `helm create` again. I don't think it harms anyone to have it available but turned off.

helm/blueapi/templates/statefulset.yaml
DiamondJoseph: #884 has now gone in, so the probes now get `/healthz` which returns a 200 code

helm/blueapi/values.yaml
DiamondJoseph: No longer required due to other changes to how the scratch is loaded.

helm/blueapi/templates/ingress.yaml
callumforrester: #871 wasn't meant to be an exhaustive set of tests, just a framework/starting point. Right now we are confident that our chart works because we've experienced it working in production, but changes are risky. The idea was that people would write new tests as and when they changed/added things to increase confident and reduce the need to test on the cluster over time. In other words: You should write a test for everything this PR touches that doesn't already have a test.

ZohebShaikh (changes requested):
The PR looks good overall! One suggestion that I believe would be helpful for developers using this Helm chart is to include a table in the documentation listing all the parameters—along with their names, descriptions, and default values. Something similar to what's done [here](https://artifacthub.io/packages/helm/bitnami/rabbitmq#common-parameters)

helm/blueapi/templates/ingress.yaml
DiamondJoseph: This change now adds tests that check the structure of the Ingress (and service) as the first commit, then as a final commit make the changes to the inputs (but not outputs!) that are expected, to prove that the new system outputs the same as the old system.
image

helm/blueapi/README.md
```diff
@@ -0,0 +1,58 @@
...
+| worker.env.sources[0] | object | `{"kind":"deviceFunctions","module":"blueapi.startup.example_devices"}` | modules (must be installed in the venv) to fetch devices/plans from |
+| worker.logging | object | `{"graylog":{"enabled":false,"url":"http://graylog-log-target.diamond.ac.uk:12232/"},"level":"INFO"}` | Configures logging |
+| worker.logging.graylog.url | string | `"http://graylog-log-target.diamond.ac.uk:12232/"` | This is the `dodal` input which will be renamed to blueapi |
+| worker.scratch | object | `{"repositories":[],"root":"/blueapi-plugins/scratch"}` | If initContainer is enabled the default branch of python projects in this section are installed into the venv *without their dependencies* |
+| worker.stomp | object | `{"auth":{"password":"guest","username":"guest"},"enabled":false,"url":"http://rabbitmq:61613/"}` | Message bus configuration for returning status to GDA/forwarding documents downstream |
+
+----------------------------------------------
+Autogenerated from chart metadata using [helm-docs v1.14.2](https://github.com/norwoodj/helm-docs/releases/v1.14.2)
```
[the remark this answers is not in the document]
DiamondJoseph: The linting will fail if the output should have changed, as it's run as a pre-commit hook

DiamondJoseph:
> As the pod will take time to startup due to scratch, I was just going through the docs to see if it starts the probe after the initContainer is done booting up

The probes are defined on the container level, so they not attempt to run until the initContainer has run to completion. We could have a startupProbe that is slightly more permissive in case we are slow to start up, but with the default resources and nothing in scratch startup takes ~4s. The probes runs every 10s and needs to fail 3 in a row to die, so our startup is fast enough to not fail too many- whether they're permissive enough to not die on a filesystem blip I'm not sure. We do probably want a startupProbe for the potential 10s to connect to devices, but I don't know if we start serving the API prior to the subprocess being ready?

> And the other was we can have a cookie created on the fly and pass it to the healthz probe as a header so that at the fastapi side we can just accept requests from that which has that header for the probable case of the healthz getting DDOS'ed and making blueapi go into boot-loop

It's a possibility, I think if we're going to add that it can be after 1.0.0, since it's non breaking. If the REST API is being hammered enough to freeze and kill the subprocess running a scan, someone is doing something wrong or malicious. After the document store we can consider re-architecting to extract the subprocess- then if the API goes down the scan is unaffected.

> I also didn't see anything about if the liveliness probe fails it trying to restart the pod

If the liveness probes fails failureThreshold times (default 3), the pod is killed and subject to its restartpolicy.

helm/blueapi/README.md
```diff
@@ -0,0 +1,54 @@
...
+| initContainer.persistentVolume.existingClaimName | string | `""` | May be set to an existing persistent volume claim to re-use the volume, else a new one is created for each blueapi release |
+| livenessProbe | object | `{"failureThreshold":3,"httpGet":{"path":"/healthz","port":"http"},"periodSeconds":10}` | Liveness probe, if configured kubernetes will kill the pod and start a new one if failed consecutively. This is automatically disabled when in debug mode. |
+| nameOverride | string | `""` |  |
+| nodeSelector | object | `{}` | May be required to run on specific nodes (e.g. the control machine) |
+| podAnnotations | object | `{}` |  |
+| podLabels | object | `{}` |  |
+| podSecurityContext | object | `{}` |  |
+| readinessProbe | object | `{"httpGet":{"path":"/healthz","port":"http"}}` | Readiness probe, if configured kubernetes will not route traffic to this pod if failed consecutively. This is automatically disabled when in debug mode. |
```
ZohebShaikh, line 31:
```suggestion
| readinessProbe | object | `{"httpGet":{"path":"/healthz","port":"http"}}` | Readiness probe, if configured kubernetes will not route traffic to this pod until the read condition is met. If failed for 10 time this will restart the pod. This is automatically disabled when in debug mode. |
```
The same `periodSeconds` and `failureThreshold` should be mentioned here

helm/blueapi/README.md
```diff
@@ -0,0 +1,54 @@
...
+| resources | object | `{}` | Sets the compute resources available to the pod. When using debug mode or an internal PVC we recommend limits.cpu: 2000m, limits.memory: 4000m requests must be >= 0.1*limits |
```
ZohebShaikh, line 32: Can we have `resourcesPreset` I think we will need only 2
1. default
2. debug-mode

helm/blueapi/values.yaml
```diff
@@ -2,129 +2,214 @@
...
-# runAsNonRoot: true
-# runAsUser: 1000
-
-# Recommended for production to change service.type to ClusterIP and set up an Ingress
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
```
ZohebShaikh, line 48:
```suggestion
  readOnlyRootFilesystem: true
  runAsNonRoot: true
  runAsUser: 1000
```

This is part of best practices, The only thing to look for is
if the scratch clone that is done is by the correct user

ZohebShaikh (commented):
It looks good overall, few minor change requests.

As this is all helm changes can we see this working with all the configuration before merge? Just for more confidence in helm changes

helm/blueapi/values.yaml
```diff
@@ -2,129 +2,214 @@
...
+# This is to override the chart name.
 nameOverride: ""
 fullnameOverride: ""
 
+# This section builds out the service account more information can be found here: https://kubernetes.io/docs/concepts/security/service-accounts/
 serviceAccount:
   # Specifies whether a service account should be created
   create: false
+  # Automatically mount a ServiceAccount's API credentials?
+  automount: true
```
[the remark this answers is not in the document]
DiamondJoseph:
```yaml
  # Automatically mount a ServiceAccount's API credentials?
```

That's how it is rendered when you do `helm create blueapi`, I'm not sure why

helm/blueapi/README.md
ZohebShaikh: The limits have been added,and the `resourcesPreset` is consider not important considering our scale

ZohebShaikh (approved):

PR #884, opened 2025-04-01
Add probe for k8s

DiamondJoseph:


docs/reference/openapi.yaml
```diff
@@ -381,6 +395,18 @@ paths:
                 $ref: '#/components/schemas/EnvironmentResponse'
           description: Successful Response
       summary: Get Environment
+  /healthz:
```
tpoliaw, line 398: why `healthz` instead of `health` or `status` or similar?

tpoliaw (commented):
Is it possible to mark some endpoints as not requiring authentication? As it is, the health endpoint currently requires auth headers which presumably k8s won't have (or will it?).

callumforrester:
Could you explain the motivation for this? Was k8s not killing the pod when it should have?

DiamondJoseph:
> Is it possible to mark some endpoints as not requiring authentication? As it is, the health endpoint currently requires auth headers which presumably k8s won't have (or will it?).

Yes, good catch- the auth_router (which is where the client get the OIDC well-known from the server) is on the auth_router, which (perhaps confusingly) doesn't require authentication- can move it to there or construct another router.

```
# app includes all routes on all routers 
    if config.oidc:
        dependencies = [Depends(verify_access_token(config.oidc))]
        app.include_router(auth_router)   # routes on this router do not require auth
    app.include_router(router, dependencies=dependencies)  # all routes on this router require auth
```

DiamondJoseph:
> Could you explain the motivation for this? Was k8s not killing the pod when it should have?

Probes are a default part of the helm chart (see #880): `/` on blueapi returns a 404, so using the default probes would have the server die. Additionally, having an explicit endpoint allows us to e.g. disable logging for the endpoint.

docs/reference/openapi.yaml
DiamondJoseph: Sorry meant to say that it was convention but apparently I just sent a link without explanation.

src/blueapi/service/main.py
```diff
@@ -449,7 +451,7 @@ def get_python_environment(
     return runner.run(interface.get_python_env, name, source)
 
 
-@router.get(
+@open_router.get(
```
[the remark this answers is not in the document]
DiamondJoseph: Yup, but if we're OK with returning 204/None from /config/oidc then we can enable it always. 
Or else, move to another router which is always constructed.

DiamondJoseph:
NB this is now based on top of #889 to allow the open_router to always be added to the app.

tpoliaw (approved):

PR #887, opened 2025-04-02
Update config map for worker if scratch enabled

### LINKED ISSUE #885: Helm chart does not propagate scratch config to blueapi container
Following #861 I believe https://github.com/DiamondLightSource/blueapi/issues/646 is fixed, but a new bug has been introduced: The chart generates `scratch` configuration, but only propagates it into the `initContainer`. It does not propagate that config into the main container, the Python environment endpoint uses that config to determine which packages are in the scratch area, so it will never report any packages in scratch in the present state.

## Steps To Reproduce
1. Deploy to kubernetes with a configured scratch area containing at least one package
2. Poll the python environment endpoint with `curl` or similar
3. Confirm blueapi does not believe there are any packages in the scratch area

As an aside, no more helm changes are going in without https://github.com/DiamondLightSource/blueapi/pull/871, we need tests for that area.

## Acceptance Criteria
- Packages in the scratch area are reported
- There are regression tests to prevent a re-emergence of this and #646

ZohebShaikh:
Closes #885

### DIFF
--- a/docs/how-to/edit-live.md
+++ b/docs/how-to/edit-live.md
@@ -54,10 +54,23 @@ The helm chart can be configured to mount a scratch area from the
 host machine, include the following in your `values.yaml`:
 
 ```yaml
+  worker:
+    scratch:
+      root: /path/to/my/scratch/directory
+      required_gid: 12345
+      repositories:
+        - name: dodal
+          remote_url: https://github.com/DiamondLightSource/dodal.git
+        - name: mx-bluesky
+          remote_url: https://github.com/DiamondLightSource/mx-bluesky.git
+
   initContainer:
     enabled: true
 ```
-When initContainer is enabled worker must have scratch config.
+
+When `initContainer` is enabled, it will iterate over the scratch configuration, cloning the specified repositories and setting the correct group ownership based on the `required_gid`. This ensures that the scratch area is properly initialized before the main container starts.
+
+If `initContainer` is not enabled, the main container will still attempt to load the scratch area as specified in the configuration. However, it will not perform any cloning or group setup, so the scratch area must already be prepared and correctly configured.
 
 The scratch folder that you're pointing to must exist, not already have a copy of the repositories that will be cloned into it and have correct permissions e.g.
 
--- a/docs/how-to/edit-live.md
+++ b/docs/how-to/edit-live.md
@@ -56,11 +56,8 @@ host machine, include the following in your `values.yaml`:
 ```yaml
   initContainer:
     enabled: true
-    scratch:
-      root: path/to/scratch/area  # e.g. /dls_sw/<my_beamline>/software/blueapi/scratch
-      repositories: []
 ```
-
+When initContainer is enabled worker must have scratch config.
 
 The scratch folder that you're pointing to must exist, not already have a copy of the repositories that will be cloned into it and have correct permissions e.g.
 
--- a/helm/blueapi/templates/configmap.yaml
+++ b/helm/blueapi/templates/configmap.yaml
@@ -5,8 +5,14 @@ metadata:
 data:
   config.yaml: |-
     {{- toYaml .Values.worker | nindent 4 }}
-
+    {{- if .Values.initContainer.enabled }}
+    scratch:
+      root: {{ .Values.initContainer.scratch.root }}
+      repositories: {{ .Values.initContainer.scratch.repositories }}
+    {{- end }}
 ---
+
+
 
 apiVersion: v1
 kind: ConfigMap
--- a/helm/blueapi/templates/configmap.yaml
+++ b/helm/blueapi/templates/configmap.yaml
@@ -7,8 +7,7 @@ data:
     {{- toYaml .Values.worker | nindent 4 }}
     {{- if .Values.initContainer.enabled }}
     scratch:
-      root: {{ .Values.initContainer.scratch.root }}
-      repositories: {{ .Values.initContainer.scratch.repositories }}
+      {{- toYaml .Values.initContainer.scratch | nindent 6 }}
     {{- end }}
 ---
 
--- a/helm/blueapi/templates/configmap.yaml
+++ b/helm/blueapi/templates/configmap.yaml
@@ -5,6 +5,7 @@ metadata:
 data:
   config.yaml: |-
     {{- toYaml .Values.worker | nindent 4 }}
+
 ---
 
 apiVersion: v1
--- a/helm/blueapi/templates/configmap.yaml
+++ b/helm/blueapi/templates/configmap.yaml
@@ -5,10 +5,6 @@ metadata:
 data:
   config.yaml: |-
     {{- toYaml .Values.worker | nindent 4 }}
-    {{- if .Values.initContainer.enabled }}
-    scratch:
-      {{- toYaml .Values.initContainer.scratch | nindent 6 }}
-    {{- end }}
 ---
 
 apiVersion: v1
--- a/helm/blueapi/templates/configmap.yaml
+++ b/helm/blueapi/templates/configmap.yaml
@@ -12,8 +12,6 @@ data:
     {{- end }}
 ---
 
-
-
 apiVersion: v1
 kind: ConfigMap
 metadata:
--- a/helm/blueapi/templates/init-configmap.yaml
+++ b/helm/blueapi/templates/init-configmap.yaml
@@ -1,7 +1,10 @@
+{{- if .Values.initContainer.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ include "blueapi.fullname" . }}-initconfig
 data:
   initconfig.yaml: |-
-    {{- toYaml .Values.initContainer | nindent 4 }}
+      {{- toYaml .Values.worker.scratch | nindent 4 }}
+---
+{{- end }}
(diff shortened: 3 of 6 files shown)


tests/unit_tests/test_helm_chart.py
```diff
@@ -305,3 +293,71 @@ def group_manifests(ungrouped: Iterable[Mapping[str, Any]]) -> GroupedManifests:
...
+    )
+
+    config = yaml.safe_load(
+        manifests["ConfigMap"]["blueapi-config"]["data"]["config.yaml"]
+    )
+
+    assert config.get("scratch", None) is None
+
+
+def test_init_container_config_copied_to_worker_when_enabled():
```
callumforrester, line 342: Could: Also have a test that proves the scratch config is appended, not written over, the worker config

tests/unit_tests/test_helm_chart.py
```diff
@@ -305,3 +293,71 @@ def group_manifests(ungrouped: Iterable[Mapping[str, Any]]) -> GroupedManifests:
             )
         group[name] = manifest
     return groups
+
+
+def test_init_container_config_not_available_when_disabled():
```
callumforrester, line 298: Should: Split into 2 tests (or parametrize)

tests/unit_tests/test_helm_chart.py
```diff
@@ -305,3 +293,71 @@ def group_manifests(ungrouped: Iterable[Mapping[str, Any]]) -> GroupedManifests:
...
+            ]
+        )
+        is None
+    )
+
+    config = yaml.safe_load(
+        manifests["ConfigMap"]["blueapi-config"]["data"]["config.yaml"]
+    )
+
+    assert config.get("scratch", None) is None
```
callumforrester, line 320: Should: Be specific, either you're expecting an empty scratch block or no scratch block. The test should fail if you get the one you're not expecting, so

```suggestion
    assert "scratch" not in config
```

Or 

```suggestion
    assert config["scratch"] is None
```

I think the first one?

tests/unit_tests/test_helm_chart.py
```diff
@@ -305,3 +293,71 @@ def group_manifests(ungrouped: Iterable[Mapping[str, Any]]) -> GroupedManifests:
...
+            "initContainers"
+        ]
+        is None
+    )
+
+    config = yaml.safe_load(
+        manifests["ConfigMap"]["blueapi-config"]["data"]["config.yaml"]
+    )
+
+    assert config.get("scratch", None) is None
```
callumforrester, line 339: Should: See above

tests/unit_tests/test_helm_chart.py
callumforrester, line 298: Nit: The new tests should go above the utility functions like `render_chart` et al.

helm/blueapi/templates/configmap.yaml
```diff
@@ -5,7 +5,10 @@ metadata:
 data:
   config.yaml: |-
     {{- toYaml .Values.worker | nindent 4 }}
```
callumforrester, line 7: Hmm, what would happen if my `values.yaml` looked something like this?

```yaml
initContainer:
    enabled: True
    scratch:
        root: /foo

worker:
    scratch:
        root: /bar
```

I think this would produce two `scratch`s, which would turn this into invalid yaml. Could you write a test to see what happens? Ideally the chart should fail to render, if that's hard then one `scratch` should win out over the other in a predictable way.

callumforrester (changes requested):
Mostly comments about tests, we're building up helm tests as we go so makes sense there's a few that need writing. Thanks for looking at this @ZohebShaikh!

helm/blueapi/templates/configmap.yaml
callumforrester: Yeah, that could work

callumforrester:
also fixes #646

docs/how-to/edit-live.md
```diff
@@ -56,11 +56,8 @@ host machine, include the following in your `values.yaml`:
 ```yaml
   initContainer:
     enabled: true
-    scratch:
-      root: path/to/scratch/area  # e.g. /dls_sw/<my_beamline>/software/blueapi/scratch
-      repositories: []
 ```
-
+When initContainer is enabled worker must have scratch config.
```
DiamondJoseph, line 60: Please amend this to add the scratch info to the main container config and explain that enabling the initContainer iterates over the scratch config, cloning and setting groups if enabled. If the initContainer is not enabled then the scratch area will still try to be loaded.

helm/blueapi/templates/init-configmap.yaml
```diff
@@ -1,7 +1,11 @@
...
 metadata:
   name: {{ include "blueapi.fullname" . }}-initconfig
 data:
-  initconfig.yaml: |-
-    {{- toYaml .Values.initContainer | nindent 4 }}
+  initconfig.yaml: 
+    scratch:
+      {{- toYaml .Values.worker.scratch | nindent 6 }}
+---
+{{- end }}
```
DiamondJoseph, line 11:
```suggestion
{{- end }}
---
```

nit: I think I'd prefer all of the yaml documents are self contained- if we pipe the output of helm into a file and it has a bunch of adjacent
```
---
---
---
``` 

then it looks a bit ugly but I'd rather have it than accidentally merge two documents

tests/unit_tests/test_helm_chart.py
```diff
@@ -96,81 +96,18 @@ def test_helm_chart_creates_config_map(worker_config: ApplicationConfig):
...
-            }
-        },
-    ],
-)
-def test_helm_chart_creates_init_config_map(values: Values):
-    manifests = render_chart(values=values)
-    rendered_config = yaml.safe_load(
-        manifests["ConfigMap"]["blueapi-initconfig"]["data"]["initconfig.yaml"]
-    )
-    assert rendered_config == values["initContainer"]
```
DiamondJoseph, line 137: This test would have caught the `enabled` issue, so I'd like to see a version of it reinstatated that turns init-config.yaml into an ApplicationConfig to check that no extra fields creep in.

tests/unit_tests/test_helm_chart.py
```diff
@@ -96,81 +96,18 @@ def test_helm_chart_creates_config_map(worker_config: ApplicationConfig):
...
-                    "repositories": [],
-                    "root": "/blueapi-plugins/scratch",
-                },
-            }
-        }
-    )
-    init_containers = manifests["StatefulSet"]["blueapi"]["spec"]["template"]["spec"][
-        "initContainers"
-    ]
-    assert len(init_containers) == 1
```
DiamondJoseph, line 155: This can also come back, with scratch on the main config

tests/unit_tests/test_helm_chart.py
```diff
@@ -256,11 +238,13 @@ def test_do_not_have_to_provide_scratch_host_path_twice():
     config = yaml.safe_load(
         manifests["ConfigMap"]["blueapi-config"]["data"]["config.yaml"]
     )
-    init_config = yaml.safe_load(
-        manifests["ConfigMap"]["blueapi-initconfig"]["data"]["initconfig.yaml"]
-    )
+    init_config = manifests["ConfigMap"]["blueapi-initconfig"]["data"][
```
DiamondJoseph, line 241: Somewhere in the initial safe_load of the grouped manifest, the initContainer config and the main containerConfig are being found to be the same object and being cached. I'm not certain why- we could have even have assert config["scratch"] is init_config["scratch"] below, they're the same object, so trying to yaml_Load it a second time as it was doing was breaking..
I'm going to replicate this comment on the PR into main

tests/unit_tests/test_helm_chart.py
```diff
@@ -305,3 +289,60 @@ def group_manifests(ungrouped: Iterable[Mapping[str, Any]]) -> GroupedManifests:
...
+            "initResources": LOW_RESOURCES,
+            "initContainer": {"enabled": False},
+        },
+        {
+            "initResources": HIGH_RESOURCES,
+            "initContainer": {"enabled": False},
+        },
+    ],
+)
+def test_init_container_config_not_available_when_disabled(config):
```
DiamondJoseph, line 307: Duplicates test_init_container_spec_disablable above, this test can be removed.

tests/unit_tests/test_helm_chart.py
```diff
@@ -305,3 +289,60 @@ def group_manifests(ungrouped: Iterable[Mapping[str, Any]]) -> GroupedManifests:
...
+            },
+        }
+    )
+
+    config = yaml.safe_load(
+        manifests["ConfigMap"]["blueapi-config"]["data"]["config.yaml"]
+    )
+    init_config = manifests["ConfigMap"]["blueapi-initconfig"]["data"][
+        "initconfig.yaml"
+    ]
```
DiamondJoseph, line 346: Load these as ApplicationConfigs

tests/unit_tests/test_helm_chart.py
DiamondJoseph: See comment on `test_init_container_config_copied_from_worker_when_enabled`: I would remove that test and move the logic into this, just to remove the size of the diff

DiamondJoseph (approved):

PR #892, opened 2025-04-02
Coerce ConfigMap to string

DiamondJoseph:


ZohebShaikh (approved):
It is good that you are checking it with `ApplicationConfig` type adapter because it is being loading at run time as `ApplicationConfig`

Issue #723, opened 2024-11-18
 Blueapi should not be a valid target for checking out to the scratch repository


DiamondJoseph:
Blueapi should be a robust, tested and stable platform for developing beamline code. It should be running a known, versioned, release, for reasons including security (reduce attack surface to only what can be run as a plan, not executing all arbitrary code), debugging.

## Acceptance Criteria
- If blueapi is in the scratch `repositories` configuration field, an error message is displayed to the user
- If the scratch init code attempts to install a package named `blueapi`, it errors gracefully
- There are tests for both of these cases

callumforrester:
I think this is a fairly simple addition to the scratch code, when pip installing we can fail if the package is called `blueapi` and display a sensible error message to the user.

PR #893, opened 2025-04-03
add pydantic level validation for blueapi to not be a valid target

stan-dot:
solves #723 

Added a check in scratch.py at the cloning stage (before even pip install). it is kind of redundant - as I also added a check on the pydantic validation of the config that happens earlier than that. The former might be useful if someone chooses to provide config not-through pydantic models - not sure how that would happen but it's more robust that way. Also added a test for parsing the config.

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -13,11 +13,14 @@ from pydantic import (
     Field,
     TypeAdapter,
     ValidationError,
+    field_validator,
 )
 
 from blueapi.utils import BlueapiBaseModel, InvalidConfigError
 
 LogLevel = Literal["NOTSET", "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
+
+FORBIDDEN_OWN_REMOTE_URL = "https://github.com/DiamondLightSource/blueapi.git"
 
 
 class SourceKind(str, Enum):
@@ -90,6 +93,13 @@ class ScratchRepository(BlueapiBaseModel):
         description="URL to clone from",
         default="https://github.com/example/example.git",
     )
+
+    @field_validator("remote_url")
+    @classmethod
+    def check_remote_url(cls, value: str) -> str:
+        if value == FORBIDDEN_OWN_REMOTE_URL:
+            raise ValueError(f"remote_url '{value}' is not allowed.")
+        return value
 
 
 class ScratchConfig(BlueapiBaseModel):
--- a/tests/unit_tests/cli/test_scratch.py
+++ b/tests/unit_tests/cli/test_scratch.py
@@ -188,6 +188,23 @@ def test_setup_scratch_fails_on_wrong_gid(
         setup_scratch(config)
 
 
+def test_setup_scratch_fails_on_blueapi_included(
+    directory_path_with_sgid: Path,
+):
+    b = ScratchRepository.model_construct(
+        name="blueapi",
+        remote_url="https://github.com/DiamondLightSource/blueapi.git",
+    )
+    config = ScratchConfig.model_construct(
+        root=directory_path_with_sgid,
+        required_gid=12345,
+        repositories=[b],
+    )
+    assert get_owner_gid(directory_path_with_sgid) != 12345
+    with pytest.raises(PermissionError):
+        setup_scratch(config)
+
+
 @pytest.mark.skip(
     reason="""
 We can't chown a tempfile in all environments, in particular it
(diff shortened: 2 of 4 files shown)


stan-dot:
I think it's another one of the cases when `raise` is not reported correctly by codecov.

src/blueapi/config.py
```diff
@@ -13,12 +13,15 @@
...
     TypeAdapter,
     ValidationError,
+    field_validator,
 )
 
 from blueapi.utils import BlueapiBaseModel, InvalidConfigError
 
 LogLevel = Literal["NOTSET", "DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
 
+FORBIDDEN_OWN_REMOTE_URL = "https://github.com/DiamondLightSource/blueapi.git"
```
DiamondJoseph, line 23: This won't catch if I fork blueapi, or use the ssh address, or if I use different casing

```
$git clone https://github.com/diamondlightsource/blueapi.git
Cloning into 'blueapi'...
```

DiamondJoseph:
@dan-fernandes is going to be in charge of reviewing this

src/blueapi/config.py
stan-dot: I considered the ssh option and rejected it, as at the moment we use https for scratch cloning.

casing is a good question, however the automatic copy feature from github has different casing.

forking - who does it? 

@dan-fernandes would a filter on package name be better? alternatively we could implement casing finally the third option is to use an environment variable.

dan-fernandes: Github URLs [seem to be case-insensitive](https://github.com/github/docs/issues/32838), so I reckon we should `.lower` both the remote url and the banned one for comparison, then pull the uncapitalised URL.

Would it be reasonable for `FORBIDDEN_OWN_REMOTE_URLS` to be a tuple, and add the ssh address to be safe?

I can't think of a sure-fire way to forbid forks, but if it's not overkill you can query [github's REST API at `/repos/{owner}/{repo}` to get a repository, and check its `parent`.](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository)

stan-dot: the query check would be some wait, making this loading longer (network IO)

arguably the initial issue spec (https://github.com/DiamondLightSource/blueapi/issues/723#issuecomment-2483256681) mentions primarily package name rather than url.
https://github.com/DiamondLightSource/blueapi/issues/723#issue-2668744657


I see forks as quite an exotic scenario - why would someone do that? Does the number of times that is done warrant a slowdown for all the users?

dan-fernandes: I agree that it isn't proportional to the likelihood of someone trying to pull a fork of BlueAPI.

The initial spec does suggest we should filter on package name. However if we're ignoring forks, I'd argue that the URL + basic formatting (`lower`) is more robust than trusting the user to correctly label the package name. That being said, I don't immediately see a downside to filtering off of both.

stan-dot: >  trusting the user to correctly label the package name

by default the package name is encoded in python and it's in all the python files with `import from blueapi.something` internally. Someone would need to do lots of rewiring to effectively change the name. namespace is only extended here by reading out one directory, and we can assume people would use `git clone` to get that.

I'm happy to do both lowercase filter and `blueapi` as name - if we run into issues with that we can make fixes against a specific failure mode later.

dan-fernandes: Ah okay - I think I understand. 

>I'm happy to do both lowercase filter and blueapi as name - if we run into issues with that we can make fixes against a specific failure mode later.

This sounds like a good solution.

stan-dot: applied the change into the latest commit

tests/unit_tests/cli/test_scratch.py
```diff
@@ -188,6 +188,23 @@ def test_setup_scratch_fails_on_wrong_gid(
         setup_scratch(config)
 
 
+def test_setup_scratch_fails_on_blueapi_included(
```
dan-fernandes, line 191: Could: Parametrize to repo name and remote url to exercise different conditions in
```
        if (
            repo.remote_url.lower() == FORBIDDEN_OWN_REMOTE_URL.lower()
            or repo.name == "blueapi"
        ):
```

stan-dot: remote_url test just would test for the lowercaseness - the happy test we have elsewhere anyway

PR #898, opened 2025-04-04
Dummy auth

DiamondJoseph:
Requires API version bump if this is sufficient

### DIFF
--- a/docs/reference/openapi.yaml
+++ b/docs/reference/openapi.yaml
@@ -339,8 +339,13 @@ components:
       title: WorkerTask
       type: object
   securitySchemes:
-    OAuth2:
-      flows: {}
+    OAuth2AuthorizationCodeBearer:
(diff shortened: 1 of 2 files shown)


src/blueapi/service/main.py
```diff
@@ -108,6 +108,8 @@ def get_app(config: ApplicationConfig):
     dependencies = []
     if config.oidc:
         dependencies.append(Depends(verify_access_token(config.oidc)))
+    else:
+        dependencies.append(Depends(OAuth2(auto_error=False)))
```
tpoliaw, line 112:
```suggestion
        dependencies.append(Depends(HTTPBearer(auto_error=False)))
```

I think this would have the same effect in terms of adding the security context to the generated schema and doesn't need the diamond configuration (or dummy values) hard-coded into the schema generation.

PR #906, opened 2025-04-08
Fix FastAPI lint issues

### LINKED ISSUE #905: Add `FAST` Linter Rules from Ruff for FastAPI Projects
We should consider enabling the [`FAST` linter rules](https://docs.astral.sh/ruff/rules/#fastapi-fast) in Ruff, specifically for FastAPI-based projects like **blueapi**. These rules enforce conventions recommended by the FastAPI documentation, potentially improving code quality and consistency.

After a quick run on the existing blueapi codebase, there are around 25 violations — mostly `FAST001` and `FAST002`, which appear to be easy to fix. Given this, incorporating these checks now would be low-effort and help catch issues earlier in development.

Since the rules are FastAPI-specific, they should **only be included in  blueapi**, not globally across all templates.

## Acceptance Criteria
- [ ] Decision made on whether to include `FAST` rules in Ruff config for FastAPI project (blueapi)
- [ ] If approved, update  `blueapi` to include these and fix existing `FAST001` and `FAST002` issues in `blueapi`

ZohebShaikh:
closes #905


DiamondJoseph (approved):
Love that it has no impact on the API version

PR #907, opened 2025-04-08
Use optional HTTP Bearer auth when no OIDC config is present

tpoliaw:
Allows the schema to have enough information without having to include DLS configuration in the repo schema.


tpoliaw:
No longer required with the schema not being used to generate the java client

PR #914, opened 2025-04-10
feat: add StartDocumentPathProvider

### LINKED ISSUE #875: Make blueapi create its own path provider and not import from dodal
We currently need to use the `StartDocumentPathProvider` from dodal to use the numtracker client for file naming. It would be preferable to have blueapi do this instead so that you could setup blueapi with the correct configuration and it would do it for you. Then you would not have to put the  line in the start of the dodal modules `set_path_provider(StartDocumentPathProvider())` and we can leave beamline modules to define their own offline defaults.

## Acceptance Criteria
- Blueapi defines the StartDocumentPathProvider

abbiemery:
Needed for #875 . As well as https://github.com/DiamondLightSource/dodal/pull/1170 and https://github.com/DiamondLightSource/dodal/pull/1171


abbiemery:
This now needs doing first too : https://github.com/DiamondLightSource/dodal/issues/1191

src/blueapi/service/interface.py
```diff
@@ -117,26 +120,28 @@ def _update_scan_num(md: dict[str, Any]) -> int:
...
-    path_provider = get_path_provider()
+    try:
+        path_provider = get_path_provider()
+    except NameError:
+        path_provider = None
+
     run_engine = context().run_engine
 
-    if isinstance(path_provider, StartDocumentPathProvider):
+    if path_provider is None and numtracker_client() is not None:
```
callumforrester, line 142: Could: Use `logging.debug` to note if you're not going to configure a path provider

src/blueapi/path_provider.py
callumforrester, line 1: Should: I think this and the corresponding test file should go somewhere other than the top level, maybe in `utils`?

tests/unit_tests/service/test_interface.py
```diff
@@ -436,6 +440,32 @@ def test_setup(mock_stomp: MagicMock):
     assert interface.worker()._ctx is not None
     assert interface.context().run_engine.scan_id_source == interface._update_scan_num
 
+    clear_path_provider()
+    interface.teardown()
+
+
+@patch("blueapi.service.interface.StompClient")
+def test_setup_without_path_provider_with_numtracker_makes_start_document_provider(
+    mock_stomp: MagicMock,
```
callumforrester, line 449: Should: Is this needed? Given the config provided all stomp settings should be ignored anyway

tests/unit_tests/service/test_interface.py
```diff
@@ -436,6 +440,32 @@ def test_setup(mock_stomp: MagicMock):
...
+        ),
+        numtracker=NumtrackerConfig(),
+    )
+    interface.set_config(conf)
+    interface.setup(conf)
+
+    path_provider = get_path_provider()
+
+    assert isinstance(path_provider, StartDocumentPathProvider)
+    assert interface.worker()._ctx is not None
```
callumforrester, line 465: Could: I would prefer not to have private member access unless you think it adds a lot of value to the test

callumforrester (changes requested):

tests/unit_tests/service/test_interface.py
abbiemery: I've removed it from the other one too.

src/blueapi/service/interface.py
abbiemery: shifted this round a bit.

src/blueapi/path_provider.py
abbiemery: Done

src/blueapi/service/interface.py
```diff
@@ -117,27 +121,34 @@ def _update_scan_num(md: dict[str, Any]) -> int:
...
-    if isinstance(path_provider, StartDocumentPathProvider):
+    if path_provider is None:
+        path_provider = StartDocumentPathProvider()
+        set_path_provider(path_provider)
         run_engine.subscribe(path_provider.update_run, "start")
+    else:
+        logging.debug(
+            "A StartDocumentPathProvider has not been configured for numtracker"
+            f"because a different path provider was already set: {path_provider}"
+        )
```
callumforrester, line 151: I may be contradicting my past self, but is this a valid use case? Would we ever expect someone to have numtracker config and also their own `PathProvider`? If not I think we should just raise

src/blueapi/utils/path_provider.py
```diff
@@ -0,0 +1,42 @@
...
+    def __call__(self, device_name: str | None = None) -> PathInfo:
+        """Returns the directory path and filename for a given data_session.
+
+        The default template for file naming is: "{device_name}-{instrument}-{scan_id}"
+        however, this can be changed by providing a template in the start document. For
+        example: "template": "custom-{device_name}--{scan_id}".
+
+        If you do not provide a data_session_directory it will default to "/tmp".
+        """
+        template = self._doc.get("template", DEFAULT_TEMPLATE)
```
callumforrester, line 39: Since this isn't in use yet, can we change the key to something more unique that's less likely to confict with other start document metadata e.g. 

```suggestion
        template = self._doc.get("data_file_path_template", DEFAULT_TEMPLATE)
```

I'm okay with making a separate issue for this, but it's a small change so can do it here if preferred.

tests/unit_tests/service/test_interface.py
callumforrester, line 1: To check, do these tests cover all 4 config permutations?
- Numtracker config and preexisting path provider
- Numtracker config and no preexisting path provider
- No numtracker config and preexisting path provider
- No numtracker config and no preexisting path provider

The 4th is the base case so I guess effectively covered by other tests?

callumforrester (approved):
Happy to merge, a few questions/suggestions

src/blueapi/utils/path_provider.py
abbiemery: Good idea

src/blueapi/service/interface.py
abbiemery: No I don't think so. I'll make it raise

tests/unit_tests/service/test_interface.py
abbiemery: These were mostly covered but i've restructured/added some tests to make them clearer and more explicit.

callumforrester (approved):
LGTM

PR #918, opened 2025-04-14
ci: Enforce Conventional Commit PR titles

DiamondJoseph:
See also: [the proposed ADR in this PR](https://github.com/DiamondLightSource/python-copier-template/pull/255/files)

And how release-please was [able to automate a release](https://github.com/DiamondJoseph/monster-stats/pull/18) in this example repository.

### DIFF
--- a/helm/blueapi/templates/statefulset.yaml
+++ b/helm/blueapi/templates/statefulset.yaml
@@ -12,6 +12,9 @@ spec:
   template:
     metadata:
       annotations:
+      {{- if ((.Values.worker.logging).graylog).enabled }}
+        fluentd-ignore: "true"
+      {{- end }}
       {{- if .Values.restartOnConfigChange }}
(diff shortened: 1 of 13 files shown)


callumforrester:
If https://github.com/DiamondLightSource/python-copier-template/pull/255 is merged upstream then we can just update the copier template rather than merge this, correct?

callumforrester:
Specifically I don't think we should enforce this

> PRs will be squash-and-merged, with the complete git history of the change preserved in the PR but only a single commit on the default branch.

DiamondJoseph:
No, the change to the python-copier-template only enables it for the copier template.

Why do you not want to enforce squash-and-merge?

callumforrester (changes requested):
Happy to enforce it for blueapi, not for copier template projects, so all good. Should have a separate ADR here though.

PR #926, opened 2025-04-22
docs: Add basic logging documentation

dan-fernandes:


docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,31 @@
+
+# Configure Logging
+
+By default BlueAPI will log to stdout at the [INFO level](https://docs.python.org/3/library/logging.html#logging-levels), but can be reconfigured to log at any level, and to output to Graylog.
+
+As the BlueAPI log handlers are set at root level they will recieve messages from all loggers in the same process that [allow propagation](https://docs.python.org/3/library/logging.html#logging.Logger.propagate).
+
+To add logging to a custom module in BlueAPI, instantiate a logger from the [standard library logging package, then use any of its log methods](https://docs.python.org/3/library/logging.html#logger-objects). Propagation is enabled by default.
```
DominicOram, line 8: Could: A small 2-3 line example here would be great.

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,31 @@
+
+# Configure Logging
+
+By default BlueAPI will log to stdout at the [INFO level](https://docs.python.org/3/library/logging.html#logging-levels), but can be reconfigured to log at any level, and to output to Graylog.
+
+As the BlueAPI log handlers are set at root level they will recieve messages from all loggers in the same process that [allow propagation](https://docs.python.org/3/library/logging.html#logging.Logger.propagate).
```
DominicOram, line 6: Nit: The allowing propagation stuff seems like a lot of detail. I didn't even know you could turn this off, I suspect those new to Python logging will never need to touch it.

DominicOram (changes requested):
Minor comments. Some additional questions:
* So this isn't enabled by default? Why not just make a `bluesky` stream in graylog and log into that by default? I see no reason anyone would not want it on?
* If we do just enable it as default I think the info about `fluentd` is too much detail. Instead you just need to say where it's going by default and how to change it if you need to
* A link to our graylog instance would be good, even better if it's directly to the default stream
* What fields are added by default? Can the user change fields if they want to? This is probably a more "Advanced" feature

dan-fernandes:
@DominicOram 
> So this isn't enabled by default?
If it's enabled by default you would log to the Graylog instance by just checking out / spinning up the repo while on a Diamond computer. Would a better compromise be having Graylog enabled by default, but to `localhost`?

The rest I agree with and will implement before requesting another review

docs/how-to/configure-logging.md
dan-fernandes: Added (very) basic example

docs/how-to/configure-logging.md
dan-fernandes: References to propagation removed

DiamondJoseph:
> Hmm... Does `blueAPI` have no general way of knowing if you're in dev or prod? Is there any plan for this? I appreciate it's outside the scope of the issue so happy for it not to be on as default for now

I don't think there's any plans for that- development can mean so many things and a single setting can't encapsulate them all:
Does dev mean development of the blueapi service or of plans and devices?
Is any deployment in kubernetes "production"?

What does dev mean in terms of behaviour? 
- Doesn't connect/connects mock mode/connects to particular [sim?] devices?
- Doesn't connect to a message bus/connects to a local/specific message bus
- Doesn't send to graylog? Sends to local development graylog? Logs at DEBUG?
- Doesn't use an identity provider/uses a mock/mocks the impl?
- Doesn't talk to numTracker/talks to a dummy instance/mocks the impl?

All of these are things you sometimes need to do while developing blueapi, and half of them are things you might need to do while developing plans.

I think the config should just expose as much as is reasonable, and provide convenient defaults that work without any external infrastructure, both when starting the service locally/in a container and when installing the Helm chart to a fresh cluster.

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,54 @@
...
+logger.info("FOO")
+```
+
+# Running BlueAPI on the Cluster
+
+Services hosted on the cluster automatically have their stdout forwarded to Graylog via a service called fluentd. Due to this, BlueAPI services hosted on the cluster will always log to Graylog.
+
+In this instance, the difference between enabling and disabling graylog in BlueAPI's configuration, is that fluentd only fowards plaintext logs, while BlueAPI's native graylog produces structured data with rich metadata included.
+
+# Were to Find Logs
```
callumforrester, line 44:
```suggestion
# Where to Find Logs
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,54 @@
...
+
+BlueAPI is written to intercept logs from any python code it executes.
+
+```
+import logging
+logger = logging.getLogger(__name__)
+logger.info("FOO")
+```
+
+# Running BlueAPI on the Cluster
```
callumforrester, line 38:
```suggestion
# Kubernetes
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,54 @@
+
+# Configure Logging
+
+By default BlueAPI will log to stdout at the [INFO level](https://docs.python.org/3/library/logging.html#logging-levels), but can be reconfigured to log at any level, and to output to Graylog.
+
+When Graylog is enabled, BlueAPI will continue to log to stdout.
```
callumforrester, line 6:
```suggestion
When logging to [Graylog](https://graylog.org) is enabled, BlueAPI will also continue to log to stdout.
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,54 @@
...
+
+```
+import logging
+logger = logging.getLogger(__name__)
+logger.info("FOO")
+```
+
+# Running BlueAPI on the Cluster
+
+Services hosted on the cluster automatically have their stdout forwarded to Graylog via a service called fluentd. Due to this, BlueAPI services hosted on the cluster will always log to Graylog.
```
callumforrester, line 40:
```suggestion
Services hosted on the DLS clusters automatically have their stdout forwarded to Graylog via a service called fluentd. Due to this, BlueAPI services hosted on the cluster will always log to Graylog.
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,54 @@
...
+
+An example logging config is shown below:
+```
+worker:
+    logging:
+      level: "INFO"
+      graylog:
+        enabled: True
+        host: "graylog-log-target.diamond.ac.uk"
+        port: 12232
```
callumforrester, line 22: The `worker` part is helm-specific

```suggestion
  logging:
    level: "INFO"
    graylog:
      enabled: True
      host: "graylog-log-target.diamond.ac.uk"
      port: 12232
```

callumforrester (changes requested):
Looks good, a few suggestions

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,53 @@
...
+      host: "graylog-log-target.diamond.ac.uk"
+      port: 12232
+```
+
+Here, BlueAPI will accept all messages are INFO level and above and will forward these to both the referenced Graylog address, and to stdout.
+
+# Instrumenting Libraries
+To instrument to a custom module in BlueAPI, instantiate a logger from the [standard library logging package, then use any of its log methods](https://docs.python.org/3/library/logging.html#logger-objects).
+
+BlueAPI is written to intercept logs from any python code it executes.
```
tpoliaw, line 29:
```suggestion
BlueAPI is configured to handle logging from any python code it executes.
```

This is a bit picky but I'm not aware of blueapi doing any logging handling/intercepting itself.

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,53 @@
...
+import logging
+logger = logging.getLogger(__name__)
+logger.info("FOO")
+```
+
+# Kubernetes
+
+Services hosted on the DLS clusters automatically have their stdout forwarded to Graylog via a service called fluentd. Due to this, BlueAPI services hosted on the cluster will always log to Graylog.
+
+In this instance, the difference between enabling and disabling graylog in BlueAPI's configuration, is that fluentd only fowards plaintext logs, while BlueAPI's native graylog produces structured data with rich metadata included.
```
tpoliaw, line 41: Presumably the logs will be duplicated if graylog is configured directly. Is it worth mentioning that and/or is it possible to disable stdout logs while keeping graylog?

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,53 @@
...
+
+# Where to Find Logs
+
+By default logs can be found wherever your stdout is. 
+
+If Graylog is enabled, logs will be forwarded to whichever Graylog instance the configuration addresses. By default this is the main Diamond instance, which can be accessed via [graylog.diamond.ac.uk](https://graylog.diamond.ac.uk/)
```
tpoliaw, line 47: where is the default being set? In the helm charts?

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,53 @@
...
+
+If your BlueAPI server is running on the cluster, stdout is likely being forwarded to the above Graylog instance.
```
tpoliaw, line 49:
```suggestion
If your BlueAPI server is running on the Diamond cluster, stdout is likely being forwarded to the above Graylog instance.
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,53 @@
...
+```
+  logging:
+    level: "INFO"
+    graylog:
+      enabled: True
+      host: "graylog-log-target.diamond.ac.uk"
+      port: 12232
+```
+
+Here, BlueAPI will accept all messages are INFO level and above and will forward these to both the referenced Graylog address, and to stdout.
```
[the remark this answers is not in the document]
dan-fernandes: Agreed - the sentence has been reworded.

docs/how-to/configure-logging.md
dan-fernandes: This is a good point - we have set up the helm chart to automatically disable fluentd when direct graylog logging is enabled. That is:
```
      annotations:
      {{- if ((.Values.worker.logging).graylog).enabled }}
        fluentd-ignore: "true"
      {{- end }}
```

I'll add a sentence explaining this.

docs/how-to/configure-logging.md
dan-fernandes: When [set_up_logging](https://github.com/DiamondLightSource/blueapi/blob/e0a51d5ed32923dda5c6390ee9b774d5a4b92259/src/blueapi/log.py#L24) is called it always sets up a stream handler which will forward all logs to stdout.

If Graylog is enabled, both a Graylog handler and a stream handler are set up.

Currently there is no case where the logs are not sent to stdout. Would this be useful?

tpoliaw: I meant the `By default this is the main Diamond instance...` bit. Where is diamond's graylog the default?

tpoliaw: Fair enough, the docs match reality but it's a bit odd for Diamond specific graylog config to be in the default values when other urls use example values, eg `example.diamond.ac.uk` or `http://opentelemetry-collector.tracing`. Probably not an issue for this PR though

callumforrester: It would alarm me more if it was also enabled by default, so could be worse. I would say it's mainly a "prettiness" thing.

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,55 @@
+
+# Configure Logging
+
+By default BlueAPI will log to stdout at the [INFO level](https://docs.python.org/3/library/logging.html#logging-levels), but can be reconfigured to log at any level, and to output to Graylog.
```
DiamondJoseph, line 4:
```suggestion
By default BlueAPI will log to stdout at the [INFO level](https://docs.python.org/3/library/logging.html#logging-levels), but can be configured to log at any level, and to output simultaneously at the same level to [Graylog](https://graylog.org).
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,55 @@
...
+
+When logging to [Graylog](https://graylog.org) is enabled, BlueAPI will also continue to log to stdout.
+
```
DiamondJoseph, line 7:
```suggestion

```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,55 @@
...
+:::
+
+An example logging config is shown below:
+```yaml
+  logging:
+    level: "INFO"
+    graylog:
+      enabled: True
+      host: "graylog-log-target.diamond.ac.uk"
+      port: 12232
```
DiamondJoseph, line 21:
```suggestion
      port: 12231
```

We should adjust the defaults to hit the right graylog too

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+With this configuration, all logs at INFO level or above will be logged to both stdout and the configured graylog.
+
+# Instrumenting Libraries
+To instrument to a custom module in BlueAPI, instantiate a logger from the [standard library logging package, then use any of its log methods](https://docs.python.org/3/library/logging.html#logger-objects).
+
+BlueAPI is configured to handle logging from any python code it executes.
```
DiamondJoseph, line 29:
```suggestion
Logs from plans run by BlueAPI will use inherit the configured logging, simply instantiate a logger from the [standard library logging package, then use any of its log methods](https://docs.python.org/3/library/logging.html#logger-objects).
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+import logging
+logger = logging.getLogger(__name__)
+logger.info("FOO")
```
DiamondJoseph, line 34:
```suggestion
import logging
logger = logging.getLogger(__name__)

def my_plan() -> MsgGenerator:
  logger.info("Running my_plan")
  
logger.debug("Loaded module with my_plan")
```

docs/how-to/configure-logging.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+# Kubernetes
+
+Services hosted on the DLS clusters automatically have their stdout forwarded to Graylog via a service called fluentd. Due to this, BlueAPI services hosted on the cluster will always log to Graylog.
+
+When BlueAPI's native Graylog support is enabled it forwards structured data rather than plaintext.
+
+When BlueAPI's native Graylog is enabled fluentd will be automatically disabled to avoid log duplication.
```
DiamondJoseph, line 43:
```suggestion
Services hosted on the DLS clusters automatically have their stdout forwarded to Graylog via a service called fluentd. Therefore BlueAPI hosted on the cluster will always have logs available from Graylog, however BlueAPI's native Graylog forwards structured data rather than plaintext.

When native Graylog is enabled fluentd will be automatically disabled to avoid log duplication.
```

PR #928, opened 2025-04-23
feat: Enable/disable stomp with a flag in config

### LINKED ISSUE #919: Enable/disable stomp with `bools` instead of `nulls` in the configuration
In order to disable stomp on the server side I have to pass a null in my config e.g.

```yaml
api:
  port: 8000
stomp: null 
env: ...
```

This doesn't feel very clean and can cause confusion with downstream applications, e.g. we have observed it conflicting with https://github.com/argoproj/argo-cd/issues/16312 on the beamline. A better solution would be to pass bools e.g.

```yaml
stomp:
  enabled: False
```

## Acceptance Criteria
- I can disable any attempt to connect to stomp by passing an explicit `False` in config
- There are tests for all code modules that touch stomp, to make sure they respect this

callumforrester:
Fixes #919 

BREAKING CHANGE: Stomp configuration is now disabled by default in the helm chart. To make blueapi connect to a stomp message broker you have to include`enabled: true` under `worker.stomp` in your `values.yaml` file.

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -44,9 +44,12 @@ class StompConfig(BlueapiBaseModel):
         "event publishing",
         default=False,
     )
-    host: str = "localhost"
-    port: int = 61613
-    auth: BasicAuthentication | None = None
+    host: str = Field(description="STOMP broker host", default="localhost")
+    port: int = Field(description="STOMP broker port", default=61613)
+    auth: BasicAuthentication | None = Field(
+        description="Auth information for communicating with STOMP broker, if required",
+        default=None,
+    )
 
 
 class WorkerEventConfig(BlueapiBaseModel):
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -39,6 +39,11 @@ class StompConfig(BlueapiBaseModel):
     Config for connecting to stomp broker
     """
 
+    enabled: bool = Field(
+        description="True if blueapi should connect to stomp for asynchronous "
+        "event publishing",
+        default=False,
+    )
     host: str = "localhost"
     port: int = 61613
     auth: BasicAuthentication | None = None
@@ -194,7 +199,7 @@ class ApplicationConfig(BlueapiBaseModel):
     config tree.
     """
 
-    stomp: StompConfig | None = None
+    stomp: StompConfig = Field(default_factory=StompConfig)
     env: EnvironmentConfig = Field(default_factory=EnvironmentConfig)
     logging: LoggingConfig = Field(default_factory=LoggingConfig)
     api: RestConfig = Field(default_factory=RestConfig)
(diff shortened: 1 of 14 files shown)


src/blueapi/config.py
```diff
@@ -39,9 +39,17 @@ class StompConfig(BlueapiBaseModel):
...
 
-    host: str = "localhost"
-    port: int = 61613
-    auth: BasicAuthentication | None = None
+    enabled: bool = Field(
+        description="True if blueapi should connect to stomp for asynchronous "
+        "event publishing",
+        default=False,
+    )
+    host: str = Field(description="STOMP broker host", default="localhost")
```
DiamondJoseph, line 47: nit: We could also use

```suggestion
    host: Annotated[str, "STOMP broker host"] = "localhost"
```

but I think it's a matter of personal taste.

DiamondJoseph (changes requested):
```
feat: Enable/disable stomp with a flag in config

BREAKING CHANGE: config key `stomp.enabled` must now be enabled when using stomp: this was previously implicit if `stomp` was a valid configuration but now must be explicitly set
```

In keeping with conventional commits, I'd like to see [a footer explaining the breaking change](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer). If this was >=1.0.0 it'd be `feat!` but as we're <1.0.0 I think `feat` and a footnote is sufficient?

src/blueapi/config.py
callumforrester: Would be good to standardise across the repo though, maybe a different PR?

DiamondJoseph: Fine with that, looks like our use of Annotated is very small, but is used for fastapi Depends and for one of the SkipJsonSchemas

src/blueapi/config.py
```diff
@@ -39,9 +39,17 @@ class StompConfig(BlueapiBaseModel):
     Config for connecting to stomp broker
     """
 
-    host: str = "localhost"
-    port: int = 61613
-    auth: BasicAuthentication | None = None
+    enabled: bool = Field(
```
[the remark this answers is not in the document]
DiamondJoseph: I think that Helm's default values.yaml should always aim to work out of the box, with as little required extra infrastructure as possible. Since Helm is what people will actually be configuring most of the time, it gets to be opinionated about what the defaults should therefore be.

DiamondJoseph (approved):
Approving with the proviso that the merge commit gets a footer as discussed

src/blueapi/config.py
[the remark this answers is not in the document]
noemifrisina: Fair enough, I was just wondering at the reasoning behind the choice but I don't have a strong opinion either way. Could this choice be added to the config docs please?

PR #935, opened 2025-05-02
fix!: Improve error reporting from CLI

tpoliaw:


DiamondJoseph (approved):

DiamondJoseph (changes requested):
This does not behave nicely when invalid JSON is passed

tpoliaw:
> This does not behave nicely when invalid JSON is passed

This should handle both invalid JSON cases now

```
$ blueapi -c conf.yaml controller run stp_snapshot '[]'
Error: Incorrect parameters supplied
    Invalid value [] for field params: Input should be a valid dictionary
```
for the non-dictionary, can't-build-the-task error, and
```
$ blueapi -c conf.yaml controller run stp_snapshot ']'
Error: Parameters are not valid JSON: Expecting value: line 1 column 1 (char 0)
```
for the invalid JSON error.

DiamondJoseph (approved):

PR #940, opened 2025-05-09
Remove workaround

### LINKED ISSUE #927: Debugging a unit test in vscode also triggers all system tests
Debugging 1 or more tests via the vscode test menu (screenshot below) also seems to trigger all of the system tests to run afterwards. I've noticed this in my and @abbiemery's environments. 

Image

## Steps To Reproduce
1. Go to the test menu in vscode, make sure it is a passing test and that blueapi is not running
2. Debug the test (see screenshot above)
3. The test should pass, then the system tests should run, try to connect to blueapi and fail
4. The exception thrown when failing to connect should halt the debugger

## Acceptance Criteria
- Any test in the codebase can be run in debug mode without triggering any other test

ZohebShaikh:
closes #927


ZohebShaikh:
I recommend testing this change in your local setup before merging, just to be safe.
I've verified it both inside the devcontainer and outside, and everything works as expected.

callumforrester:
@ZohebShaikh is it worth just directly pushing this to the Python copier template? https://github.com/DiamondLightSource/python-copier-template/blob/main/tests/conftest.py

PR #941, opened 2025-05-09
fix:  change set up order of the worker to make a StartDocumentPathProvider before device instantiation

abbiemery:
Fixes https://github.com/DiamondLightSource/blueapi/issues/950

Previously the order of device instantiation was missed, so we tired to make devices before a provider was set for you if have configured numtracker. This changes the order so that if numtracker is configured, it will set the path provider as a `StartDocumentBasedPathProvider` first. If there is a path provider set in a module file this will be overwritten and numtracker will not be able to be used.


src/blueapi/service/interface.py
```diff
@@ -48,10 +48,13 @@ def set_config(new_config: ApplicationConfig):
 @cache
 def context() -> BlueskyContext:
     ctx = BlueskyContext()
-    ctx.with_config(config().env)
     return ctx
 
 
+def instantiate_devices_in_context() -> None:
+    context().with_config(config().env)
```
DiamondJoseph, line 55: Is instantiating all of the devices all that with_config does on Context? If so it could use a better name, if not this function should get a better name

src/blueapi/service/interface.py
```diff
@@ -125,29 +128,29 @@ def setup(config: ApplicationConfig) -> None:
...
 
+    instantiate_devices_in_context()
 
-def _hook_run_engine_and_path_provider() -> None:
-    try:
-        path_provider = get_path_provider()
-    except NameError:
-        path_provider = None
+    if numtracker_client() is not None:
+        if not isinstance(get_path_provider(), StartDocumentPathProvider):
```
DiamondJoseph, line 140:
```suggestion
    if numtracker_client() is not None and not isinstance(get_path_provider(), StartDocumentPathProvider):
```

Merge these two

abbiemery: Done.

src/blueapi/service/interface.py
abbiemery: Also no, thats not all it does. It registers the plans and devices then instantiates and connects the devices

tests/unit_tests/service/test_interface.py
```diff
@@ -425,28 +427,23 @@ def test_configure_numtracker_with_no_metadata_fails():
...
-
-    assert interface.config() is not None
-    assert interface.worker() is not None
+    assert get_path_provider() == mock_provider
 
     clear_path_provider()
     interface.teardown()
 
 
 def test_setup_without_numtracker_without_exisiting_provider_does_not_make_one():
```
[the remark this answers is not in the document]
abbiemery: Done.

PR #944, opened 2025-05-12
chore: Update python-copier-template to 3.0.0

DiamondJoseph:
Currently leaves Pyright strict mode disabled, see #945.
Gives ownership of the project in the dev-portal to all of the DAQ Core team.
Drops support for python 3.10, adds support for 3.13: is this a breaking change for the library applications of blueapi?

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -16,5 +16,5 @@ github_org: DiamondLightSource
 package_name: blueapi
 pypi: true
 repo_name: blueapi
-strict_typing: true
+strict_typing: false  # https://github.com/DiamondLightSource/blueapi/issues/945
 type_checker: pyright
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,6 +1,6 @@
 # Changes here will be overwritten by Copier
 _commit: 2.5.0
-_src_path: gh:DiamondLightSource/python-copier-template
+_src_path: https://github.com/DiamondLightSource/python-copier-template
 author_email: callum.forrester@diamond.ac.uk
 author_name: Callum Forrester
 component_lifecycle: production
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,13 +1,12 @@
 # Changes here will be overwritten by Copier
-_commit: 2.5.0
+_commit: 3.0.0
 _src_path: https://github.com/DiamondLightSource/python-copier-template
 author_email: callum.forrester@diamond.ac.uk
 author_name: Callum Forrester
 component_lifecycle: production
-component_owner: user:vid18871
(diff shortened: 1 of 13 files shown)


tpoliaw (commented):
Does blueapi work with 3.13? It may be something in my local environment but trying to run in 3.13 gives several pages of stacktrace that seems to be from `aioca` not supporting it. The [project page](https://github.com/DiamondLightSource/aioca/blob/main/pyproject.toml) doesn't list 3.13 as supported although it might just be out of date.

DiamondJoseph:
> Does blueapi work with 3.13? It may be something in my local environment but trying to run in 3.13 gives several pages of stacktrace that seems to be from `aioca` not supporting it. The [project page](https://github.com/DiamondLightSource/aioca/blob/main/pyproject.toml) doesn't list 3.13 as supported although it might just be out of date.

Looks like I need to update the copier template there first- I'm assuming that it *should* but we'll find out

tpoliaw:
Digging into this a bit more, I don't think `epicscorelibs` (used by aioca) supports 3.13 at all. With an empty environment and a clean install of Python 3.13

```
❯ uv run --with epicscorelibs --python /dls/local_python/Python-3.13.3/python python
  × Failed to build `epicscorelibs==7.0.7.99.1.1`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)
// Pages of C compile errors
```

The same thing with 3.12 runs as expected

tpoliaw:
That's a different error from using the prebuilt python-build-standalone versions of python at runtime. This one is from the library itself and is reproducible without uv.

```
/tmp/epics_demo
❯ python --version                                                                                                                                                                                                                    12:34:21
Python 3.13.3

/tmp/epics_demo
❯ python -m venv env                                                                                                                                                                                                                  12:34:29

took 2s
/tmp/epics_demo
❯ source env/bin/activate                                                                                                                                                                                                             12:34:40

v3.13.3 (env) /tmp/epics_demo
❯ python -m pip install epicscorelibs
// ...
Building wheels for collected packages: epicscorelibs
  Building wheel for epicscorelibs (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Building wheel for epicscorelibs (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [1594 lines of output]
```

DiamondJoseph:
Building aioca with Python 3.13 in their CI uses `epicscorelibs==7.0.7.99.1.1`- but it fails several of the tests and times out on other OSs, so there likely is something more serious underlying that.

tpoliaw:
Think this was done in #1013

PR #946, opened 2025-05-12
fix: Pin click to minimum version with default seperated stdout/stderr

DiamondJoseph:
click [version 8.2.0](https://github.com/pallets/click/releases/tag/8.2.0) tracks stdout and stderr separately by default, where previously they were unified unless explicitly requested.

We pin to >=8.2.0 (which also drops support for versions of python we do not support either anymore, and so is likely to be using more modern code and practices) and replace the handling in tests that expected to use stderr. There are other tests that check stdout, and we should ensure that we pipe to stdout and stderr correctly.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -18,7 +18,7 @@ certifi==2025.1.31
 cffi==1.17.1
 cfgv==3.4.0
 charset-normalizer==3.4.1
-click==8.1.8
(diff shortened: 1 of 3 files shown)


tpoliaw:
The click issue/PR was open for two years and they make the release the day after we start relying on the old behaviour? 🤦

PR #952, opened 2025-05-13
docs: Add example of device addition to docs

### LINKED ISSUE #949: Documentation: Document how to point to specific devices
https://diamondlightsource.github.io/blueapi/main/how-to/add-plans-and-devices.html promises to tell you how to add plans _and devices_
 to your environment but it only tells you how to add plans. It should also tell you how to add devices

## Acceptance Criteria
- Documentation tells you how to add devices to an environment

JamesDoingStuff:
Fixes #949

### DIFF
--- a/docs/how-to/add-plans-and-devices.md
+++ b/docs/how-to/add-plans-and-devices.md
@@ -27,15 +27,12 @@ env:
       module: my_plan_library.tomography.plans
 ```
 
-Devices are added in the same way, just using different values of `kind` depending on their source. [Dodal](https://github.com/DiamondLightSource/dodal) devices use `dodal`, while devices from elsewhere use `deviceFunctions`, like so: 
-
+Devices are added similarly, using `dodal` as the `kind`, like so: 
 ```yaml
 env:
   sources:
     - kind: planFunctions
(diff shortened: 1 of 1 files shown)


docs/how-to/add-plans-and-devices.md
```diff
@@ -19,14 +19,27 @@ For example:
...
-You would add the following into your configuration file:
+To add plans, you would add the following into your configuration file:
 ```yaml
 env:
   sources:
     - kind: planFunctions
       module: my_plan_library.tomography.plans
 ```
 
+Devices are added in the same way, just using different values of `kind` depending on their source. [Dodal](https://github.com/DiamondLightSource/dodal) devices use `dodal`, while devices from elsewhere use `deviceFunctions`, like so:
```
callumforrester, line 30: Could: I don't believe `deviceFunctions` is in use any more, we could just delete it and simplify these docs.

callumforrester (approved):
LGTM, thanks

docs/how-to/add-plans-and-devices.md
JamesDoingStuff: I did wonder - will do :+1:

PR #954, opened 2025-05-19
feat: Add support for swagger UI login

ZohebShaikh:


ZohebShaikh:
I noticed an annoying behavior in Swagger UI related to login. After logging in, the token expires after one minute — which is expected. However, when making a request after the token has expired, you have to manually log out and re-authorize to refresh the token. So, while token refreshing technically works, the need to log out and authorize again each time is quite inconvenient.

ZohebShaikh:
I found a issue related to this https://github.com/swagger-api/swagger-ui/issues/7257  Not much progress has been made here ...

The requestIntercepters type thing was quite intriguing until I found out there is no nice /(possible ) solution to get the refresh token from swaggerUI to fastAPI ...and the other way of injecting javascript to swaggerUI from fastapi is also not very nice .....

I was looking at the oauth proxy as well which might be a better solution than this ....

We can add `leeway=timedelta(hrs=1)` but I don't think it is a good long term solution

ZohebShaikh:
Closing in favour of using oauth2-proxy which has the ability to refresh token and manage sessions

PR #959, opened 2025-05-20
Ouath2 proxy

ZohebShaikh:


helm/blueapi/templates/ingress.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         pathType: Prefix
         backend:
           service:
-            name: {{ include "blueapi.fullname" . }}
+            name: blueapi-oauth2-proxy
```
DiamondJoseph, line 19:
```suggestion
            name: {{ include "blueapi.fullname" . }}
```

helm/blueapi/values.yaml
```diff
@@ -124,5 +124,17 @@ worker:
...
   enabled: false
+
+oauth2-proxy:
+  config:
+    existingSecret: "oauth2-proxy-secret"
+  configFile: |
+    email_domains = [ "*" ]
+    upstreams = [ "{{ include "blueapi.fullname" . }}" ]
+    skip_auth_routes = ["GET=^/config/oidc"]
+    provider = keycloak-oidc
```
DiamondJoseph, line 140:
```suggestion
  config:
    existingSecret: "oauth2-proxy-secret"
  configFile: |
    email_domains = [ "*" ]
    upstreams = [ "{{ include "blueapi.fullname" . }}" ]
    skip_auth_routes = ["GET=^/config/oidc"]
    skip_provider_button = true
    oidc_issuer_url = https://authn.diamond.ac.uk/realms/master
  ingress:
    enabled: true
    hosts:
      - ckk43848.diamond.ac.uk
    path: /
    pathType: Prefix
```

Possibly? I don't know the exact mapping from the alphaConfig that I have used to the old config file, and mix-and-matching with the Secret?

helm/blueapi/values.yaml
```diff
@@ -42,8 +42,8 @@ service:
   port: 80
 
 ingress:
-  create: false
-#  host: foo.diamond.ac.uk  (assumes port = service.port)
+  create: true
+  host: ckk43848.diamond.ac.uk
```
DiamondJoseph, line 46:
```suggestion
  create: false
# host: foo.diamond.ac.uk (assumes port = service.port)
```

You only want the ingress created in one place- we allow the proxy to create it, blueapi only makes a service that is an upstream inside the cluster only

src/blueapi/service/main.py
```diff
@@ -112,6 +112,9 @@ def get_app(config: ApplicationConfig):
     dependencies = []
     if config.oidc:
         dependencies.append(Depends(verify_access_token(config.oidc)))
+        app.swagger_ui_init_oauth = {
+            "clientId": "NOT_SUPPORTED",
+        }
```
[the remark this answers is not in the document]
ZohebShaikh: This is generated from the openapi Yaml spec ...I will need to remove the security from them to remove the pad locks ...
I can do that but I don't think it is a good idea to change the openapi yaml

PR #960, opened 2025-05-21
feat: Add debugging container target to Helm

keithralphs:


Dockerfile
```diff
@@ -25,6 +25,24 @@ COPY . /context
 WORKDIR /context
 RUN touch dev-requirements.txt && pip install --upgrade pip && pip install -c dev-requirements.txt .
 
+FROM build AS debug
+
+# Set origin to use ssh
+RUN git remote set-url origin git@github.com:{{github_org}}/DiamondLightSource/blueapi.git
```
DiamondJoseph, line 31:
```suggestion
RUN git remote set-url origin git@github.com:diamondlightsource/DiamondLightSource/blueapi.git
```

helm/blueapi/templates/statefulset.yaml
```diff
@@ -52,6 +52,14 @@ spec:
         emptyDir:
           sizeLimit: 5Gi
       {{- end }}
+      {{- if .Values.debug.enabled }}
+      - name: home  # Required for vscode to install plugins
+        hostPath:
+          path: {{ .Values.debug.home }}
+      - name: nslcd  # Shared volume between main and sidecar container
+        emptyDir:
+          sizeLimit: 500Mi
```
DiamondJoseph, line 61:
```suggestion
          sizeLimit: 5Mi
```

nit: could be much smaller

Dockerfile
keithralphs: you know it's already shorter than line 26 right?

helm/blueapi/templates/statefulset.yaml
```diff
@@ -96,15 +104,31 @@ spec:
             - name: venv
               mountPath: /venv
             {{- end }}
+            {{- if .Values.debug.enabled }}
+            - mountPath: /home
+              name: home
+            - mountPath: /var/run/nslcd
+              name: nslcd
+            {{- end }}
+          {{- if not .Values.debug.enabled }}
```
dan-fernandes, line 113: Would this be more readable as an `else`?

dan-fernandes (approved):
LGTM

PR #965, opened 2025-05-21
fix: Split tasks into pending and complete

### LINKED ISSUE #915: Trying to run a plan when one is already running gives internal server error
Trying to start a plan while one is already running gives a 500 server error instead of a 4xx user error (confict/failed precondition?).

## Steps To Reproduce
Steps to reproduce the behavior (with the default sample plans and devices):
1. `curl localhost:8000/tasks -d '{"name":"stp_snapshot","params":{"detectors":["det"]}}' -H 'Content-Type: application/json'`
2. `curl localhost:8000/tasks -d '{"name":"stp_snapshot","params":{"detectors":["det"]}}' -H 'Content-Type: application/json'`
3. `curl -XPUT localhost:8000/worker/task -d '{"task_id":"<task_id from #1"}' -H 'Content-Type: application/json'`
4. `curl -XPUT localhost:8000/worker/task -d '{"task_id":"<task_id from #2"}' -H 'Content-Type: application/json'`

## Acceptance Criteria
- Error should be a 4xx user error instead of a server error

tpoliaw:
Prevents tasks being started twice.

Fixes #915

### DIFF
--- a/src/blueapi/worker/task_worker.py
+++ b/src/blueapi/worker/task_worker.py
@@ -83,7 +83,8 @@ class TaskWorker:
     _ctx: BlueskyContext
     _start_stop_timeout: float
 
-    _tasks: dict[str, TrackableTask]
+    _pending_tasks: dict[str, TrackableTask]
+    _completed_tasks: dict[str, TrackableTask]
 
     _state: WorkerState
     _errors: list[str]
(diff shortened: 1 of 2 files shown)


DiamondJoseph:
Code looks sensible, but does this fix #915? Your example of how to trigger #915 looked like it only ran each task once?

DiamondJoseph (approved):

PR #967, opened 2025-05-22
test: Add test for run_plan with stomp events

tpoliaw:


tpoliaw:
> ...is this not covered by the system tests that are also run in CI?

Probably but the coverage linting wouldn't let me move these lines in #966 unless they were tested and it didn't make sense to add these tests in that PR.

PR #969, opened 2025-05-27
ci: Group opentelemetry packages together

DiamondJoseph:


DiamondJoseph:
This isn't necessary- config:recommends automatically groups these monorepo dependencies together

PR #976, opened 2025-05-27
ci: Adopt renovate best practices

DiamondJoseph:
Adopts the Renovate "best practices" config, which is slightly more exacting than the "recommended" config.

[This config](https://docs.renovatebot.com/presets-config/#configbest-practices) additionally pins to digests for docker tags and github actions, pins dev dependencies (NB, we can set up some automerge for those we are happy to) and auto-updates its own config file

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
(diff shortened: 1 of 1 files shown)


tpoliaw (approved):
Odd that they don't recommend the best practices.

Any idea what's going wrong with pinning Python?

PR #981, opened 2025-05-27
feat: allow mounting development area in a persistent volume

### LINKED ISSUE #986: Allow mounting development area in a persistent volume
To avoid issues with filesystem permissions and make the deployment by Argo more declarative for the state of the scratch area, we should move the scratch area into persistence in the cluster.

## Acceptance Criteria
- the initContainer can be configured to use a persistent volume claim for the scratch area instead of a file mounting
- the initContainer can be configured to re-use an existing claim
- when the initContainer is configured to use and not to re-use, the Helm chart creates a new claim
- There are tests to show that: The claim is created when needed and not when not; the mounting to the volume of both the initContainer and the main container is correctly configured for a new claim and for an existing one; the name of the claim is well chosen for discoverability in the kubernetes dashboard (for now)

DiamondJoseph:
Closes #986

### DIFF
--- a/helm/blueapi/values.yaml
+++ b/helm/blueapi/values.yaml
@@ -95,8 +95,7 @@ tracing:
 # Config for the worker goes here, will be mounted into a config file
 worker:
   api:
-    host: 0.0.0.0 # Allow non-loopback traffic
-    port: 8000
+    url: http://0.0.0.0:8000/ # Allow non-loopback traffic
   env:
     sources:
       - kind: deviceFunctions
@@ -112,8 +111,7 @@ worker:
     auth:
       username: guest
       password: guest
-    host: rabbitmq
-    port: 61613
+    url: http://rabbitmq:61613/
   scratch:
     root: /blueapi-plugins/scratch
     repositories: []
@@ -123,8 +121,7 @@ worker:
     level: "INFO"
     graylog:
       enabled: False
-      host: "graylog-log-target.diamond.ac.uk"
-      port: 12232
+      url: http://graylog-log-target.diamond.ac.uk:12232/
 initContainer:
   enabled: false
   persistentVolume:
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -186,11 +186,13 @@ def listen_to_events(obj: dict) -> None:
     config: ApplicationConfig = obj["config"]
     if not config.stomp.enabled:
         raise BlueskyStreamingError("Message bus needs to be configured")
+    assert config.stomp.url.host is not None, "Stomp URL missing host"
+    assert config.stomp.url.port is not None, "Stomp URL missing port"
     event_bus_client = EventBusClient(
         StompClient.for_broker(
             broker=Broker(
-                host=config.stomp.host,
-                port=config.stomp.port,
+                host=config.stomp.url.host,
+                port=config.stomp.url.port,
                 auth=config.stomp.auth,
             )
         )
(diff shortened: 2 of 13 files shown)


helm/blueapi/templates/statefulset.yaml
```diff
@@ -79,17 +85,26 @@ spec:
...
+          {{- end }}
           - name: venv
             mountPath: /artefacts
       {{- end }}
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}{{ ternary "-debug" "" .Values.debug.enabled }}"
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
DiamondJoseph, line 107:
```suggestion
          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}{{ ternary "-debug" "" .Values.debug.enabled }}"
```

helm/blueapi/templates/statefulset.yaml
```diff
@@ -79,17 +85,26 @@ spec:
           - name: init-config
             mountPath: "/config"
             readOnly: true
+          {{- if .Values.initContainer.persistentVolume }}
+          - name: scratch
+            mountPath: {{ .Values.worker.scratch.root }}
+          - name: home
+            mountPath: /home
+          - name: nslcd
+            mountPath: /var/run/nslcd
```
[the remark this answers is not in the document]
dan-fernandes: I think you're right.

I've removed `home` and `nslcd` but left line 90 (assuming that's a mistake).

helm/blueapi/templates/statefulset.yaml
```diff
@@ -44,15 +44,21 @@ spec:
...
         hostPath:
           path: {{ required "initContainer enabled but no worker.scratch.root!" (.Values.worker.scratch).root }}
           type: Directory
-      - name: venv
-        emptyDir:
-          sizeLimit: 5Gi
       {{- end }}
-      {{- if .Values.debug.enabled }}
+      {{- end }}
+      {{- if or .Values.debug.enabled (and .Values.initContainer.enabled .Values.initContainer.persistentVolume.enabled)}}
```
ZohebShaikh, line 61: Is the or at the correct place ?

```suggestion
      {{- if  .Values.debug.enabled or ( .Values.initContainer.enabled and .Values.initContainer.persistentVolume.enabled)}}
```

If both are identical in syntax I think the changes that I have suggested might be better just for readability

DiamondJoseph: You could do `a | or b` but otherwise it has to be `or a b`: with the function first

helm/blueapi/values.yaml
```diff
@@ -127,6 +127,9 @@ worker:
       port: 12232
 initContainer:
   enabled: false
+  persistentVolume:
+    enabled: true
```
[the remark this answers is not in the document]
DiamondJoseph: In order for it to not be a breaking change, I think yes this should become false by default

helm/blueapi/templates/volumes.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if and .Values.initContainer.persistentVolume.enabled (not .Values.initContainer.persistentVolume.existingClaimName) }}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: scratch-{{ .Values.image.tag | default .Chart.AppVersion }}
+  annotations:
+    helm.sh/resource-policy: keep
+spec:
```
ZohebShaikh, line 8: I think this bit will change to a hook after this [issue](https://github.com/DiamondLightSource/blueapi/issues/987)

ZohebShaikh (approved):

PR #1013, opened 2025-05-29
chore: Update copier template

tpoliaw:
Accepted all changes other than Python 3.13 as it is not yet supported
by dependencies (scanspec, aioca etc)

See #1014

### DIFF
--- a/.copier-answers.yml
+++ b/.copier-answers.yml
@@ -1,13 +1,12 @@
 # Changes here will be overwritten by Copier
-_commit: 2.5.0
+_commit: 3.0.0
 _src_path: https://github.com/DiamondLightSource/python-copier-template
 author_email: callum.forrester@diamond.ac.uk
 author_name: Callum Forrester
 component_lifecycle: production
 component_owner: user:vid18871
 component_type: service
-description:
-    Lightweight bluesky-as-a-service wrapper application. Also usable as
+description: Lightweight bluesky-as-a-service wrapper application. Also usable as
     a library.
 distribution_name: blueapi
 docker: true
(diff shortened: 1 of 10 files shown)


.copier-answers.yml
```diff
@@ -1,13 +1,12 @@
 # Changes here will be overwritten by Copier
-_commit: 2.5.0
+_commit: 3.0.0
 _src_path: https://github.com/DiamondLightSource/python-copier-template
 author_email: callum.forrester@diamond.ac.uk
 author_name: Callum Forrester
 component_lifecycle: production
 component_owner: user:vid18871
```
ZohebShaikh, line 7:
```suggestion
component_owner: group:default/core-data-acquisition
```

ZohebShaikh: Can a ticket be created for this ?

pyproject.toml
```diff
@@ -7,7 +7,6 @@ name = "blueapi"
 classifiers = [
     "Development Status :: 3 - Alpha",
     "License :: OSI Approved :: Apache Software License",
-    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
 ]
```
ZohebShaikh, line 12: Can I a issue be linked here for why python3:13 is not currently supported 

I think there are issues in the dependency for it but not in here

.copier-answers.yml
tpoliaw: Probably but it's small enough that it could be a direct-to-PR change

DiamondJoseph (approved):
I'll make an issue for support 3.13

PR #1018, opened 2025-05-29
chore: Reduce noisiness of Renovate PRs

DiamondJoseph:
Reduces the noisiness of Renovate PRs by grouping indirect dependencies (those specified in dev-requirements.txt, not those in pyproject.toml). Also fixes the pinning to a particular Python container version

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,5 +3,14 @@
   "extends": [
     "config:best-practices",
     ":automergeStableNonMajor"
+  ],
+  "packageRules": [
+    {
+      "matchUpdateTypes": "patch",
+      "matchManagers": [
+        "pip_requirements"
+      ],
+      "description": "Group indirect python patch versions together"
+    }
   ]
 }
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -3,5 +3,36 @@
...
     ":automergeStableNonMajor"
+  ],
+  "packageRules": [
+    {
+      "matchUpdateTypes": [
+        "patch",
+        "minor"
+      ],
+      "matchManagers": [
+        "pip_requirements"
```
DiamondJoseph, line 14: == dev-requirements.txt

tpoliaw: Is it worth switching to json5 so we can include these comments in the source?

tpoliaw (approved):

renovate.json
DiamondJoseph: With the best-practices config we're now using, Renovate may convert it back to .json at any time, removing the comments. I tried to make the description sufficient that it's clear what's going on, I can be even more explicit?

PR #1031, opened 2025-06-02
fix: Use 'data_session_directory' as key in start document

tpoliaw:
To match the key expected in path provider


callumforrester:
Is https://github.com/DiamondLightSource/blueapi/issues/1032 too complicated to do as part of this?

tpoliaw:
It was while trying to test other things. The main issue is that the start document is generated by the run engine using the metadata being modified as a side-effect of the scan_id_source function. To test it we'd either need to set up the full environment and run a plan or mock out the runengine part which defeats the point of the test.

callumforrester:
System test?

DiamondJoseph (approved):

PR #1051, opened 2025-06-10
docs: Include stylisation in README

callumforrester:
Since people keep getting it wrong


callumforrester:
@keithralphs has persuaded me otherwise, people can spell it `BlueAPI`

PR #1062, opened 2025-06-13
Require instrument session to be passed when requesting a plan

### LINKED ISSUE #793: Pass visit ID to blueapi where appropriate

callumforrester:
Fixes #793

### DIFF
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -33,8 +33,8 @@ from blueapi.config import (
 from blueapi.core import OTLP_EXPORT_ENABLED, DataEvent
 from blueapi.log import set_up_logging
 from blueapi.service.authentication import SessionCacheManager, SessionManager
-from blueapi.service.model import SourceInfo
-from blueapi.worker import ProgressEvent, Task, WorkerEvent
+from blueapi.service.model import SourceInfo, TaskRequest
+from blueapi.worker import ProgressEvent, WorkerEvent
 
 from .scratch import setup_scratch
 from .updates import CliEventRenderer
@@ -250,7 +250,11 @@ def run_plan(
             callback(event.name, event.doc)
 
     try:
-        task = Task(name=name, params=parsed_params)
+        task = TaskRequest(
+            name=name,
+            params=parsed_params,
+            instrument_session="cm12345-1",
+        )
     except ValidationError as ve:
         ip = InvalidParameters.from_validation_error(ve)
         raise ClickException(ip.message()) from ip
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -33,7 +33,7 @@ from super_state_machine.errors import TransitionError
 
 from blueapi.config import ApplicationConfig, OIDCConfig
 from blueapi.service import interface
-from blueapi.worker import Task, TrackableTask, WorkerState
+from blueapi.worker import TrackableTask, WorkerState
 from blueapi.worker.event import TaskStatusEnum
 
 from .model import (
@@ -47,6 +47,7 @@ from .model import (
     PythonEnvironmentResponse,
     SourceInfo,
     StateChangeRequest,
+    TaskRequest,
     TaskResponse,
     TasksListResponse,
     WorkerTask,
@@ -54,7 +55,7 @@ from .model import (
 from .runner import WorkerDispatcher
 
 #: API version to publish in OpenAPI schema
-REST_API_VERSION = "0.0.10"
+REST_API_VERSION = "0.1.0"
 
 RUNNER: WorkerDispatcher | None = None
 
@@ -248,23 +249,27 @@ def get_device_by_name(
     return runner.run(interface.get_device, name)
 
 
-example_task = Task(name="count", params={"detectors": ["x"]})
+example_task_request = TaskRequest(
+    name="count",
+    params={"detectors": ["x"]},
+    instrument_session="cm12345-1",
+)
 
 
 @secure_router.post(
     "/tasks",
     status_code=status.HTTP_201_CREATED,
 )
-@start_as_current_span(TRACER, "request", "task.name", "task.params")
+@start_as_current_span(TRACER, "request", "task_request.name", "task_request.params")
 def submit_task(
     request: Request,
     response: Response,
-    task: Annotated[Task, Body(..., example=example_task)],
+    task_request: Annotated[TaskRequest, Body(..., example=example_task_request)],
     runner: Annotated[WorkerDispatcher, Depends(_runner)],
 ) -> TaskResponse:
     """Submit a task to the worker."""
     try:
-        task_id: str = runner.run(interface.submit_task, task)
+        task_id: str = runner.run(interface.submit_task, task_request)
         response.headers["Location"] = f"{request.url}/{task_id}"
         return TaskResponse(task_id=task_id)
     except ValidationError as e:
--- a/src/blueapi/worker/task.py
+++ b/src/blueapi/worker/task.py
@@ -19,6 +19,10 @@ class Task(BlueapiBaseModel):
     params: Mapping[str, Any] = Field(
         description="Values for parameters to plan, if any", default_factory=dict
     )
+    metadata: Mapping[str, Any] = Field(
+        description="Any metadata to apply to all runs within this task",
+        default_factory=dict,
+    )
 
     def prepare_params(self, ctx: BlueskyContext) -> Mapping[str, Any]:
         model = _lookup_params(ctx, self)
@@ -26,11 +30,14 @@ class Task(BlueapiBaseModel):
         return {field: getattr(model, field) for field in model.__pydantic_fields__}
 
     def do_task(self, ctx: BlueskyContext) -> None:
-        LOGGER.info(f"Asked to run plan {self.name} with {self.params}")
+        LOGGER.info(
+            f"Asked to run plan {self.name} with {self.params} and "
+            f"metadata {self.metadata} for all runs"
+        )
 
         func = ctx.plan_functions[self.name]
         prepared_params = self.prepare_params(ctx)
-        ctx.run_engine(func(**prepared_params))
+        ctx.run_engine(func(**prepared_params), metadata_kw=self.metadata)
 
 
 def _lookup_params(ctx: BlueskyContext, task: Task) -> BaseModel:
--- a/tests/unit_tests/service/test_interface.py
+++ b/tests/unit_tests/service/test_interface.py
@@ -13,6 +13,7 @@ from dodal.common.beamlines.beamline_utils import (
 )
 from ophyd.sim import SynAxis
 from stomp.connect import StompConnection11 as Connection
+from tests.unit_tests.service.test_rest_api import FAKE_INSTRUMENT_SESSION
 
 from blueapi.client.numtracker import NumtrackerClient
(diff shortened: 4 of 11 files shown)


callumforrester:
https://github.com/DiamondLightSource/blueapi/pull/1084 is a refactor that makes the auth code and this use the same cache, but can be reviewed separately if preferred

src/blueapi/cli/cli.py
```diff
@@ -231,6 +243,17 @@ def on_event(
...
 )
+@click.option(
+    "-i",
+    "--instrument-session",
+    type=str,
+    help=textwrap.dedent("""Instrument session associated with running the plan,
+    used to tell blueapi where to store any data and as a secuirty check:
+    the session must be valid and active and you must be a member of it.
+    If you have saved your current session with
+    blueapi controller set-instrument-session, passing this will superseed that."""),
```
tpoliaw, line 254:
```suggestion
    help=textwrap.dedent("""
        Instrument session associated with running the plan,
        used to tell blueapi where to store any data and as a security check:
        the session must be valid and active and you must be a member of it.
        If you have saved your current session with
        blueapi controller set-instrument-session, passing this will supersede that."""),
```

dedent does nothing if there's no indentation on the first line (+typos)

src/blueapi/service/main.py
```diff
@@ -254,23 +255,27 @@ def get_device_by_name(
...
+    instrument_session="cm12345-1",
+)
 
 
 @secure_router.post(
     "/tasks",
     status_code=status.HTTP_201_CREATED,
 )
-@start_as_current_span(TRACER, "request", "task.name", "task.params")
+@start_as_current_span(TRACER, "request", "task_request.name", "task_request.params")
```
tpoliaw, line 269: Would it be useful to be logging the visit as well?

tests/conftest.py
```diff
@@ -497,3 +498,11 @@ def mock_numtracker_server() -> Iterable[responses.RequestsMock]:
             json=response_with_errors,
         )
         yield requests_mock
+
+
+@pytest.fixture
+def temp_dir() -> Generator[Path]:
```
tpoliaw, line 504: There is already a `directory_path` fixture in `test_scratch` that is exactly the same as this. Could they both use the same one?

tests/unit_tests/service/test_interface.py
```diff
@@ -13,6 +13,7 @@
 )
 from ophyd.sim import SynAxis
 from stomp.connect import StompConnection11 as Connection
+from tests.unit_tests.service.test_rest_api import FAKE_INSTRUMENT_SESSION
```
tpoliaw, line 16: Test modules depending on other test modules is a bit weird. Should this be a fixture or a per module constant? The `test_client` module is using literals everywhere.

tests/unit_tests/test_cli.py
```diff
@@ -213,26 +213,99 @@ def test_invalid_config_via_env(runner: CliRunner):
...
     runner.invoke(
-        main, ["-c", config_path, "controller", "run", "sleep", '{"time": 5}']
+        main,
+        ["controller", "set-instrument-session", "cm12345-1"],
     )
+    config_path = "tests/unit_tests/example_yaml/rest_and_stomp_config.yaml"
 
-    assert response.call_count == 1
+    mock_cache.return_value = Mock()
+    mock_cache.return_value.get.return_value = "cm12345-1"
```
tpoliaw, line 268:
```suggestion
    mock_cache().get.side_effect = {"instrument_session": "cm12345-1"}.get
```

tests/unit_tests/test_cli.py
```diff
@@ -213,26 +213,99 @@ def test_invalid_config_via_env(runner: CliRunner):
...
+
+    response = responses.post(
+        url="http://a.fake.host:12345/tasks",
+        match=[matchers.json_params_matcher(body_data)],
+    )
+
     runner.invoke(
-        main, ["-c", config_path, "controller", "run", "sleep", '{"time": 5}']
+        main,
+        ["controller", "set-instrument-session", "cm12345-1"],
```
tpoliaw, line 263: Why is this call needed if the cache is then mocked and set directly below?

tests/unit_tests/test_cli.py
```diff
@@ -213,26 +213,99 @@ def test_invalid_config_via_env(runner: CliRunner):
...
+
+
+@patch("blueapi.cli.cli.DiskCache")
+def test_caches_instrument_session(_: Mock, runner: CliRunner):
+    output = runner.invoke(
+        main,
+        ["controller", "set-instrument-session", "cm12345-1"],
+    )
+
+    assert output.stdout == "Default instrument session set to cm12345-1\n"
```
tpoliaw, line 291: Probably also worth asserting that it was set

```
mock_cache.set.assert_called_once_with("instrument_session", "cm12345-1")
```

(or whatever the called_once_with method should be called)

tests/unit_tests/test_cli.py
```diff
@@ -346,14 +447,39 @@ def test_invalid_stomp_config_for_listener(runner: CliRunner):
...
+        main,
+        [
+            "controller",
+            "run",
+            "--background",
+            "sleep",
+            '{"time": 5}',
+        ],
+    )
+    assert result.exit_code == 1
```
tpoliaw, line 479: This test fails if the user has a visit cached. Probably want to mock some things out. Possibly with an autouse fixture to block the user's real cache being used in any of the tests.

src/blueapi/utils/caching.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, path: Path | None = None) -> None:
+        self._path = path or default_cache_dir()
+
+    def set(self, key: str, value: Any) -> None:
+        self._ensure_root()
+        path = self._path_to_value(key)
+        path.unlink(missing_ok=True)
+        with path.open("xb") as writer:
+            as_json = TypeAdapter(T).dump_json(value)
+            writer.write(base64.b64encode(as_json))
```
tpoliaw, line 28: what's the benefit in base64 encoding everything?

src/blueapi/utils/caching.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class DiskCache:
+    def __init__(self, path: Path | None = None) -> None:
+        self._path = path or default_cache_dir()
+
+    def set(self, key: str, value: Any) -> None:
+        self._ensure_root()
+        path = self._path_to_value(key)
+        path.unlink(missing_ok=True)
```
tpoliaw, line 25: Why unlink the file first instead of opening with `wb`?

src/blueapi/utils/caching.py
```diff
@@ -0,0 +1,50 @@
...
+                    base64.b64decode(reader.read()).decode("utf-8")
+                )
+        else:
+            return default
+
+    def _ensure_root(self) -> None:
+        os.makedirs(self._path, exist_ok=True)
+
+    def _path_to_value(self, key: str) -> Path:
+        return self._path / key
```
tpoliaw, line 50: should probably do some kind of sanitisation so that you can't do something like `cache.set('../../file_in_users_home', 'value')`

tpoliaw (changes requested):
The changes to require a visit when submitting the plan look ok.

I'm not sure all the caching stuff is really needed and if it is, it should probably be its own PR added separately.

src/blueapi/worker/task.py
```diff
@@ -19,18 +19,25 @@ class Task(BlueapiBaseModel):
     params: Mapping[str, Any] = Field(
         description="Values for parameters to plan, if any", default_factory=dict
     )
+    metadata: Mapping[str, Any] = Field(
+        description="Any metadata to apply to all runs within this task",
+        default_factory=dict,
+    )
```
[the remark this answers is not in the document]
callumforrester: Decided not important offline

callumforrester:
@tpoliaw and @DiamondJoseph I have moved the caching changes into https://github.com/DiamondLightSource/blueapi/pull/1062 so they don't block merge here. I think I've addressed all points that weren't related to that.

src/blueapi/cli/cli.py
```diff
@@ -231,6 +232,19 @@ def on_event(
...
+    "-i",
+    "--instrument-session",
+    type=str,
+    help=textwrap.dedent("""
+        Instrument session associated with running the plan,
+        used to tell blueapi where to store any data and as a security check:
+        the session must be valid and active and you must be a member of it.
+        If you have saved your current session with
+        blueapi controller set-instrument-session,
+        passing this will supersede that."""),
```
tpoliaw, line 245:
```suggestion
    help=textwrap.dedent("""
        Instrument session associated with running the plan,
        used to tell blueapi where to store any data and as a security check:
        the session must be valid and active and you must be a member of it."""),
```

tests/unit_tests/service/test_authentication.py
```diff
@@ -131,9 +131,9 @@ def test_processes_valid_token(
     inner(access_token=valid_token_with_jwt["access_token"])
 
 
-def test_session_cache_manager_returns_writable_file_path(tmp_path):
-    os.environ["XDG_CACHE_HOME"] = str(tmp_path)
-    cache = SessionCacheManager(token_path=None)
-    Path(cache._file_path).touch()
-    assert os.path.isfile(cache._file_path)
-    assert cache._file_path == f"{tmp_path}/blueapi_cache"
+def test_session_cache_manager_returns_writable_file_path(tmp_path: Path):
```
tpoliaw, line 134: Good change but not for this PR

tests/unit_tests/service/test_rest_api.py
```diff
@@ -310,15 +326,23 @@ def test_get_tasks(mock_runner: Mock, client: TestClient) -> None:
                 "is_complete": False,
                 "is_pending": True,
                 "request_id": None,
-                "task": {"name": "sleep", "params": {"time": 0.0}},
+                "task": {
+                    "name": "sleep",
+                    "params": {"time": 0.0},
+                    "metadata": {},
```
tpoliaw, line 332: Would these retain the visit when called against a real server?

tests/unit_tests/test_cli.py
```diff
@@ -346,14 +398,39 @@ def test_invalid_stomp_config_for_listener(runner: CliRunner):
...
+            "run",
+            "--background",
+            "sleep",
+            '{"time": 5}',
+        ],
+    )
+    assert result.exit_code == 2
+    assert "Error: Missing option '-i' / '--instrument-session'.\n" in result.stderr
+
+
```
tpoliaw, line 433: Could use a test to cover the environment variable equivalent

```python
def test_instrument_session_from_environment(runner: CliRunner):
    with patch.dict(os.environ, {"BLUEAPI_CONTROLLER_RUN_INSTRUMENT_SESSION": "cm12345-1"}, clear=True):
        # assert visit passed to rest
```

tpoliaw (commented):
Looks ok to me. Will let Joseph look at it again before approving it

PR #1067, opened 2025-06-18
fix: Ensure blueapi is editable install when using initContainer

DiamondJoseph:
Prior to this change, if using an initContainer to setupScratch, the editable install of blueapi that was created by the debug stage of the Dockerfile was overwritten by the non-editable install installed into the venv of the init container when the venv overwrote the main containers.


callumforrester (dismissed):
Please write a unit test, otherwise LGTM

PR #1075, opened 2025-06-19
feat: Make calls to numtracker async

tpoliaw:


callumforrester:
@tpoliaw the fact that this isn't making any tests fail is concerning

ZohebShaikh:
[This](https://github.com/DiamondLightSource/blueapi/blob/ced8f44d42f1b335c065f0e6764a365fca7db91f/tests/system_tests/test_blueapi_system.py#L193) call in system-test can be made async for consistency but won't have any effect.

PR #1077, opened 2025-06-19
chore: Update to copier template v4

DiamondJoseph:
Moves the copier template beyond the boundary that requires manual intervention: copier v4 and beyond does not require --trust and so can become another dependency managed by Renovate.


tpoliaw (commented):
Looks ok assuming nothing here was changed other than stuff that was changed in the template

tpoliaw (approved):

PR #1080, opened 2025-06-20
feat: Add click ParamType to handle JSON parsing

tpoliaw:
Argument parsing has been moved into a click option validator. This
lets clicks error handling deal with invalid JSON being passed to the
run_plan method instead of having it with the rest of the run logic.


DiamondJoseph (approved):
Nothing blocking so approving but had some comments

abbiemery (approved):
Appears to run plans and also raise errors when I expect it to.

PR #1084, opened 2025-06-25
feat!: Refactor auth code to use new cache

callumforrester:
Tidyup/refactor that shouldn't' block merge of #1062 

BREAKING CHANGE: `auth_token_path` has changed to `cache_path` in the configuration and now refers to a directory rather than a file. The directory contains arbitrary caching information including the auth token.


tests/unit_tests/utils/test_caching.py
```diff
@@ -0,0 +1,107 @@
...
+def test_caches_to_correct_file_path(cache: DiskCache, temp_dir: Path):
+    assert not (temp_dir / "foo").exists()
+    cache.set("foo", "bar")
+    assert (temp_dir / "foo").exists()
+
+
+def test_writes_b64_encoded_string(cache: DiskCache, temp_dir: Path):
+    cache.set("foo", "bar")
+    with (temp_dir / "foo").open("rb") as reader:
+        assert reader.read() == b"ImJhciI="
```
ZohebShaikh, line 36: Instead of magic letter Could use `base64encode` of `bar`

ZohebShaikh:
As Callum is working in other team,I have asked him and taken this PR over,I got to know that you (@tpoliaw) are not in favor of this PR, Can I know what was the reason...? I think this PR will be useful as there are currently 2 ways to set the instrument_session 
1. By -i in each call
2. By exporting it as a env variable might be difficult for novice users
This PR adds the 3rd way which is by setting it in the cache

With the order of precedence being 1 > 2 > 3 ...I think just letting the user know the instrument_session they are using will be helpful.


Currently I have just merged main and fixed a test which was having issue,

There was one thing apart from the comments above that I did not like was that it is creating 2 files for the 2 things `auth_token `and `instrument_session`, I think it will be better to just save all the cache into one file ...but it is not a strong opinion

DiamondJoseph:
Closing for now as 1.0.0 has come along and it's not important enough to break things right now.

PR #1086, opened 2025-06-26
fix: Force parameter_schema and version API

### LINKED ISSUE #308: Version the API and making parameter_schema non-optional
Sometime after #306 is merged (possibly just prior to the next release), we should increment the version of the API and make parameter_schema a non-optional field on PlanResponse.

DiamondJoseph:
Fixes #308. We already put the API_VERSION into the headers of all requests, but this is the final change before releasing v1.0


DiamondJoseph:
```
API_VERSION != HELM_VERSION
HELM_VERSION == CLI_VERSION == PYPI_VERSION
```

Is the decision that's been made between me and Callum at the least, I will document somewhere.

ZohebShaikh (approved):
Approving after offline discussion regarding this,

We might need to think of way how we will know which `API_VERSION ` the `CLI` is going to support

tpoliaw:
Should the schema change and 1.0 version be different PRs? Wtih the API version bumped at the same time as the app version.

DiamondJoseph:
So this changes to 0.2.0 and then once we're happy with 1.0 we have a PR just to increment API to 1.0?

src/blueapi/service/main.py
```diff
@@ -55,7 +55,7 @@
 from .runner import WorkerDispatcher
 
 #: API version to publish in OpenAPI schema
-REST_API_VERSION = "0.1.0"
+REST_API_VERSION = "0.2.0"
```
DiamondJoseph, line 58:
```suggestion
REST_API_VERSION = "0.1.1"
```

PR #1087, opened 2025-06-26
fix!: Remove ophyd

DiamondJoseph:
Removes our direct dependency on ophyd. This was intended to fix some spurious log messages on startup, but we still get ophyd as a dependency of dodal. It at least allows us to only have that dependency via dodal, and when dodal no longer requires ophyd it will not still be a trailing edge.

This is a breaking change to anyone using config generated from the default


tests/system_tests/config.yaml
```diff
@@ -0,0 +1,13 @@
+api:
+  url: http:/0.0.0.0:8000
```
DiamondJoseph, line 2:
```suggestion
  url: http://0.0.0.0:8000
```

Interesting that this still works

tests/unit_tests/core/test_context.py
```diff
@@ -551,11 +552,11 @@ class Derived(Base1[int], Base2):
         pass
 
     derived_instance = Derived()
-    assert generic_bounds(derived_instance, Base1) == (int,)  # type: ignore
+    assert generic_bounds(derived_instance, Base1) == (int,)  #  type: ignore
```
DiamondJoseph, line 555:
```suggestion
    assert generic_bounds(derived_instance, Base1) == (int,)  # type: ignore
```

tests/unit_tests/core/test_context.py
```diff
@@ -535,10 +536,10 @@ class Derived(Base[int]):
         pass
 
     derived_instance = Derived()
-    assert generic_bounds(derived_instance, Base) == (int,)  # type: ignore
+    assert generic_bounds(derived_instance, Base) == (int,)  #  type: ignore
```
DiamondJoseph, line 539:
```suggestion
    assert generic_bounds(derived_instance, Base) == (int,)  # type: ignore
```

tests/unit_tests/core/test_context.py
```diff
@@ -13,9 +14,17 @@
     Stoppable,
     SyncOrAsync,
 )
+from bluesky.run_engine import RunEngine
 from bluesky.utils import MsgGenerator
 from dodal.common import PlanGenerator, inject
-from ophyd.sim import SynAxis, SynGauss
+from ophyd_async.core import (
```
tpoliaw, line 20: we should probably have ophyd-async in our dev dependencies if we rely on it directly

tpoliaw (approved):
Could almost do with a separate change for all the unrelated test typing fixes

PR #1089, opened 2025-06-27
Convert to using uv

tpoliaw:
- **Use dependency-group instead of optional-dependencies**
- **Specify python version upper bound**
- **Initial workflow rewrite**


abbiemery (changes requested):
Please can you change/add to the docs to reflect the new uv commands?

.devcontainer/devcontainer.json
```diff
@@ -48,5 +59,5 @@
     // Mount the parent as /workspaces so we can pip install peers as editable
     "workspaceMount": "source=${localWorkspaceFolder}/..,target=/workspaces,type=bind",
     // After the container is created, install the python project in editable form
-    "postCreateCommand": "pip install $([ -f dev-requirements.txt ] && echo '-c dev-requirements.txt') -e '.[dev]' && pre-commit install"
+    "postCreateCommand": "uv sync --frozen && uvx pre-commit install && SHELL=bash uv tool update-shell"
```
coretl, line 62: Some questions:
- Why `--frozen`? I would argue it is safer to do it without, as that will potentially update your lockfile if you added a dependency to `pyproject.toml` which you should at least know about before deciding whether it was intentional or not and committing the result
- Why `uvx` rather than `uv run`? Pre-commit should be listed in the dev dependencies, and this will ensure we get a consistent version of it.
- Do we use any uv tools? Why did you need to run this command to add the tool dir to the path?

`uv sync` I get, `--frozen` I think makes sense, but why `uvx pre-commit install` rather than `uv run pre-commit install`?
Also what does

.github/actions/install_requirements/action.yml
coretl, line 1: I've deleted this action in my version as I use a `.python-version` to determine the default python, then `uv run` everywhere

.github/workflows/_dist.yml
```diff
@@ -24,11 +27,12 @@ jobs:
           path: dist
 
       - name: Check for packaging errors
-        run: pipx run twine check --strict dist/*
+        run: uvx twine check --strict dist/*
 
       - name: Install produced wheel
         uses: ./.github/actions/install_requirements
         with:
+          # TODO: figure out what's going on
```
coretl, line 35: I haven't checked dists yet, what problems do you see?

.github/workflows/_system_test.yml
```diff
@@ -50,4 +50,4 @@ jobs:
         run: blueapi -c ${{ github.workspace }}/tests/system_tests/config.yaml serve &
 
       - name: Run tests
-        run: tox -e system-test
+        run: uv run --frozen tox -e system-test
```
coretl, line 53: Hmm, maybe `--frozen` does make sense here, but maybe `--locked` would make even more sense? What do you reckon?

Dockerfile
```diff
@@ -3,31 +3,33 @@
 # Version SHA has been removed, see: https://github.com/DiamondLightSource/blueapi/issues/1053
 ARG PYTHON_VERSION=3.11
 FROM python:${PYTHON_VERSION} AS developer
```
coretl, line 5: I switched to ubuntu and use a uv managed python, this seems to work now

Dockerfile
```diff
@@ -43,8 +45,8 @@ RUN DEBIAN_FRONTEND=noninteractive apt install libnss-ldapd -y
 RUN sed -i 's/files/ldap files/g' /etc/nsswitch.conf
 
 # Make editable and debuggable
-RUN pip install debugpy
-RUN pip install -e .
+RUN uv tool install debugpy
```
coretl, line 48: Ah, this is why you need uv tools to be on your path, but I don't see why you need it in the devcontainer.json too?

pyproject.toml
```diff
@@ -38,9 +38,19 @@ dependencies = [
 dynamic = ["version"]
 license.file = "LICENSE"
 readme = "README.md"
-requires-python = ">=3.11"
+requires-python = ">=3.11,<3.13"
```
coretl, line 41: The current best practice appears to be never to cap python version for libraries:
https://iscinumpy.dev/post/bound-version-constraints/#tldr

I would argue it probably makes sense here too...

.devcontainer/devcontainer.json
tpoliaw: Not sure I follow the frozen comment. `--frozen` doesn't update the lockfile and ensures the environment you have is the same as the one expected. Without it I couldn't get it to not need `--prerelease=allow` to deal with dodal et al depending on alpha versions of things and I didn't want that to be something we encourage.

I can't remember the motivation for `uvx` rather than `uv run`, might have been related to making pre-commit available everywhere without setting the `$PATH` to the venv as you did for ophyd-async. I will try it without and see what fails.

> Also what does

I'm assuming there was meant to be more to this question

.github/actions/install_requirements/action.yml
tpoliaw: That'd be useful, I'll see if that's a viable option here

.github/workflows/_dist.yml
tpoliaw: None but I had no idea what to look for. The comment was more that I had no idea

.github/workflows/_system_test.yml
tpoliaw: `--locked` fails if any packages are updatable so would fail pretty much every run, `--frozen` makes it use the versions in the lock file.

pyproject.toml
tpoliaw: This was probably a hang over from epics(?) not supporting 3.13 and can be removed now, especially if we're using `.python-version`. I'm pretty sure `uv` ignores upper bounds anyway.

Dockerfile
coretl: Difficult to tell, the download takes a fraction of a second on my machine, so I miss the output. I think it is downloaded each time, but I can't be sure

PR #1110, opened 2025-07-08
feat: Add values json schema

### LINKED ISSUE #1095: No values.schema.json
As a developer I would like to have a values schema to easily understand required and optional config, and to inform my dev environment tools.

## Acceptance Criteria
- A `values.schema.json` file exists
- The current `values.yaml` validates against it

dan-fernandes:
Fixed #1095

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -23,7 +23,7 @@ colorama==0.4.6
 colorlog==6.9.0
 compress-pickle==2.1.0
 contourpy==1.3.3
-copier==9.8.0
+copier==9.9.1
 coverage==7.10.1
 cryptography==45.0.5
 cycler==0.12.1


.github/workflows/_test.yml
```diff
@@ -53,6 +53,9 @@ jobs:
         with:
           python-version: ${{ inputs.python-version }}
           pip-install: ".[dev]"
+      
+      - name: Install helm plugins
+        uses: ./.github/actions/install_helm_plugins
```
DiamondJoseph, line 58: Do we need this? Do our tests require that the plugin is available? If they do I still think maintaining a whole action just to run one command rather than having repetition is a little overkill.

.pre-commit-config.yaml
```diff
@@ -24,10 +24,23 @@ repos:
...
+      - id: helm-docs-built
         args:
           # Make the tool search for charts only under the `helm` directory
           - --chart-search-root=helm
+
+  - repo: https://github.com/losisin/helm-values-schema-json
+    rev: v1.7.2
+    hooks:
+      - id: helm-schema
+        args: ["--values", "values.yaml"]
```
DiamondJoseph, line 46: Does this not need any guidance to find the Chart?

.pre-commit-config.yaml
```diff
@@ -24,10 +24,23 @@ repos:
         types: [python]
         require_serial: true
 
+      - id: export-config-schema
+        name: export config schema
+        entry: blueapi config-schema -u
+        language: python
+        files: "src\/blueapi\/config.py|src\/blueapi\/utils\/"
+        pass_filenames: false
```
DiamondJoseph, line 32:
```suggestion
```

I think we can replace this with a test like `test_schema_updated` to ensure that config_schema.json is always kept update

DiamondJoseph (changes requested):

.github/workflows/_test.yml
dan-fernandes: I think it is required for the tests, as they run the pre-commit hooks in a new container.
I have removed the action and directly run the install command where relevant.

.pre-commit-config.yaml
dan-fernandes: No idea, but to be safe I've put a more exact path.

src/blueapi/cli/cli.py
```diff
@@ -108,6 +108,31 @@ def schema(output: Path | None = None, update: bool = False) -> None:
...
+    schema = ApplicationConfig.model_json_schema()
+
+    if update:
+        output = config.CONFIG_SCHEMA_LOCATION
+    if output is not None:
+        with output.open("w") as stream:
+            json.dump(schema, stream, indent=4)
+            stream.write("\n")
+    else:
+        print(json.dumps(schema))
```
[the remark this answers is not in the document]
DiamondJoseph: If we do, should then move them from openapi.py to cli.py and make them private?

tests/unit_tests/test_cli.py
```diff
@@ -1281,3 +1282,48 @@ def test_python_env_output_formatting():
...
+    if output_flag and (not update):
+        with tmp_path.open("r") as stream:
+            assert (
+                DeepDiff(json.load(stream), ApplicationConfig.model_json_schema()) == {}
+            )
+
+    elif update:
+        config_schema_location_mock.open.assert_called()
+        with config_schema_location_mock.open() as stream:
+            stream.write.assert_called()
```
DiamondJoseph, line 1323: I tried doing something clever with `stream.write.call_args` to check that result.output is still caught, but because of how json.dumps works we only actually capture the (`\n`) which is annoying.

tests/unit_tests/test_cli.py
```diff
@@ -1281,3 +1282,48 @@ def test_python_env_output_formatting():
...
+
+    else:
+        assert (
+            DeepDiff(json.loads(result.output), ApplicationConfig.model_json_schema())
+            == {}
+        )
```
DiamondJoseph, line 1329:
```suggestion
    expected = ApplicationConfig.model_json_schema()

    if output_flag and (not update):
        with tmp_path.open("r") as stream:
            assert json.load(stream) == expected

    elif update:
        config_schema_location_mock.open.assert_called()
        with config_schema_location_mock.open() as stream:
            stream.write.assert_called()

    else:
            assert json.load(result.output) == expected
```

This works when I run it locally and prevents importing DeepDiff

tests/unit_tests/test_config.py
```diff
@@ -500,3 +506,17 @@ def validate_field_annotations(model_class: Any, model_field: str) -> None:
...
+    reason="If the schema file does not exist, the test is being run"
+    " with a non-editable install",
+)
+def test_config_schema_updated() -> None:
+    with CONFIG_SCHEMA_LOCATION.open("r") as stream:
+        config_schema = json.load(stream)
+    assert DeepDiff(config_schema, ApplicationConfig.model_json_schema()) == {}, (
+        f"ApplicationConfig model is out of date with schema at \
+            {CONFIG_SCHEMA_LOCATION}. You may need to run `blueapi config-schema -u`"
+    )
```
DiamondJoseph, line 522:
```suggestion
    assert config_schema == ApplicationConfig.model_json_schema(), (
        f"ApplicationConfig model is out of date with schema at \
            {CONFIG_SCHEMA_LOCATION}. You may need to run `blueapi config-schema -u`"
    )
```

helm/blueapi/.schema.yaml
```diff
@@ -0,0 +1,22 @@
+# .schema.yaml
```
DiamondJoseph, line 1: Which of these fields are required? Can we get away with using the defaults? e.g. noAdditionalProperties- why false but then additionalProperties: false in schema root? And why id of example.com?

DiamondJoseph (changes requested):

tests/unit_tests/test_cli.py
dan-fernandes: LGTM, but I think line 1324 should be `json.loads`

.github/workflows/_test.yml
dan-fernandes: You're right - it's been removed

PR #1113, opened 2025-07-08
fix(devcontainer): Use non-containerised helm-docs

dan-fernandes:
The current pre-commit hooks cannot all be run from inside the dev container.

By changing this id we target a hook that is compatible.
As the built version does not have a version, the `helm-docs.version-footer` was causing unexpected behaviour, so it has been removed.


DiamondJoseph:
@dan-fernandes try running pre-commit locally and see if it passes CI?

PR #1118, opened 2025-07-10
test: Add system tests using NumTracker

DiamondJoseph:
Necessarily picks up dodal 1.54 for the adsim module to be able to use the PathProvider that we want to, and ophyd-async 0.12.1 for a bug fix in the ADHDFWriter

### DIFF
--- a/tests/system_tests/test_blueapi_system.py
+++ b/tests/system_tests/test_blueapi_system.py
@@ -54,6 +54,13 @@ For more details, see: https://github.com/DiamondLightSource/blueapi/issues/676.
 To enable and execute these tests, set `REQUIRES_AUTH=1` and provide valid credentials.
 """
 
+# These system tests are primarily intended to be run from the Github CI,
+# and therefore these instructions may require multiple terminals, running
+# outside of the officially support devcontainer and other inefficiencies.
+#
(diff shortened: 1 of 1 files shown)


tpoliaw (commented):
Looks ok other for when dodal's been released

tests/system_tests/test_blueapi_system.py
```diff
@@ -51,13 +51,18 @@
...
-# Step 1: Ensure a message bus that supports stomp is running and available:
-#   src/script/start_rabbitmq.sh
+# Start devices
+#   1. clone https://github.com/epics-containers/example-services
+#   2. cd example-services
+#   2. docker compose up bl01t-di-cam-01 bl01t-mo-sim-01 ca-gateway
 #
-# Step 2: Start the BlueAPI server with valid configuration:
-#   blueapi -c tests/system_tests/config.yaml serve
+# Start services
```
tpoliaw, line 59: Can it be made a bit cleared that the rest of these commands are back in blueapi not still in example-services from above

DiamondJoseph: How's the new wording/arrangment?

tpoliaw: much better, thanks

callumforrester (approved):

PR #1122, opened 2025-07-14
chore(deps): Update dependencies

DiamondJoseph:
Update dependencies manually while awaiting #1089 to pick up updates to ScanSpec that allow us to use Pydantic 2.11 and changes to our own observability-utils and bluesky-stomp.


src/blueapi/service/main.py
tpoliaw, line 1: Are these changes meant to be included in the dependency updates?

PR #1124, opened 2025-07-14
fix: Better handling for optional arguments

DiamondJoseph:
To be merged into main after #1122 rather than into #1122 just arranged for convenient review.

Adjusts handling of optional arguments in plan parameters to use the `required` field from the JsonSchema only instead of additionally explicitly allowing `null` as an argument. This makes the schema easier to read and automated tools should work better with it.


tests/system_tests/plans.json
```diff
@@ -89,7 +82,7 @@
                             },
                             "radius": {
                                 "description": "Radius of the circle",
-                                "exclusiveMinimum": 0,
+                                "exclusiveMinimum": 0.0,
```
DiamondJoseph, line 85: I do not know why these changes have appeared here just from adjusting handling of `| None`, as they do not allow None or defaults... I have always been confused why they got set to integer 0 instead of float 0.0, and I think this is more correct but why!?

```python
StrictConfig: ConfigDict = {"extra": "forbid", "arbitrary_types_allowed": True}

@dataclass(config=StrictConfig)
class Circle(Region[Axis]):
      [...]
      radius: float = Field(description="Radius of the circle", gt=0.0)

```

DiamondJoseph:
> I'm not sure what I think of `int | None` being treated differently to `int | None = None` but I can't think of a better way and this does make the schema more useful in the majority of cases.

`int | None = None` -> This argument is optional, if you provide it it must be an int
`int | None` -> This argument can be an int or None, but must be provided

Can I think of any examples where they are using the 2nd and shouldn't actually be using the first? Not right now.

But if that is what your plan should have, cutting the possibility of passing None means you must pass an int.

PR #1130, opened 2025-07-17
refactor(container): Split dev and debug requirements

DiamondJoseph:
Refactors the container build so that the non-debug image can be used to debug traces outside of the cluster using `kubectl exec` and `kubectl port-forward`, and allow editing when using `pv-mounter` to mount the scratch dir locally.

This allows for seperate `debug` and `development` workflows: when debugging plans, the debugger can be used to step through and examine variables in the namespace, and when developing local tooling can be used. 

The devcontainer attach workflow is for now preserved, and `-debug` tag to mark an image as allow debugging of the service itself, separate from the ability to debug the plans and devices.

### DIFF
--- a/.github/workflows/_container.yml
+++ b/.github/workflows/_container.yml
@@ -49,6 +49,7 @@ jobs:
         with:
           context: .
           # Need load and tags so we can test it below
+          target: runtime
           load: true
           tags: tag_for_testing
 
@@ -94,6 +95,7 @@ jobs:
         with:
           context: .
           push: true
+          target: runtime
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,32 +25,18 @@ FROM developer AS build
 RUN mkdir -p /.cache/pip; chmod o+wrX /.cache/pip
 COPY --chmod=o+wrX . /workspaces/blueapi
 WORKDIR /workspaces/blueapi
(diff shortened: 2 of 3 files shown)


.vscode/launch.json
```diff
@@ -15,6 +15,27 @@
             ],
             "console": "integratedTerminal",
         },
+        {
+            "name": "Attach debugger",
+            "type": "debugpy",
+            "request": "attach",
+            "justMyCode": false,
+            "connect": {
+                "host": "0.0.0.0",
```
keithralphs: Can we make this "localhost" for consistency (I've checked it and it still connects)

.vscode/tasks.json
```diff
@@ -11,6 +11,34 @@
...
+        {
+            "type": "shell",
+            "label": "kubectl port-forward",
+            "command": "kubectl",
+            "args": [
+                "port-forward",
+                "-n",
+                "${input:namespace}",
+                "${input:pod}",
+                "5678",
```
keithralphs: I think this has to be "5678:5678" or it just uses some rando port and then you have to change your launch config to the use the port it chooses. Presumably we could then add "preLaunchTask": "kubectl port-forward" back into the launch .json as it is still necessary to do the port forward manually in a terminal to get it to work at the moment.

.vscode/launch.json
keithralphs: It is no longer necessary to include these in this change as the the launch config will be defined in the workspace file going forward

PR #1155, opened 2025-08-08
Add example services as a submodule

callumforrester:
Add example services as a submodule and include it in the system tests docker compose file, rewrite the system test docs to make use of this.

The system tests will now be easier to run and debug locally, since this automates a lot of the setup and tear down process.


tests/system_tests/test_blueapi_system.py
```diff
@@ -101,6 +97,24 @@ def client_with_stomp() -> BlueapiClient:
     )
 
 
+@pytest.fixture(scope="module", autouse=True)
+def epics_config_environment_variables():
+    ca_port = 9064
+    pva_port = 9075
+    localhost = "127.0.0.1"
```
ZohebShaikh, line 104: Can the .env file be used to get these value example [here](https://dev.to/jakewitcher/using-env-files-for-environment-variables-in-python-applications-55a1) ? 
It's a draft PR so I might be reviewing a unfinished PR

ZohebShaikh (commented):
I personally don't like having submodule`s for test but if this is the approach that we are taking then it should be atleast consistent among the athena repo system test (dodal,etc) and a ADR added to python copier template to make sure that it is documented that we are going to use submodules for system test so people are find a consistent way of running system test locally.

Also is there no way that this can be done inside the same vscode rather than having a different terminal outside the devcontainer.

callumforrester:
I am not hugely in love with submodules in general, but the previous set of instructions was getting silly and was only going to get more complicated over time. I really do feel that we need to keep the process for spinning up a dummy environment relatively easy and simple so developers can just do it rather than spend 10 minutes faffing around before they can do so, especially as it is not just for system tests but for playing around with a local blueapi server. Now that there are no dummy devices.

Having to run outside of the devcontainer is really annoying but from talking to @DiamondJoseph and @gilesknap I don't think they're confident that podman will ever support it enough for us to trust it (they can correct me if I'm wrong though).

gilesknap:
> Having to run outside of the devcontainer is really annoying but from talking to @DiamondJoseph and @gilesknap I don't think they're confident that podman will ever support it enough for us to trust it (they can correct me if I'm wrong though).

I feel we have had this conversation before (more than once?) and that perhaps your use case here has issues. But in general docker compose based developer containers will work fully at DLS if you do the following setup https://epics-containers.github.io/main/how-to/compose-quickstart.html#diamond-light-source-workstation

DiamondJoseph:
I was going to approve (sans one nit, it's spelt "associated"), I've just been looking at docker-in-docker, [which may support our Ubuntu base images soon](https://github.com/devcontainers/features/pull/1442) to see if we can get an even simpler workflow soon

PR #1166, opened 2025-08-14
feat: Add logout redirect

ZohebShaikh:
Currently we can logout by going to this https://p46-blueapi.diamond.ac.uk/oauth2/sign_out?rd=https%3A%2F%2Fauthn.diamond.ac.uk%2Frealms%2Fmaster%2Fprotocol%2Fopenid-connect%2Flogout. It will be much simpler to logout by going to https://p46-blueapi.diamond.ac.uk/logout. This is when we have oauth2-proxy enabled for blueapi docs page

### DIFF
--- a/docs/how-to/authenticate.md
+++ b/docs/how-to/authenticate.md
@@ -1,4 +1,8 @@
-# Authenticate to BlueAPI
+> [!NOTE]
+> If you are using `oauth2-proxy` to secure the Swagger UI documentation page, you can log out by visiting the `/logout` URL. For this to work correctly, ensure that the blueapi server is configured with 
+> `oidc.logout_redirect_endpoint` set to `/oauth2/sign_out`, which is required for `oauth2-proxy`.
+
+# Authenticate to BlueAPI-Cli
 
 ## Introduction
 BlueAPI provides a secure and efficient way to interact with its services. This guide walks you through the steps to log in and log out using BlueAPI with OpenID Connect (OIDC) authentication.
@@ -63,9 +67,3 @@ To log out and securely remove the cached access token, follow these steps:
    ```
    Logged out
    ```
-
-
-> [!NOTE]
-> The login and logout instructions above apply to the CLI. If you are using `oauth2-proxy` to secure the Swagger
-> UI documentation page, you can log out by visiting the `/logout` URL. For other OIDC providers, update the 
-> `oidc.logout_redirect_endpoint` configuration to the appropriate logout endpoint.
--- a/docs/how-to/authenticate.md
+++ b/docs/how-to/authenticate.md
@@ -63,3 +63,9 @@ To log out and securely remove the cached access token, follow these steps:
    ```
    Logged out
    ```
+
+
+> [!NOTE]
+> The login and logout instructions above apply to the CLI. If you are using `oauth2-proxy` to secure the Swagger
+> UI documentation page, you can log out by visiting the `/logout` URL. For other OIDC providers, update the 
+> `oidc.logout_redirect_endpoint` configuration to the appropriate logout endpoint.
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -162,6 +162,9 @@ class OIDCConfig(BlueapiBaseModel):
     )
     client_id: str = Field(description="Client ID")
     client_audience: str = Field(description="Client Audience(s)", default="blueapi")
+    logout_redirect_endpoint: str = Field(
+        description="The oidc endpoint required to logout", default="/oauth2/sign_out"
+    )
 
     @cached_property
     def _config_from_oidc_url(self) -> dict[str, Any]:
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -162,8 +162,8 @@ class OIDCConfig(BlueapiBaseModel):
     )
     client_id: str = Field(description="Client ID")
     client_audience: str = Field(description="Client Audience(s)", default="blueapi")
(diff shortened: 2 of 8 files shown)


src/blueapi/service/main.py
```diff
@@ -56,7 +57,7 @@
 from .runner import WorkerDispatcher
 
 #: API version to publish in OpenAPI schema
-REST_API_VERSION = "1.0.2"
+REST_API_VERSION = "1.0.3"
```
tpoliaw, line 60: Should adding new features be a minor bump?

src/blueapi/service/main.py
```diff
@@ -534,6 +535,19 @@ def health_probe() -> HealthProbeResponse:
     return HealthProbeResponse(status=Health.OK)
 
 
+@secure_router.get("/logout", status_code=status.HTTP_200_OK, include_in_schema=False)
+def logout(runner: Annotated[WorkerDispatcher, Depends(_runner)]) -> RedirectResponse:
+    """Redirect to logout url"""
+    config = runner.run(interface.get_oidc_config)
+    if config is None:
+        raise HTTPException(status_code=status.HTTP_204_NO_CONTENT)
```
tpoliaw, line 543: Returning a 2xx status with an error response seems odd?

src/blueapi/service/main.py
```diff
@@ -534,6 +535,19 @@ def health_probe() -> HealthProbeResponse:
     return HealthProbeResponse(status=Health.OK)
 
 
+@secure_router.get("/logout", status_code=status.HTTP_200_OK, include_in_schema=False)
```
tpoliaw, line 538: when is the 200 ever used?

tpoliaw (commented):
Initial thoughts - I don't really know how the auth redirect stuff is intended to work so may not be relevant.

src/blueapi/service/main.py
ZohebShaikh: Good catch will remove it

src/blueapi/service/main.py
ZohebShaikh: This should really never happen, this is the same pattern that is used for /config/oidc.

This endpoint is created when there is a config,This is more for the type checking to be happy

docs/how-to/authenticate.md
```diff
@@ -63,3 +63,9 @@ To log out and securely remove the cached access token, follow these steps:
    ```
    Logged out
    ```
+
+
+> [!NOTE]
+> The login and logout instructions above apply to the CLI. If you are using `oauth2-proxy` to secure the Swagger
+> UI documentation page, you can log out by visiting the `/logout` URL. For other OIDC providers, update the 
+> `oidc.logout_redirect_endpoint` configuration to the appropriate logout endpoint.
```
DiamondJoseph, line 71: Move enough of this to be understood into the introduction block at the top of the file.

src/blueapi/config.py
```diff
@@ -162,6 +162,9 @@ class OIDCConfig(BlueapiBaseModel):
     )
     client_id: str = Field(description="Client ID")
     client_audience: str = Field(description="Client Audience(s)", default="blueapi")
+    logout_redirect_endpoint: str = Field(
+        description="The oidc endpoint required to logout", default="/oauth2/sign_out"
+    )
```
DiamondJoseph, line 167:
```suggestion
    logout_redirect_endpoint: str | None = Field(
        description="The oidc endpoint required to logout", default=None
    )
```

We shouldn't redirect unless we have been configured to do so- this default can go into the template, but now we won't accidentally log people out of unrelated services.

src/blueapi/service/main.py
```diff
@@ -534,6 +535,19 @@ def health_probe() -> HealthProbeResponse:
     return HealthProbeResponse(status=Health.OK)
 
 
+@secure_router.get("/logout", include_in_schema=False)
+def logout(runner: Annotated[WorkerDispatcher, Depends(_runner)]) -> RedirectResponse:
```
DiamondJoseph, line 539:
```suggestion
def logout(runner: Annotated[WorkerDispatcher, Depends(_runner)]) -> Response:
```

src/blueapi/service/main.py
```diff
@@ -534,6 +535,19 @@ def health_probe() -> HealthProbeResponse:
...
+    """Redirect to logout url"""
+    config = runner.run(interface.get_oidc_config)
+    if config is None:
+        raise HTTPException(status_code=status.HTTP_204_NO_CONTENT)
```
DiamondJoseph, line 543:
```suggestion
        return Response(status_code=status.HTTP_205_RESET_CONTENT)
```

DiamondJoseph (changes requested):
Requesting some low hanging fruit while I grapple with the test

src/blueapi/config.py
```diff
@@ -162,6 +162,9 @@ class OIDCConfig(BlueapiBaseModel):
     )
     client_id: str = Field(description="Client ID")
     client_audience: str = Field(description="Client Audience(s)", default="blueapi")
+    logout_redirect_endpoint: str | None = Field(
```
DiamondJoseph, line 165:
```suggestion
    logout_redirect_endpoint: str | SkipJsonSchema[None] = Field(
```
Will this cause the apischema.yaml to not have

```yaml
        logout_redirect_endpoint:
          anyOf:
          - type: string
          - type: 'null'
```

the string or null should be captured in the required fields which is already correct

DiamondJoseph (commented):
Just one nit

DiamondJoseph (approved):

PR #1200, opened 2025-09-12
refactor: Move numtracker into the BlueskyContext

tpoliaw:
Moving the numtracker into the bluesky context means we don't need to do the 'create context -> create numtracker -> configure context -> check path provider` shenanigans in the top level setup function and the path provider can be available to pass to device factories if we move to a dodal-less approach to devices.

### DIFF
--- a/src/blueapi/core/context.py
+++ b/src/blueapi/core/context.py
@@ -118,6 +118,10 @@ class BlueskyContext:
         if self.numtracker is not None:
             numtracker = self.numtracker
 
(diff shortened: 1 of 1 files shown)


callumforrester:
@tpoliaw while I like the reduction in shenanigans I was already worried about how much of a god object `BlueskyContext` was becoming. I don't think that should block merge but it might be useful to have some thoughts (and make a ticket) about breaking it up

PR #1204, opened 2025-09-15
fix: Unsubscribe from events when worker stops

tpoliaw:


tpoliaw:
Possibly. It's a bit more awkward there with the current design because there's nowhere to hold the subscription. I'll have a look at what can be done. This is all about to change in #1200 so it might be easier to do it there.

PR #1205, opened 2025-09-15
refactor: Remove redundant teardown from interface tests

tpoliaw:
The teardown method is called for all tests in the test_interface module
by the ensure_worker_stopped fixture.


callumforrester (approved):
Good spot

PR #1214, opened 2025-09-25
chore(deps): Update to 5.0 pre-release of copier template

tpoliaw:
Uses uv for python management

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,8 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:best-practices",
-    ":automergeStableNonMajor"
+    "config:recommended"
   ],
   "allowScripts": true,
   "ignoreScripts": false,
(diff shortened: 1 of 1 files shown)


coretl:
Issues:
- https://github.com/DiamondLightSource/python-copier-template/issues/294
- https://github.com/DiamondLightSource/python-copier-template/issues/295
- Need to restore the checkout of main branch
- Need to set pytest default paths to be `src docs tests/unit_tests` instead of `src docs tests`

.devcontainer/devcontainer.json
```diff
@@ -33,20 +49,42 @@
...
+            "source": "persistent-pre-commit-cache",
+            "target": "/root/.cache/pre-commit",
+            "type": "volume"
+        }
+        // Use a volume mount for the uv venv so it is local to the container
+        {
+            "target": "/workspaces/${localWorkspaceFolderBasename}/.venv",
+            "source": "devcontainer-shared-cache",
+            "type": "volume"
+        }
```
tpoliaw: This has been changed upstream to use a single cache for everything, do we want to keep in sync with that?

Issue #1210, opened 2025-09-24
`StartDocumentPathProvider` does not clear its state after a run


callumforrester:
Describe the bug, including a clear and concise description of the expected behaviour, the actual behavior and the context in which you encountered it (ideally include details of your environment).

## Steps To Reproduce
1. Call `prepare` outside of a run, it should complain about a lack of metadata
2. Now run a normal scan
3. Call `prepare` again, it should now work 

## Acceptance Criteria
- Calling `prepare` outside of a run should always fail

PR #1221, opened 2025-10-01
fix: Ensure path provider clears start doc after run

abbiemery:
Fixes https://github.com/DiamondLightSource/blueapi/issues/1210

Also now handles nested runs.


src/blueapi/utils/path_provider.py
```diff
@@ -18,14 +18,19 @@ class StartDocumentPathProvider(PathProvider):
...
+    def run_start(self, name: str, start_document: RunStart) -> None:
         if name == "start":
-            self._doc = start_doc
+            if self._doc == {} and self._uid is None:
+                self._doc = start_document
+                self._uid = start_document["uid"]
+
+    def run_stop(self, name: str, stop_document: RunStop) -> None:
+        if name == "stop":
+            if stop_document.get("doc", {}).get("run_start") == self._uid:
```
tpoliaw, line 31: Why does this have the extra `doc` layer?

src/blueapi/utils/path_provider.py
```diff
@@ -18,14 +18,19 @@ class StartDocumentPathProvider(PathProvider):
 
     def __init__(self) -> None:
         self._doc = {}
+        self._uid = None
```
tpoliaw, line 21: Can we use `None` for the 'there is no document' case as well? Having an empty dict is odd

tests/unit_tests/utils/test_path_provider.py
```diff
@@ -192,18 +193,152 @@ def test_start_document_path_provider_sets_data_session_directory_default_to_tmp
...
+        "name": "stop",
+        "doc": {
+            "run_start": "27c48d2f-d8c6-4ac0-8146-fedf467ce11f",
+            "time": 1741264732.96875,
+            "uid": "401ad197-5456-4a7d-ba5b-9cf8ad38d914",
+            "exit_status": "success",
+            "reason": "",
+            "num_events": {"primary": 1},
+        },
+    }
```
tpoliaw, line 216: I think there is a mismatch here between the [`RunStop`](https://blueskyproject.io/event-model/main/explanations/data-model.html#run-stop-document) that bluesky sends from the RunEngine and the [`DataEvent`](https://github.com/DiamondLightSource/blueapi/blob/b8b2adf5a86ac7e6d35f476529893fdd45d6b2c2/src/blueapi/core/bluesky_types.py#L98) wrapping the `RunStop` that we send from server to client.

```suggestion
    return {
        "run_start": "27c48d2f-d8c6-4ac0-8146-fedf467ce11f",
        "time": 1741264732.96875,
        "uid": "401ad197-5456-4a7d-ba5b-9cf8ad38d914",
        "exit_status": "success",
        "reason": "",
        "num_events": {"primary": 1},
    }
```

tpoliaw (changes requested):

src/blueapi/utils/path_provider.py
DiamondJoseph, line 21:
```suggestion
        self._doc: RunStart | None = None
        self._uid: str | None = None
```

abbiemery: Yep, indeed. Good shout.

DiamondJoseph (approved):
Just some nits, I'm happy to get this merged tomorrow when you're back in

Issue #1225, opened 2025-10-03
bluesky version incompatibilities


DiamondJoseph:
Adding Tiled as a dependency has caused a conflict in the supported version of bluesky.

The full release `tiled>=0.1.0` requires but does not explicitly denote `bluesky>=1.14.4`
dodal currently requires `bluesky<=1.14.2`

Either we must have a release of bluesky 1.14.5, which is compatible with both dodal and tiled
Or we must pin tiled to a beta release v0.1.0-b35
Or (temporarily, to prevent tests from not catching regressions?), we can make the tiled import lazy to prevent this being an issue when not using Tiled and explicitly call out this incompatibility

## Acceptance Criteria
- Blueapi can start up

PR #1226, opened 2025-10-03
fix: Fix bluesky dependency version requirements of tiled and dodal

DiamondJoseph:
Fxes #1225

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "ophyd-async",
     "aioca",
     "pydantic>=2.0",
-    "scanspec>=0.8.0",
+    "scanspec>=0.9.0",
     "pydantic-settings",
     "stomp-py",
     "PyYAML>=6.0.2",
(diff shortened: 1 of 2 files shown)


tpoliaw:
It doesn't look like this change has anything to do with tiled?

dev-requirements.txt
```diff
@@ -31,12 +32,13 @@ copier==9.10.2
 coverage==7.10.7
 cryptography==46.0.2
 cycler==0.12.1
+daq-config-server==1.0.0rc2
```
callumforrester, line 35: Eee, wonder what pulled this in

callumforrester (approved):

DiamondJoseph:
> It doesn't look like this change has anything to do with tiled?

See ticket #1225 

tiled>=0.1.0 requires to function but not explicitly bluesky>=1.14.4
dodal<1.60.0 requires bluesky<=1.14.2

So we needed dodal>=1.60.0 which is compatible with bluesky>=1.14.5

PR #1242, opened 2025-10-17
feat: Support for dataclasses as composite types

rtuck99:
This modifies the previous PR #1231 and adds the following:

* Simplifies the creation of composites, the composite no longer needs to be an instance of `BaseModel` and instead can be a dataclass
* The composite does not need to specify a default and instead determines the device to inject from the attribute name

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -54,7 +54,7 @@ fastapi-cloud-cli==0.3.1
 filelock==3.20.0
 flexcache==0.3
 flexparser==0.4
-fonttools==4.60.1
+fonttools==4.61.0
 frozenlist==1.8.0
 fsspec==2025.10.0
 funcy==2.0


docs/how-to/write-plans.md
```diff
@@ -24,7 +24,24 @@ The type annotations in the example above (e.g. `: str`, `: int`, `-> MsgGenerat
 
 ## Injecting Devices
 
-Some plans are created for specific sets of devices, or will almost always be used with the same devices, it is useful to be able to specify defaults. [Dodal makes this easy with its factory functions](https://diamondlightsource.github.io/dodal/main/how-to/include-devices-in-plans.html).
+Some plans are created for specific sets of devices, or will almost always be used with the same devices, it is useful to be able to specify defaults. [Dodal makes this easy with its inject function](https://diamondlightsource.github.io/dodal/main/reference/generated/dodal.common.html#dodal.common.inject).
+
+## Injecting multiple devices
+
+If a plan requires multiple devices to be injected at once, rather than have a plan with several device parameters each of them with their own injection default, it is preferable to define a device composite which can be accepted as a parameter.
```
DiamondJoseph: I don't agree that it is preferable- if you require overriding one device you must override the whole Composite. 

I just think it's an alternative option that may be useful quite often, depending on how a specific beamline operates.

I'll look through the code change later, but I am admitting upfront that I am opinionated

src/blueapi/core/context.py
```diff
@@ -387,16 +392,19 @@ def _type_spec_for_function(
...
 
             no_default = para.default is Parameter.empty
-            default_factory = (
-                self._composite_factory(arg_type)
-                if isclass(arg_type)
-                and issubclass(arg_type, BaseModel)
+            if (
+                isclass(arg_type)
+                and (issubclass(arg_type, BaseModel) or is_dataclass(arg_type))
                 and isinstance(para.default, str)
```
tpoliaw: Do we have any ophyd(-async) devices which are also BaseModels or dataclasses? If not we could remove the `str` requirement and not have to use `inject("")` for composite devices where it doesn't really make sense.

rtuck99: The issue I see with this is that it would also mean that Pydantic models which are not device composites (i.e. parameter models) would also automatically become defaulted, which is probably not what we want.

There is a PR https://github.com/DiamondLightSource/dodal/pull/1649 which at least makes the string argument to inject optional

Issue #732, opened 2024-11-28
Add an option to connect to devices with `mock=True`


d-perl:
When doing end-to-end testing, e.g. calling plans from my new GUI, I'd like to be able to run an instance of BlueAPI with mock devices to run plans against.

I've looked through the code a bit and can't find anything like it, but if it already exists then this issue is to add it to the docs.

callumforrester:
This is a good idea and important. @fajinyuan has expressed a desire for a similar feature before. 

## Possible Solution

Add a `mock` flag to the env config, if it is `True` pass `mock=True` to `make_all_devices` in `BlueskyContext`

```yaml
env:
    mock: true
```

## Acceptance Criteria
- There is a `mock` flag in `config.env`
- It defaults to `False`
- If it is set to `True`, all (ophyd-async) devices are connected in mock mode

PR #1245, opened 2025-10-20
feat: Add option to connect as mock devices

rtuck99:


rtuck99:
> Do you have a use case that requires this? To my mind this should be tracked in the dodal module, as the device_factory allows setting whether to connect sim or not, rather than having multiple potentially overlapping places to define this.

The use case is the same as the parent issue
* #732 

When hyperion moves to blueapi, we will be launching as standard blueapi plan + device modules, so without this option there will be no way to test it outside of connecting to real devices, if we want to be able to do integration testing. 

I expect that this will come in handy when I start writing the hyperion-supervisor process, as then I will want to be able to verify that I can launch plans, monitor execution and do things like start and stop `hyperion-blueapi` plan execution when the baton is taken away, etc. etc.

tpoliaw:
> m not sure what the plan is currently for getting the device manager into dodal - on its own the PR https://github.com/DiamondLightSource/dodal/pull/1549 doesn't provide anything without accompanying changes in dodal device modules and blueapi, and it seems the related discussion https://github.com/DiamondLightSource/dodal/discussions/1574 is stalled at the moment.

The change in dodal is mostly complete. For now, both approaches are supported and beamlines can try using the new version when they get a chance. I have made an example change for i04 in [#1684](https://redirect.github.com/DiamondLightSource/dodal/pull/1684).

For the blueapi side, the aim is to support the old loading for a short transition period then move everyone to the new system. As part of that I'll make sure mock is support is included.

Do you mind if I merge this PR into that change?

PR #1246, opened 2025-10-20
fix: Prevent spans being swallowed by infinite startup span

### LINKED ISSUE #1243: Trace for Health monitoring is captured inside startup trace, meaning the former is not exposed and the latter is infinitely long
@keithralphs  to add details

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

DiamondJoseph:
Fixes #1243 

Removes the spans for two methods that start either start subprocesses or loop infinitely, meaning the methods that they call can be displayed as unique traces rather than being swallowed into a single trace.

### DIFF
--- a/src/blueapi/service/runner.py
+++ b/src/blueapi/service/runner.py
@@ -109,7 +109,6 @@ class WorkerDispatcher:
                 error_message=_safe_exception_message(e),
             )
 
-    @start_as_current_span(TRACER, "function", "args", "kwargs")
     def run(
         self,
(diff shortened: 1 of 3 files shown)


DiamondJoseph:
Note: from deploying this to a test environment it did not previously work to stop the span being constantly appended to @tpoliaw

abbiemery (approved):

tpoliaw:
Any idea why not and/or how to set up a dev environment to test it?

PR #1247, opened 2025-10-20
refactor: Use pydantic event model for start and stop document handling

abbiemery:
This is a change on top of this https://github.com/DiamondLightSource/blueapi/pull/1221 pr to see if using the pydantic models was better. I don't know about their support in the larger bluesky ecosystem to make any comment on their use here, but was worth playing with.


abbiemery:
Closing as not needed.

PR #1249, opened 2025-10-21
fix: Prevent duplicate logging messages

### LINKED ISSUE #1191: Stop spamming so many logs
~With blueapi set up to log to graylog I get 2 messages every 10 seconds in the form `method: GET url: http://172.23.104.10:8000/healthz body: b''` set at INFO level. I suspect these are coming from GDA polling blueAPI. These clutter up logs and make it harder to see actual logs sent by my plan.~

~I would argue that these logs aren't helpful and the should be entirely removed but if we do want them then they should at least be debug, I can't see why support staff would care about them that much in regular operation.~

Above was fixed by https://github.com/DiamondLightSource/blueapi/pull/1105 but:

Similarly with event messages going back to GDA, I get many of the form `SENDING b'{"name":"event","doc":{"uid":"4d9801ee-ee7e-4d6a-8915-7b9e2c063ac3","time":1756993825.6583254,"data":{"oav_to_redis_forwarder-uuid":"fullscreen-22737388-58e33a69-3b55-4e4e-a479-d63d474246df"},"timestamps":{"oav_to_redis_forwarder-uuid":1756993825.6583047},"seq_num":82,"filled":{},"descriptor":"20b364f6-2527-482d-97b1-0bcc5cd6b39b"},"task_id":"79b34e15-ec39-4bfa-a3d4-29ebd3d8ca3d"}' to /topic/public.worker.event` I can see these being somewhat helpful but they should probably be debug, not info.

## Acceptance Criteria
- ~`Get healthz` messages are removed or downgraded in logs~
- Event messages are downgraded or deleted

DiamondJoseph:
Fixes #1191 which had a regression: disables the uvicorn access_log logging, which logs all incoming and requests at the INFO level- instead our injected `log_request_details` logs healthz probes at DEBUG and all other endpoints at INFO. Previously we were also logging all endpoints twice, once with our middleware and once with the access probe.

### DIFF
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -573,7 +573,9 @@ def start(config: ApplicationConfig):
     app.state.config = config
     assert config.api.url.host is not None, "API URL missing host"
     assert config.api.url.port is not None, "API URL missing port"
-    uvicorn.run(app, host=config.api.url.host, port=config.api.url.port)
+    uvicorn.run(
+        app, host=config.api.url.host, port=config.api.url.port, access_log=False
+    )
 
 
 async def add_api_version_header(


dan-fernandes (approved):
LGTM

PR #1251, opened 2025-10-22
fix: Make the path provider stack start docs

abbiemery:
Makes the `StartDocumentPathProvider` stack docs, using the latest for path determination.

### DIFF
--- a/src/blueapi/utils/path_provider.py
+++ b/src/blueapi/utils/path_provider.py
@@ -29,7 +29,7 @@ class StartDocumentPathProvider(PathProvider):
                 self._docs.pop()
             else:
                 raise BlueskyRunStructureError(
-                    "Close run called, but not for the inner most run."
+                    "Close run called, but not for the inner most run. "
                     "This is not supported. If you need to do this speak to core DAQ."
                 )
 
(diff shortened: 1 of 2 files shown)


src/blueapi/utils/path_provider.py
```diff
@@ -17,19 +17,21 @@ class StartDocumentPathProvider(PathProvider):
...
-            and self._doc is not None
-            and stop_document.get("run_start") == self._doc["uid"]
-        ):
-            self._doc = None
+        if name == "stop":
+            if stop_document.get("run_start") == self._docs[-1]["uid"]:
+                self._docs.pop()
+            else:
+                raise BlueskyRunStructureError(
+                    "Close run called, but not for the inner most run."
```
tpoliaw:
```suggestion
                    "Close run called, but not for the inner most run. "
```

tests/unit_tests/utils/test_path_provider.py
```diff
@@ -268,68 +271,110 @@ def stop_doc() -> dict:
...
+    stop_doc_1: RunStop,
+    start_doc_2: RunStart,
+):
+    pp = StartDocumentPathProvider()
+    pp.run_start(name="start", start_document=start_doc_1)
+    pp.run_start(name="start", start_document=start_doc_2)
+
+    with pytest.raises(
+        BlueskyRunStructureError,
+        match="Close run called, but not for the inner most run."
```
tpoliaw:
```suggestion
        match="Close run called, but not for the inner most run. "
```

tpoliaw (approved):

PR #1267, opened 2025-11-04
feat: Add support for device manager beamline modules

tpoliaw:


rtuck99 (approved):
Approved at least as far as our requirements go for Hyperion/mx-bluesky

PR #1284, opened 2025-11-28
feat: Add a UserClient that can be used as a scripting interface for blueapi

RJCD-Diamond:
The client that exists in blueapi is functional. This class extends the ability of the client, abstracts a lot of the logic and allows beamline staff or users to start building scripts used for an experiment.

### DIFF
--- a/tests/unit_tests/client/test_user_client.py
+++ b/tests/unit_tests/client/test_user_client.py
@@ -1,3 +1,5 @@
+import os
+from pathlib import Path
 from unittest.mock import Mock, patch
 
 import pytest
@@ -9,8 +11,11 @@ from blueapi.client.rest import BlueapiRestClient
 from blueapi.client.user_client import UserClient
 from blueapi.service.model import DeviceResponse, PlanResponse
 
-BLUEAPI_CONFIG_PATH = (
-    "/workspaces/blueapi/tests/unit_tests/valid_example_config/client.yaml"
+UNIT_TEST_DIRECTORY = Path(__file__).parent.parent
+
+
+BLUEAPI_CONFIG_PATH = os.path.join(
+    UNIT_TEST_DIRECTORY, "valid_example_config/client.yaml"
 )
 
 


RJCD-Diamond:
@EmsArnold

RJCD-Diamond:
@abbiemery @DominicOram @EmsArnold @olliesilvester @Relm-Arrowny @oliwenmandiamond 

I have a feeling this may be controversial, so I would like opinions on this. This has been extremely useful for saxs-bluesky and IMO it means we have something closer to a minimum viable Domain-Specific Language (DSL)

olliesilvester:
There's definitely a need for this sort of thing. How have scientists/users been using this so far, and is there feedback?

I've experimented with similar things using ipython and jupyter notebooks, but it's hard to make it not clunky

iain-hall:
I would find this useful as well. At some point I came up with some functions to wrap Blueapi calls to do something very similar, so having a common standard solution to this problem wouild be good. My current method for testing Blueapi with real hardware on a beamline is fire up GDA and use the Jython console!

olliesilvester:
I agree with a lot of Dom's concerns. For me though, I am now aware of at least 4 different implementations where someone in DAQ has written some sort of BlueAPI wrapper purely for the purpose of allowing scientists to easily run bluesky plans through a terminal. 

Given that loads of us are trying to do the same thing ATM, it would be good to have a simple but stable solution we can all use asap

Relm-Arrowny:
> To allow scientists to be able to do any of these things you then have the option of:
> 
> 1. Add more complexity to the `UserClient` and `BlueAPI`
> 2. Teach scientists how to actually write plans and get them to add them to the server

I don't believe Option 2 is the core issue. Instead, I personallty think BlueAPI does not deliver the same User Experience as GDA. While scientists writing and submitting plans to the server is fine, we need an experience that closer to our current user expectation.
I feel we are not adequately accounting for our current user expectations. Many users require a scripting-like experience, they running experiments interactively at the terminal, executing custom scripts for simple scan/movement stacking or complex operations (like bespoke controls loop, custom UBMatrix or on the spot analysis to guide the next steps). The current BlueAPI is so restrictive that it will likely be a source of friction for these users.

tpoliaw:
> I wasn't at the meeting, what was the requirement for serializing and sending plans to blueapi?

This would not be in scope for this PR and won't be happening any time soon. I don't think it's required for this work to progress.

@RJCD-Diamond are you happy for this and #1295 to be closed in favour of #1323 ? I think it should provide very similar functionality (if it doesn't, comment on that PR) without adding a new client into the library.

PR #1302, opened 2025-12-12
fix: Handle errors from event stream callbacks

tpoliaw:
If a callback raises an exception, it shouldn't prevent other callbacks
receiving the same event. Instead, catch any exceptions and re-raise
them after all callbacks have been called as a single ExceptionGroup.
This allows the task to be aborted if any of the callbacks fail but
still allows callbacks that require the "plan failed" events to run.

### DIFF
--- a/src/blueapi/core/event.py
+++ b/src/blueapi/core/event.py
@@ -78,12 +78,11 @@ class EventPublisher(EventStream[E, int]):
             correlation_id: An optional ID that may be used to correlate this
                 event with other events
         """
-
+        errs = []
         for callback in list(self._subscriptions.values()):
             try:
                 callback(event, correlation_id)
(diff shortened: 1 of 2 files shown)


abbiemery (dismissed):
Demo'd in person

abbiemery (approved):
I have played with this and it does make the error state server side nicer. Client side instead of a 500 error you now get this: 

Image

Not sure how much information this should contain or if this is better than the latter but I'm happy to deal with niceness discussions later.

My only complaint is, and this is a general complaint I guess rather than for this PR, it is confusing having two levels of callbacks. We should probably think about how we want to deal with people wanting callbacks, run_engine or otherwise. As this PR won't help the same issue if it occurs within the run_engine.

PR #1323, opened 2026-01-06
feat: Redesign BlueapiClient for use in scripts

tpoliaw:
Where previously the BlueapiClient was mostly a thin wrapper around the
BlueapiRestClient, it is now a more standalone class intended to be
usable from other scripts/applications as well as interactive sessions.

The client can now be created from just the path to a config file
without the user having to handle parsing.

Plans can now be run with less boilerplate with a more method-like
interface. Available devices can be queried and then used as if they
were local objects when running plans.

## Using the new client

```python
from blueapi.client import BlueapiClient
bc = BlueapiClient.from_config_file("/path/to/config.yaml")
bc.instrument_session = "cm12345-1"
```

This client can then be reused to run plans or query devices.

```python
for dev in bc.devices:
    print(dev)
    print("    " + ", ".join(p.name for p in dev.model.protocols))

det = bc.devices.det
stage = bc.devices.stage
```

Sub-devices can also be accessed, while checking that the sub-device
exists on the server:

```python
# assuming the default devices in adsim
stage_x = stage.x # returns new device
det_x = det.x # raises AttributeError
```

Available plans can be accessed from the server and treated
(approximately) like local methods. Calling a plan as a method blocks
while the plan is running

```python
count = bc.plans.count
help(count) # Display docstring from plan
assert str(count) == "count(detectors, num=None, delay=None, metadata=None)"

# Arguments can be passed as positional arguments
count([det, stage.x], 3) # det and stage from previous step

# ...or by keyword
count(detectors=[det, stage.x], delay=12)

# ... but the same argument can't use both
count([det], detectors=[stage.x]) # raises TypeError
```

Other client methods have been replaced by (cached) properties:

```python
print(bc.environment)
print(bc.oidc_config)

if bc.state == WorkerState.IDLE
    # previous methods are mostly still available
    bc.run_task(...)
else:
    print(bc.active_task)
```

While running in an interactive session, plans and devices are
auto-completable if the REPL supports it.

### DIFF
--- a/tests/unit_tests/client/test_client.py
+++ b/tests/unit_tests/client/test_client.py
@@ -166,8 +166,8 @@ def test_get_nonexistent_device(
 
 
 def test_get_child_device(mock_rest: Mock, client: BlueapiClient):
-    mock_rest.get_device.side_effect = (
-        lambda name: DeviceModel(name="foo.x", protocols=[ProtocolInfo(name="One")])
+    mock_rest.get_device.side_effect = lambda name: (
+        DeviceModel(name="foo.x", protocols=[ProtocolInfo(name="One")])
         if name == "foo.x"
         else None
     )


tpoliaw:
Thanks for looking at it. You're right, it needs a lot more docs but I wanted to wait to check the interface was vaguely right before writing it up. I'll add something to the PR for now.

> * It looks like I can run my plan by either doing `client.plans.my_plan(args)` or something like `client.run_task(TaskRequest(name="my_plan"...)`. We should only allow one and make the other private.  

The `client.plans.my_plan(client.devices.stage.x, optional=other)` interface is the one added here and is the one intended for use in scripts.

The `run_task`/`create_and_start_task` methods were there before and I didn't want to break backwards compatibility too much. Maybe if we're changing it anyway, breaking everything at once isn't the end of the world.

> * Should: What is the difference between `run_task` and `create_and_start_task` seems to me we only need one?

`run_task` blocks, `create_and_start` doesn't

> * Could: I think that for a user a `task` and a `plan` are pretty synonymous so I would change the wording so that the client just talks about plans, not tasks

Not a bad idea but all the `task` related things were here before this PR so should be a separate change.  Could be rolled into #1080 as that already removes the `TaskRequest` parameters

tpoliaw:
This requires #1315

tpoliaw:
I have added a `login` method to the client so users do not need to login beforehand. This will make the scripts that use it interactive if the user is not logged in. In future I think the login process could use refactoring (will open an issue) but for now I think this is a useful addition even if people want to avoid it and rely on users logging in first.

abbiemery (approved):
Having played with this and ran test scans locally I am happy enough for this to go in as is. Let people use it and make improvements from there.

Before merging just make the following docs changes:

- change to `from blueapi.client import BlueapiClient`
- document setting instrument session
- (maybe) routes for logging in if this is now doable in the script? (otherwise just leave it as login as normal)
- that this is run as a python script, so you will need a local environment with blueapi as a dependency, and not as a blueapi command. I can see this happening so one line here to state it might help.

Things we need to consider going forward:

- having a collection of useful callbacks (such as the example feedback one)
- How we want to address child devices and unifying that experience.
- Removing the backwards compatibility changes when we can which confuse the interface.
- stubgen for autocomplete


The stubs gen pr will probably want to follow this one reasonably quickly as not having autocomplete is pretty limiting. The changes we played with I have made as comments on that PR.

Issue #1236, opened 2025-10-15
blueapi should have option to not start if not all devices can connect


rtuck99:
Currently the behaviour is to still start even if some devices can't connect. However when Hyperion moves to blueapi, we will want blueapi to check whether devices can't be connected and refuse to start if this is the case. So we will need a configuration option or similar to allow this, while still allowing other consumers to have the existing behaviour

See

* #1235 

## Acceptance Criteria
- A configuration option exists to ensure beamline is fully connected at startup.

tpoliaw:
Adding an option to blueapi to abort if any devices fail to connect is possible although I have a feeling being in a subprocess might make it awkward to propagate back to the user.

There is currently no support in DeviceManager to distinguish between optional/required devices and I don't think it should block the current implementation being adopted. It does return a dict of device name to error though so we could potentially have a list of required devices in blueapi that must connect?

PR #1324, opened 2026-01-06
feat: Implement ensure_connected configuration option

rtuck99:
This change addresses the requirements of Hyperion as discussed in GH issue

* #1236 

DeviceManagerSource now accepts `ensure_connected` boolean option. When enabled any failure to build or connect a device will result in an exception. The default value for `ensure_connected` is False, which retains existing behaviour.

Launching `blueapi` in the event that devices cannot be connected will result in blueapi running, however the environment will not be initialised. Probing the environment via the REST API using `GET /environment` will result in a response similar to the following:

```
{
  "environment_id": "0631a4de-2f36-487e-9b3c-cf8edbcc2482",
  "initialized": false,
  "error_message": "ExceptionGroup: Errors occurred while connecting the following devices: fastcs_eiger, pin_tip_detection, panda (3 sub-exceptions)"
}
```


abbiemery (commented):
I think this is a good idea, but there are some things I think I need clarification on and other which are potentially worth discussing.

For some clarification:  So with this change (and =true)  blueapi still runs, there is an error code on the `/environment` endpoint, and the environment is now in unusable state.  The initial issue said it shouldn't start at all. Is this endpoint being queried as a check on whether to proceed instead?

I played with this change with `ensure_connected` as `true` with one of my devices down, and brought it up and down in various states.

With a device down:

If you have `ensure_connected` as `false` :
- you get one set of logs showing you all the pv's that didn't connect
- if from the client you do: `blueapi -c config.yaml controller devices` only the devices which built and connected are returned. 
- you can still query for plans
- if you try and run a plan with the device that didn't connect (and therefore isn't in the context) you get the error 
```
Error: Incorrect parameters supplied
    Invalid value 'det' for field detectors.0: Value error, Device det is not of type bluesky.protocols.Readable
```
and the plan doesn't run. Error message should be improved but hey ho.

If you have `ensure_connected` as `true` :

- On start in addition to the previous pv connection error logs, there are now duplicated `NotConnectedError`s  which is noisy. 
- if from the client you do: `blueapi -c config.yaml controller devices` you now get a 500 error and a reasonably obscure stack trace.
- You get the same if you query for plans.

So blueapi  is "up", but in a broken state which is not obviously different from the previous behaviour if you are just looking at the server logs.

Would the following be more helpful to you?:

Have the blueapi context instead add all devices that have built, even if they don't connect, but have some distinction for those that connected. 

For the `false` case as if you query the devices you get a more useful list of all _potentially_ available devices, as well as the ones which are available. We could also add a `connect <device>` command to the client so if you started blueapi with `motor1` disconnected, then connected it, you could ensure blueapi has caught up properly before trying to run a scan with it. 

In the `true` case the same occurs but if there are any connection errors the `/environment` endpoint says "error you can't scan with me, i don't have all my devices". Maybe if it receives a run scan command in that state it returns a "no i cant until you connect my devices". If there is a "connect" command as above it could also potentially avoid the need to reload the whole environment to check things have connected.

There are undoubtedly requirements and ramifications I have not considered here so some weigh in would be appreciated: @tpoliaw @rtuck99

PR #1335, opened 2026-01-14
fix: Replace logging.info with LOGGER.info

dan-fernandes:
Minor logging issue that messes with formatters


tpoliaw (dismissed):

dan-fernandes:
Good spot - the other instances of "direct logging" have been addressed

PR #1337, opened 2026-01-14
feat: Log connected devices when using deviceManager

tpoliaw:
The report_successful_devices function should be moved when the dodal
module support is removed but it can stay where it is for now.


keithralphs (commented):
Looks good, just needs the conflict resolving

keithralphs (approved):
All good now, please merge

PR #1338, opened 2026-01-15
Generate cache type stubs to help edit user scripts

tpoliaw:
When editing scripts that make use of the dynamic plan/device access in
the new client, there is no type checking available by default (as the
available plans are unknown). This adds a subcommand to generate a
type-stubs package for the currently running server to help write scripts.

    blueapi generate-stubs /tmp/blueapi-stubs
    pip install --editable /tmp/blueapi-stubs

type checkers and LSPs will then be able to check scripts for correct plan
names/arguments/etc as well as providing completions while editing.

### DIFF
--- a/tests/unit_tests/client/test_client.py
+++ b/tests/unit_tests/client/test_client.py
@@ -416,6 +416,15 @@ def test_run_task_fails_on_failing_event(
     on_event.assert_called_with(FAILED_EVENT)
 
 
+@patch("blueapi.client.client.BlueapiClient.run_task")
+def test_run_plan(run_task, client, mock_rest):
+    client.instrument_session = "cm12345-2"
+    client.run_plan("foo", {"foo": "bar"})
+    run_task.assert_called_once_with(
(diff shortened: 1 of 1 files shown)


src/blueapi/stubs/templates/cache_template.pyi
```diff
@@ -0,0 +1,57 @@
...
+    model: DeviceModel
+    _cache: DeviceCache
+    def __new__(cls, name: str, cache: DeviceCache, model: DeviceModel): ...
+    def __getattr__(self, name) -> DeviceRef: ...
+    def __repr__(self) -> str: ...
+
+class DeviceCache:
+    def __init__(self, rest: BlueapiRestClient) -> None: ...
+    def __getitem__(self, name: str) -> DeviceRef: ...
+    def __getattr__(self, name: str) -> DeviceRef: ...
```
tpoliaw: Think this should be removed to ensure missing devices are picked up

PR #1357, opened 2026-01-27
feat: Include plan outcome in WorkerEvents

tpoliaw:
The `TaskStatus` included in each `WorkerEvent` now contains a `result`
field that will be populated when the plan is complete. This can be
either a `TaskResult` that contains a serialized representation of the
return value from a plan that succeeded, or a `TaskError` containing the
Exception type and message raised by a plan that failed.

The existing `task_complete` and `task_failed` flags have been left for
backwards compatibility but are now redundant and may be removed in a
future version.

Error reporting in WorkerEvents is left unchanged for now but may also
change in future as plan errors are included in the worker error list
making it unclear where an error occurred.

The new outcome states in the returned `TaskStatus` allows better
feedback in the CLI and errors/return values are included in the output.

The BlueapiClient run_task method now returns only the TaskStatus as it
now contains all relevant information about the plan. A plan failing
will now not raise an exception with exceptions being raised only in the
case of the communication errors or internal server errors. Users of the
client should check if plans were successful.

### DIFF
--- a/tests/unit_tests/client/test_client.py
+++ b/tests/unit_tests/client/test_client.py
@@ -816,6 +816,7 @@ def test_adding_removing_callback(client):
                 task_id="foo",
                 task_complete=False,
                 task_failed=False,
+                result=None,
             ),
         ),
         ProgressEvent(task_id="foo"),


src/blueapi/utils/base_model.py
```diff
@@ -1,5 +1,11 @@
-from pydantic import BaseModel, ConfigDict
+import logging
 
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+)
+
+logger = logging.getLogger(__name__)
```
dan-fernandes: This looks unused, and should probably be `LOGGER`.

DominicOram (approved):
Looks good to me, thanks!

PR #1366, opened 2026-01-30
chore: Remove dependency on scanspec

tpoliaw:


tpoliaw:
Where does the client use scanspec? Doesn't it use the JSON object directly?

PR #1371, opened 2026-02-03
feat: use issuer url instead of well known url

ZohebShaikh:


tpoliaw (changes requested):
I like this change - it makes the auth config much clearer, thanks.

That said, this is a breaking change that will need every beamline to update its config and for GDA to support the new API. We have a [list of breaking issues](https://github.com/DiamondLightSource/blueapi/milestone/19) that are waiting for a 2.0 release so it might be worth adding an issue there for this change and keeping this PR around until we decide to break everything at once.

src/blueapi/config.py
```diff
@@ -217,7 +215,9 @@ class OIDCConfig(BlueapiBaseModel):
 
     @cached_property
     def _config_from_oidc_url(self) -> dict[str, Any]:
-        response: requests.Response = requests.get(self.well_known_url)
+        response: requests.Response = requests.get(
+            self.issuer + "/.well-known/openid-configuration"
```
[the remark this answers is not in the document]
tpoliaw: Spec for this is [here](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) as a `MUST` so we can probably rely on it.

> OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer

ZohebShaikh:
@tpoliaw I have made it backwards compatible so in v2 we can remove the backwards compatibility.

ZohebShaikh:
@tpoliaw Can you have a look at this ?

src/blueapi/config.py
```diff
@@ -234,18 +238,47 @@ class ScratchConfig(BlueapiBaseModel):
...
     )
+    issuer: str | None = Field(description="URL of OIDC provider", default=None)
     client_id: str = Field(description="Client ID")
     client_audience: str = Field(description="Client Audience(s)", default="blueapi")
     logout_redirect_endpoint: str = Field(
         description="The oidc endpoint required to logout", default=""
     )
 
+    @model_validator(mode="after")
+    def check_well_know_urls(self) -> Self:
```
tpoliaw:
```suggestion
    def check_urls(self) -> Self:
```

src/blueapi/config.py
```diff
@@ -234,18 +238,47 @@ class ScratchConfig(BlueapiBaseModel):
...
+            )
+        return self
+
+    @cached_property
+    def _well_known_url(self) -> str:
+        if self.issuer:
+            if self.well_known_url:
+                LOGGER.warning(
+                    DeprecationWarning(
+                        "well_known_url and issuer both are set. "
```
tpoliaw:
```suggestion
                        "well_known_url and issuer are both set. "
```

tests/unit_tests/test_config.py
```diff
@@ -545,3 +546,19 @@ def test_config_schema_updated() -> None:
...
+
+
+def test_oidc_config_validation_error():
+    with pytest.raises(ValueError, match="Please provide 'OIDCConfig.issuer'"):
+        ApplicationConfig(
+            oidc=OIDCConfig(well_known_url=None, issuer=None, client_id="blueapi")
+        )
+
+
+def test_oidc_config_urls():
```
tpoliaw: This is really two tests

tpoliaw (approved):

tpoliaw (approved):

PR #1386, opened 2026-02-13
feat: Add user to metadata

dan-fernandes:
Add `user` field to run metadata, obtained via AuthN


src/blueapi/service/interface.py
```diff
@@ -147,11 +147,9 @@ def get_device(name: str) -> DeviceModel:
     return DeviceModel.from_device(device)
 
 
-def submit_task(task_request: TaskRequest) -> str:
+def submit_task(task_request: TaskRequest, metadata: dict[str, Any]) -> str:
```
ZohebShaikh:
```suggestion
def submit_task(task_request: TaskRequest, metadata: dict[str, Any] = {}) -> str:
```

ZohebShaikh (approved):
The user info of who has called any endpoint will be done in a separate PR for graylog and logs

PR #1395, opened 2026-02-16
feat: Add host CLI option as alternative to config

tpoliaw:


src/blueapi/client/rest.py
```diff
@@ -230,6 +230,9 @@ def get_oidc_config(self) -> OIDCConfig | None:
             # Server is not using authentication
             return None
 
+    def get_stomp_config(self) -> StompConfig | None:
+        return self._request_and_deserialize("/config/stomp", StompConfig)
```
tpoliaw: Need to handle servers without endpoint


```suggestion
        try:
            return self._request_and_deserialize("/config/stomp", StompConfig)
        except KeyError:
            return None
```

docs/reference/openapi.yaml
```diff
@@ -224,6 +241,28 @@ components:
...
       type: object
+    StompConfig:
+      additionalProperties: false
+      description: Config for connecting to stomp broker
+      properties:
+        auth:
+          $ref: '#/components/schemas/BasicAuthentication'
+          description: Auth information for communicating with STOMP broker, if required
+          title: Auth
+        enabled:
```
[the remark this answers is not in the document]
abbiemery: No. But i think this one is going to be a faffy one to resolve as we're changing defaults so warrants another PR.

src/blueapi/client/client.py
```diff
@@ -212,7 +208,7 @@ def __init__(
         events: EventBusClient | None = None,
     ):
         self._rest = rest
-        self._events = events
+        self._event_bus_client = events
```
abbiemery: nit: consistency?
```
    _rest_client: BlueapiRestClient
    _event_bus_client: EventBusClient | None
    _instrument_session: str | None = None
    _callbacks: dict[int, OnAnyEvent]
    _callback_id: itertools.count

    def __init__(
        self,
        rest: BlueapiRestClient,
        event_bus: EventBusClient | None = None,
    ):
        self._rest_client = rest
        self._event_bus_client = event_bus
```

abbiemery (commented):
When trying to run scans against tiled 0.2.4 (current state of main) with this I am getting an error from the tiled writer: 
```
TypeError: DataSource.__init__() got an unexpected keyword argument 'properties'
```
This may be just down to it needing rebasing but does raise the question of if we are clear somewhere about the versions of tiled we support.

Works against 0.2.3. Can login/out, get devices and run plans etc. Happy to approve pending a credential discussion.

abbiemery:
@tpoliaw So as it stands you can look at devices etc with the host option, but can't run plans?

tpoliaw:
You can run plans in the background but nothing that requires stomp

src/blueapi/cli/cli.py
```diff
@@ -81,6 +81,7 @@ def is_str_dict(val: Any) -> TypeGuard[TaskParameters]:
     invoke_without_command=True, context_settings={"auto_envvar_prefix": "BLUEAPI"}
 )
 @click.version_option(version=__version__, prog_name="blueapi")
+@click.option("-H", "--host", type=str)
```
abbiemery: Can this be lowercase to match the other options?

Please can you put a description in here `help` for what it does. Also please specify if you want to run plans with it you must use the `-bg` flag and you won't get in terminal feedback.

abbiemery (changes requested):
If you are being pedantic and not including a scheme with your host: `uv run blueapi -H localhost controller plans`, it fails loudly demanding a full url. 

Image

If we are always using a given scheme, could you add this handling? Otherwise perhaps put an example url in the `help`. Or rename it url.

src/blueapi/cli/cli.py
tpoliaw: ~~It can't be lowercase without clashing with help.~~ Can add a comment though

Apparently it can - `click != clap`

abbiemery: What happens to help then?

tpoliaw: There is `--help` for help, `-h` currently gives `no such option: -h`

tpoliaw:
Have added docs to host command and made `-h` lower case. The bg plan docs have gone in a new PR as they're not really related to host: #1515

abbiemery (approved):

PR #1413, opened 2026-02-25
feat: Run blocking plans via websockets

### LINKED ISSUE #1415: Add websocket end point to run plan
See comment in https://github.com/DiamondLightSource/blueapi/pull/1413 to list details

tpoliaw:
As an intermediate step before (or instead of?) redesigning the server sub-process handling, add a websocket endpoint to run the same submit-listen-start approach that the client runs remotely. This is an experimental feature subject to change in minor releases in future. Everything is opt-in and existing behaviour should not change.

For the CLI, the websocket connection is enabled using the `--ws` flag to the run subcommand. A stomp configuration will not be required in this case.

For the `BlueapiClient`, users should use the `run_blocking` method instead of the `run_task`. The signatures and behaviour are the same (other than the ignored `timeout` parameter of `run_task`).

On the server there is a new endpoint on the new `/api/v2/` tree. This re-uses the existing submit and start methods on the task runner but adds a new subscribe method to pipe events from the worker process to the websocket.

### DIFF
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -383,26 +383,29 @@ def run_plan(
 
 
 @controller.command(name="ws")
+@click.pass_obj
 @click.argument("name", type=str)
 @click.argument("parameters", type=ParametersType(), default={}, required=False)
+@click.option(
+    "-i",
+    "--instrument-session",
+    type=str,
+    help=textwrap.dedent("""
+        Instrument session associated with running the plan,
+        used to tell blueapi where to store any data and as a security check:
+        the session must be valid and active and you must be a member of it."""),
+    required=True,
+)
 def run_blocking(
-    name: str,
-    parameters: TaskParameters,
+    obj: dict, name: str, parameters: TaskParameters, instrument_session: str
 ):
-    instrument_session = "cm33-3"
-
-    from websockets.sync.client import connect
-
+    client = cast(BlueapiClient, obj["client"])
     task_req = TaskRequest(
         name=name,
         params=parameters,
         instrument_session=instrument_session,
     )
-
-    with connect("ws://localhost:8007/run_plan") as ws:
-        ws.send(task_req.model_dump_json())
-        for message in ws:
-            print(message)
+    client.run_blocking(task_req)
 
 
 @controller.command(name="state")
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -401,7 +401,7 @@ def run_blocking(
 
     with connect("ws://localhost:8007/run_plan") as ws:
         ws.send(task_req.model_dump_json())
-        while message := ws.recv():
+        for message in ws:
             print(message)
 
 
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -390,32 +390,6 @@ def run_plan(
         raise ClickException(f"task could not run: {ve}") from ve
 
 
-@controller.command(name="ws")
-@click.pass_obj
-@click.argument("name", type=str)
-@click.argument("parameters", type=ParametersType(), default={}, required=False)
-@click.option(
-    "-i",
-    "--instrument-session",
-    type=str,
-    help=textwrap.dedent("""
-        Instrument session associated with running the plan,
-        used to tell blueapi where to store any data and as a security check:
-        the session must be valid and active and you must be a member of it."""),
-    required=True,
-)
-def run_blocking(
-    obj: dict, name: str, parameters: TaskParameters, instrument_session: str
-):
-    client = cast(BlueapiClient, obj["client"])
-    task_req = TaskRequest(
-        name=name,
-        params=parameters,
-        instrument_session=instrument_session,
-    )
-    client.run_blocking(task_req)
-
-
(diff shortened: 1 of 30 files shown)


tpoliaw:
Rough next steps in no particular order
 - [x] Move websocket client method into BlueapiRestClient
 - [x] Add wrapper method in BlueapiClient?
 - [x] Add unstable/preview --ws flag to cli run method
 - [x] Deal with auth. Browser websocket api can't set headers
 - [x] Users + user-agents
 - [x] Include data/progress events
 - [x] Add service/meta events for error handling etc
 - [x] Handle client aborting
 - [x] Check interaction with server being busy
 - [x] Wrap pipe/unpipe into something that doesn't need finally handling
 - [ ] See what happens when listeners fall over/slow down - do messages back up
 - [x] Add middleware for debug logging

src/blueapi/service/runner.py
```diff
@@ -173,15 +223,7 @@ def import_and_run_function(
     func: Callable[..., T] = _validate_function(
         mod.__dict__.get(function_name, None), function_name
     )
-    value = func(*args, **kwargs)
-    return _valid_return(value, expected_type)
-
-
-def _valid_return(value: Any, expected_type: type[T] | None = None) -> T:
-    if expected_type is None:
-        return value
```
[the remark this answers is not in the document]
tpoliaw: Have put it back for now. We can look into whether it's needed for anything some other time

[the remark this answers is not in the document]
tpoliaw: This is still causing problems - will take it out again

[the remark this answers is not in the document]
tpoliaw: Or add the return type annotation to `unpipe_events` which was the cause of the new issues

src/blueapi/service/protocol.py
```diff
@@ -0,0 +1,103 @@
...
+    kind: Literal["pause"] = "pause"
+
+
+class Resume(BaseModel):
+    kind: Literal["resume"] = "resume"
+
+
+class Abort(BaseModel):
+    kind: Literal["abort"] = "abort"
+    reason: str | None = None
```
tpoliaw, line 53: These aren't used at the moment. Cancelling from the client side will abort the plan on the server.

src/blueapi/cli/cli.py
```diff
@@ -321,6 +321,7 @@ def on_event(
 @controller.command(name="run")
 @click.argument("name", type=str)
 @click.argument("parameters", type=ParametersType(), default={}, required=False)
+@click.option("--ws", type=bool, is_flag=True, default=False)
```
abbiemery: Maybe add a help description

src/blueapi/service/main.py
```diff
@@ -604,6 +627,81 @@ def logout(runner: Annotated[WorkerDispatcher, Depends(_runner)]) -> Response:
     )
 
 
+@secure_router_v2.websocket("/run_plan")
+async def run_plan(
```
tpoliaw, line 636: I think this method should be refactored into its own class but for now I'd like to get something usable in before improving it in further PRs

src/blueapi/service/interface.py
```diff
@@ -281,3 +285,37 @@ def get_python_env(
...
+    def handler(
+        worker_event: WorkerEvent | DataEvent | ProgressEvent,
+        _cor_id: str | None,
+    ) -> None:
+
+        try:
+            tx.send(worker_event)
+        except BrokenPipeError:
+            LOGGER.warning("Sending event to broken pipe")
+            pass
```
fajinyuan, line 309: does 'pass' required here?

tpoliaw:
Should probably have a way for plans run via the plan cache to use websockets. They currently default to using the existing message bus approach. Something like

```python
bc = BlueapiClient.from_config_file(...)
bc.use_ws(True)

bc.plans.count(detectors=[bc.devices.det])
```

It's not ideal though so open to suggestions. A new client only config field?

src/blueapi/service/runner.py
```diff
@@ -145,11 +150,57 @@ def run(
...
+class EventPipe:
+    runner: WorkerDispatcher
+    handles: list[tuple[SubHandles, Connection]]
+
+    def __init__(self, runner: WorkerDispatcher):
+        self.runner = runner
+        self.handles = []
+
+    def __enter__(self) -> EventStream:
+        tx, rx = Pipe()
```
shree-iyengar-dls, line 191: Nit: maybe could you make the variable names rx, tx and hnd full form? Just for readability

src/blueapi/service/protocol.py
```diff
@@ -0,0 +1,103 @@
+"""
+The application level sub-protocol used to communicate between the server and
+client when running plans via websockets
+"""
+
+# Client to server
+# * Submit task
+# * Pause
+# * Resume
+# * Abort
```
fajinyuan, line 10: one more use case - stop or abort after current point, ensure data collected so far are saved. like finish earlier!

fajinyuan (commented):
please see my comments.

abbiemery (approved):

abbiemery:
Approved. It works locally for me as expected, but given the size of the PR I think we just get it into a release and let people play with it.

PR #1427, opened 2026-03-02
test: Add test timeout to max of 60 seconds

ZohebShaikh:
This will be helpful in catching unit tests taking more than 3 seconds
For helm test they should ideally be not called Unit test because we generate the helm chart in a subprocess 
And for the system test the max timeout is 60 seconds because its taking to actual stuff


ZohebShaikh:
@DiamondLightSource/developers-daq-core 
Can this be reviewed ? 
I think this will help keep the unit-tests good and make it obvious why a unit-test is taking longer ( not having correct mocks like in #1425)

tpoliaw (approved):

PR #1439, opened 2026-03-10
fix: improve default error handling #1379 #1409

### LINKED ISSUE #1379: Remove stacktrace from error message when not logged into CLI
Without being logged in, running `blueapi controller plans` prints a stacktrace ending in 

    blueapi.client.rest.BlueskyRemoteControlError: (401, '<Response [401]>')

instead of an error message.

This was broken in #935 where the exception had the error code added and the wrapper no longer matches the string correctly.

### LINKED ISSUE #1409: Improve default error handling
The default exception handler in the rest client currently only handles `404` and bundles everything else in together

```python
    if code < 400:
        return None
    elif code == 404:
        return KeyError(str(response.json()))
    else:
        return BlueskyRemoteControlError(code, str(response))
```

It would be good to expand this to handle `401`/`403` as well as making `404` not a `KeyError` which only makes sense for certain requests.

Alexj9837:
Fixes https://github.com/DiamondLightSource/blueapi/issues/1379
Fixes https://github.com/DiamondLightSource/blueapi/issues/1409

Both issues related to error handling in the REST client.

1379 - Catches 'UnauthorisedAccessError` explicitly instead of string-matching
            on `BlueskyRemoteControlError`.

1409 - Expands the exception hierarchy so each
            HTTP status code raises a distinct typed exception rather than bundling
            everything into `BlueskyRemoteControlError`.

## Tests
- Update tests to match new exception signatures
- Add missing coverage for 422 non-list detail response
- Add missing coverage for streaming error in `run` command

### DIFF
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -217,13 +217,12 @@ def check_connection(func: Callable[P, T]) -> Callable[P, T]:
             raise ClickException(
                 "Failed to establish connection to blueapi server."
             ) from ce
+        except UnauthorisedAccessError as e:
+            raise ClickException(
+                "Access denied. Please check your login status and try again."
+            ) from e
         except BlueskyRemoteControlError as e:
-            if str(e) == "<Response [401]>":
-                raise ClickException(
-                    "Access denied. Please check your login status and try again."
-                ) from e
-            else:
-                raise e
+            raise e
 
     return wrapper
 
@@ -356,8 +355,10 @@ def run_plan(
         raise ClickException("Unauthorised request") from ua
     except InvalidParametersError as ip:
         raise ClickException(ip.message()) from ip
-    except (BlueskyRemoteControlError, BlueskyStreamingError) as e:
-        raise ClickException(f"server error with this message: {e}") from e
+    except BlueskyStreamingError as se:
+        raise ClickException(f"streaming error: {se}") from se
+    except BlueskyRemoteControlError as e:
+        raise ClickException(f"remote control error: {e.args[1]}") from e
     except ValueError as ve:
         raise ClickException(f"task could not run: {ve}") from ve
 
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -356,9 +356,9 @@ def run_plan(
     except InvalidParametersError as ip:
         raise ClickException(ip.message()) from ip
     except BlueskyStreamingError as se:
-        raise ClickException(f"streaming error: {se}") from se
+        raise ClickException(f"Streaming error: {se}") from se
     except BlueskyRemoteControlError as e:
-        raise ClickException(f"remote control error: {e.args[0]}") from e
+        raise ClickException(f"Remote control error: {e.args[0]}") from e
     except ValueError as ve:
         raise ClickException(f"task could not run: {ve}") from ve
 
--- a/src/blueapi/cli/cli.py
+++ b/src/blueapi/cli/cli.py
@@ -358,7 +358,7 @@ def run_plan(
     except BlueskyStreamingError as se:
         raise ClickException(f"streaming error: {se}") from se
     except BlueskyRemoteControlError as e:
-        raise ClickException(f"remote control error: {e.args[1]}") from e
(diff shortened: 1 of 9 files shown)


ZohebShaikh (commented):
Looks good only few things can be tidied up

src/blueapi/client/rest.py
```diff
@@ -33,17 +33,25 @@
...
+        super().__init__(code, message)
+
+
+class UnauthorisedAccessError(BlueskyRequestError):
     pass
 
 
-class BlueskyRemoteControlError(Exception):
+class BlueskyRemoteControlError(BlueskyRequestError):
     pass
```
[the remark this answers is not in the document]
Alexj9837: changed

tests/unit_tests/cli/test_cli.py
```diff
@@ -680,7 +678,7 @@ def test_env_reload_server_side_error(runner: CliRunner):
     assert isinstance(result.exception, BlueskyRemoteControlError), (
         "Expected a BlueskyRemoteError from cli runner"
     )
-    assert result.exception.args[0] == "Failed to tear down the environment"
+    assert result.exception.args[1] == "Failed to tear down the environment"
```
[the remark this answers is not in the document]
Alexj9837: it used to check BlueskyRemoteControlError, but since we reverted that to an Exception, this is no longer an issue and has be changed.

src/blueapi/client/rest.py
```diff
@@ -33,17 +33,25 @@
...
 
 
-class BlueskyRequestError(Exception):
-    def __init__(self, code: int, message: str) -> None:
-        super().__init__(message, code)
+class NotFoundError(BlueskyRequestError):
+    pass
```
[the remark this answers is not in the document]
Alexj9837: yeah that correct, as keyerror was mainly for missing key in a dict, where notfounderror is better for any 404's

src/blueapi/cli/cli.py
```diff
@@ -356,8 +355,10 @@ def on_event(event: AnyEvent) -> None:
         raise ClickException("Unauthorised request") from ua
```
tpoliaw: I think this will prevent your new `check_connection` changes having an effect when running plans - you'll still get the `Unauthorised request` message

src/blueapi/cli/cli.py
```diff
@@ -356,8 +355,10 @@ def on_event(event: AnyEvent) -> None:
...
     except InvalidParametersError as ip:
         raise ClickException(ip.message()) from ip
-    except (BlueskyRemoteControlError, BlueskyStreamingError) as e:
-        raise ClickException(f"server error with this message: {e}") from e
+    except BlueskyStreamingError as se:
+        raise ClickException(f"streaming error: {se}") from se
+    except BlueskyRemoteControlError as e:
+        raise ClickException(f"remote control error: {e.args[0]}") from e
```
tpoliaw:
```suggestion
    except BlueskyStreamingError as se:
        raise ClickException(f"Streaming error: {se}") from se
    except BlueskyRemoteControlError as e:
        raise ClickException(f"Remote control error: {e.args[0]}") from e
```

tests/system_tests/test_blueapi_system.py
```diff
@@ -376,10 +378,10 @@ def test_get_worker_state(client: BlueapiClient):
 def test_set_state_transition_error(client: BlueapiClient):
     with pytest.raises(BlueskyRemoteControlError) as exception:
         client.resume()
-    assert "<Response [400]>" in str(exception)
+    assert exception.value.args[0]
```
tpoliaw: What are these asserts testing for?

src/blueapi/cli/cli.py
Alexj9837: thanks, will make this change.

tests/system_tests/test_blueapi_system.py
Alexj9837: It was a leftover from when BlueskyRemoteControlError had a fixed response that broke with the new error messsages. It only checks the body is non-empty. I can remove this or I could add a description to the 400 responses in set_state, like  "Cannot transition from {current_state} to {new_state}") and check that instead? might be better for debugging

PR #1443, opened 2026-03-18
feat: check of server and client version in response header

### LINKED ISSUE #1074: Include blueapi version in metadata
Currently blueapi version is not included in the metadata this can be helpful to give warnings when blueapi-cli is talking to a different version of blueapi server. Currently the entire application in package as one so the blueapi-cli and blueapi server are only guaranteed to be compatible with the same version.

## Acceptance Criteria
- API version included in the metadata
- Warning added when blueapi-cli is talking to a different version of blueapi-server

shree-iyengar-dls:
Fixes #1074

Check of version headers in the response has been added to the client side.


ZohebShaikh (commented):
Looks good 👍🏼

Issue #1121, opened 2025-07-11
Device types in JSON schemas are non-standard


callumforrester:
See the following example JSON snippet from the `/plans` endpoint:

```json
{
  "name": "count",
  "description": "Reads from a number of devices.\n    Wraps bluesky.plans.count(det, num, delay, md=metadata) exposing only serializable\n    parameters and metadata.",
  "parameter_schema": {
    "additionalProperties": false,
    "properties": {
      "detectors": {
        "items": {
          "enum": [
            "sample_stage",
            "sample_det",
            "oav",
            "synchrotron",
            "panda"
          ],
          "type": "bluesky.protocols.Readable"
        },
        "title": "Detectors",
        "type": "array",
        "uniqueItems": true
      },
...
```

The detector type is listed as `"bluesky.protocols.Readable"`, JSONForms does not seem to be compatible with arbitrary types in this way. See the following error:

```
Uncaught Error: schema is invalid: data/properties/detectors/items/type must be equal to one of the allowed values, data/properties/detectors/items/type must be array, data/properties/detectors/items/type must match a schema in anyOf, data/properties/detectors/items must be array, data/properties/detectors/items must match a schema in anyOf
```

When I change the type to `"string"` I get a form rendered. 

@tpoliaw suggested moving the bluesky type to a non-standard field in https://github.com/DiamondLightSource/blueapi/pull/522, which may be a good idea given that the current state of affairs requires special casing in any clients.

## Acceptance Criteria
- Clients reading plan schemas do not need insider knowledge to determine device parameter types

PR #1469, opened 2026-04-01
feat: Improve BlueapiClient to add plan parameter type hints

### LINKED ISSUE #1505: BlueapiClient doesn't report the correct default values or display types for plans
When using BlueapiClient, it doesn't report the correct default values back to the user for a plan, it seems to always report None.

An example is show below:

```Python
>>> plans.count
count(detectors, num=None, delay=None, metadata=None)
```
It is showing that num default is None and delay is None when in fact this is actually 1 and zero as shown below. 

```Python
@attach_data_session_metadata_decorator()
@validate_call(config={"arbitrary_types_allowed": True})
def count(
    detectors: Annotated[
        set[Readable],
        Field(
            description="Set of readable devices, will take a reading at each point",
            min_length=1,
        ),
    ],
    num: Annotated[int, Field(description="Number of frames to collect", ge=1)] = 1,
    delay: Annotated[
        NonNegativeFloat | Sequence[NonNegativeFloat],
        Field(
            description="Delay between readings: if tuple, len(delay) == num - 1 and \
            the delays are between each point, if value or None is the delay for every \
            gap",
            json_schema_extra={"units": "s"},
        ),
    ] = 0.0,
    metadata: dict[str, Any] | None = None,
) -> MsgGenerator: 
```
It also doesn't display any type checking of what the plan argument types are meant to be to the user. I've created a fix for this https://github.com/DiamondLightSource/blueapi/

## Acceptance Criteria
- The plan schema shows the correct default values for arguments.
- The plan scema shows the correct types it expects.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/blueapi/issues/1505

Improves the help text for a plan to add the type hints and correct default values.

Before change:
```Python
>>> bc.plans.grid_analyserscan
grid_analyserscan(analyser, sequence, detectors, args, snake_axes=None, md=None)

>>> bc.plans.count
count(detectors, num=None, delay=None, metadata=None)
```

With this change

```Python
>>> bc.plans.grid_analyserscan
grid_analyserscan(
    analyser: ElectronAnalyserDetector,
    sequence: AbstractBaseSequence,
    detectors: list[Readable],
    args: Any,
    snake_axes: list[Any] | bool | None = None,
    md: dict | None = None
)

>>> bc.plans.count
count(
    detectors: list[Readable],
    num: int = 1,
    delay: float = 0,
    metadata: dict | None = None
)
```


src/blueapi/client/client.py
```diff
@@ -154,7 +179,7 @@ def help_text(self) -> str:
         return self.model.description or f"Plan {self!r}"
 
     @property
-    def properties(self) -> set[str]:
+    def properties(self) -> dict[str, Any]:
```
tpoliaw: This still returns a set, no? To be strictly correct, it should probably be `dict_keys` or `collections.abc.Set`.

Alternatively make it actually return the whole properties dict

src/blueapi/client/client.py
```diff
@@ -192,9 +217,35 @@ def _build_args(self, *args, **kwargs):
         return params
 
     def __repr__(self):
-        opts = [p for p in self.properties if p not in self.required]
-        params = ", ".join(chain(self.required, (f"{opt}=None" for opt in opts)))
-        return f"{self.name}({params})"
+        props = self.model.parameter_schema.get("properties", {})
+        required = set(self.required)
+
+        tab = "    "
```
tpoliaw: `tab = <not a tab>` is questionable


```suggestion
        indent = "    "
```

src/blueapi/core/context.py
tpoliaw, line 1: I think we want to be careful that we're not making an invalid schema. With these changes the schema contains several fields such as 
```
"group": {
    "title": "Group",
    "type": "string",
    "default": null
},
```
which in turn causes your repr to contain parameters such as `metadata: dict = None` which is not correctly typed.

src/blueapi/client/client.py
```diff
@@ -56,6 +55,32 @@
...
+        return " | ".join(_pretty_type(s) for s in schema["anyOf"])
+
+    json_type = schema.get("type")
+    type_map = {
+        "string": "str",
+        "integer": "int",
+        "boolean": "bool",
+        "number": "float",
+        "object": "dict",
+    }
```
Alexj9837, line 81: would adding something here to deal with null / none

src/blueapi/client/client.py
```diff
@@ -201,9 +226,35 @@ def _build_args(self, *args, **kwargs):
...
+                if "default" in info:
+                    default = repr(info["default"])
+                    arg = f"{arg} = {default}"
+                else:
+                    arg = f"{arg} | None = None"
+
+            args.append(arg)
+
+        single_line = f"{self.name}({', '.join(args)})"
+        max_length = 100
```
Alexj9837: not a huge fan of magic numbers, also will these change? could we set the vars to _REPR_MAX_LENGTH and _REPR_MAX_ARGS_INLINE

Alexj9837 (dismissed):
looks good, made two comments and maybe add a test for the added value

tpoliaw (changes requested):
I'm not sure about this. I agree that the current repr isn't great but trying to perfectly replicate the function signature is going to lead to edge cases where things don't behave, for instance this gives `detectors: list[Readable]` for an argument that expects a list of strings (or DeviceRefs). (Possibly fixed by #1121?)

I think this is coming back to a python function not being a perfect fit for an API endpoint. For `count`, we want `group` to be optional but we don't want the default to be in the signature (as it makes the schema invalid). I think it would be better if the repr for plans stopped trying to be a signature and became something like `Plan("count", (movable, value, group, wait))` and the type/default information was in the description (available via `help(bc.plans.count)`) as it is already. 

If the type stubs PR is merged it could go some of the way towards improving usability. For the 'optional but no default' cases we could have an `Unspecified` object or similar.

@EmsArnold, you've used the client in a repl a fair bit as well - any thoughts on usability vs correctness?

oliwenmandiamond:
> I'm not sure about this. I agree that the current repr isn't great but trying to perfectly replicate the function signature is going to lead to edge cases where things don't behave, for instance this gives `detectors: list[Readable]` for an argument that expects a list of strings (or DeviceRefs). (Possibly fixed by #1121?)
> 
> I think this is coming back to a python function not being a perfect fit for an API endpoint. For `count`, we want `group` to be optional but we don't want the default to be in the signature (as it makes the schema invalid). I think it would be better if the repr for plans stopped trying to be a signature and became something like `Plan("count", (movable, value, group, wait))` and the type/default information was in the description (available via `help(bc.plans.count)`) as it is already.
> 
> If the type stubs PR is merged it could go some of the way towards improving usability. For the 'optional but no default' cases we could have an `Unspecified` object or similar.

So this is something we should probably discuss in next drop in session. This is the balance we need to get right because you're right that the API endpoint doesn't seem to map to python typing very well and depends on what direction we want to head.

I think a compromise I could do with this change is not address the typing of devices, leave that to another change and instead handle primitive typing and default e.g 

```python
>>> bc.plans.count
count(
    detectors,
    num: int = 1,
    delay: float = 0,
    metadata: dict | None = None
)
```

I think the client plan typing should be derived from the blueapi schema rather than the original plan annotations. I think maybe we could it like this (in another change)

```Python
>>> bc.plans.count
count(
    detectors: DeviceRef[Readable],
    num: int = 1,
    delay: float = 0,
    metadata: dict | None = None
)
```
This way, it is not lying to the user using the client. They can clearly see if a device is Readable by checking protocols.

`bc.devices.my_device.model.protocols`

We could add a shortcut property for protocols
```Python
>>> bc.devices.my_device.protocols
["Readable", "Triggerable"]
```

Maybe we could also add some helpful methods to find specific protocols

```Python
>>> bc.print_readables()
["motor_x", "detector", ...]

>>> bc.print_triggerables()
["detector", ...]
```

I also think this should apply for specific server types as well e.g using a Motor and a StandardDetector plan
```Python
>>> bc.plans.my_plan
my_plan(
    motor: DeviceRef[Motor],
    detector: DeviceRef[StandardDetector],
)
```

Then we can also implement a new property for DeviceRef called server_type

```Python
>>> bc.devices.x.server_type
"Motor"
```

We could also override the __str__ of device ref to be a little more useful

```Python
>>> bc.devices.x
"Device(name=x, server_type=Motor)"
```

So to reply to your comments, I think what I've done is not quite right but can still be reused for primitive types and defaults. I think the idea of `Plan("count", (movable, value, group, wait))` is not the right solution either as this still offers no useful information to the user. 

A couple ideas to think about on how best to address this. Let me know what you think!

src/blueapi/client/client.py
oliwenmandiamond: Yeah you are right
```
>>> pl.count.properties
dict_keys(['detectors', 'num', 'delay', 'metadata'])
>>> type(pl.count.properties)
<class 'dict_keys'>
>>> type(pl.set_absolute.properties)
<class 'dict_keys'>
```
I will update this

src/blueapi/client/client.py
oliwenmandiamond: Not sure what you mean by this, it already handles null / None?

e.g

```
>>> pl.set_absolute
set_absolute(
    movable: Movable,
    value: Any,
    group: str | None = None,
    wait: bool = False
)
```
and server side

```python
Group = Annotated[str, "String identifier used by 'wait' or stubs that await"]

T = TypeVar("T")

def set_absolute(
    movable: Movable[T], value: T, group: Group | None = None, wait: bool = True
) -> MsgGenerator:
    """Set a device, wrapper for `bp.abs_set`.

    Args:
        movable (Movable[T]): The device to set.
        value (T): The new value.
        group (Group | None, optional): The message group to associate with the setting,
            for sequencing. Defaults to None.
        wait (bool, optional): The group should wait until all setting is complete (e.g.
            a motor has finished moving). Defaults to False.

    Returns:
        MsgGenerator: Plan.

    Yields:
        Iterator[MsgGenerator]: Bluesky messages.
    """
    yield from bps.abs_set(movable, value, group=group, wait=wait)
    return yield from bps.rd(movable) #type: ignore
```

oliwenmandiamond: discovered a bug with this though, seems to default wait = False when it is actually True. Will investigate that...

oliwenmandiamond: Okay all good, not sure how but my enviroment had dodal env of defaults to False for the plan itself but after updating my environment it is True as shown here https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/plan_stubs/wrapped.py#L17

and after updating env
```
>>> pl.set_absolute
set_absolute(
    movable: Movable,
    value: Any,
    group: str | None = None,
    wait: bool = True
)
```

src/blueapi/core/context.py
```diff
@@ -459,14 +459,16 @@ def _type_spec_for_function(
...
                 _type = self._convert_type(arg_type, no_default)
-            factory = None if no_default else default_factory
-            new_args[name] = (
-                _type,
-                FieldInfo(default_factory=factory),
-            )
+                if no_default:
+                    field_info = FieldInfo()
+                else:
+                    field_info = FieldInfo(default=para.default)
```
tpoliaw: This makes `null` appear in the schema where we are explicitly hiding it as an option (via `SkipJsonSchema[NoneType]` in `_convert_type`). If you want to expose the default values for the primitive cases, it would be better to keep the default_factory for the None case.

```python
                match para.default:
                    case Parameter.empty:
                        info = FieldInfo(default_factory=None)
                    case None:
                        info = FieldInfo(default_factory=lambda: None)
                    case _:
                        info = FieldInfo(default=para.default)
```

src/blueapi/client/client.py
```diff
@@ -56,6 +55,35 @@
...
+        return " | ".join(_pretty_type(s) for s in schema["anyOf"])
+
+    json_type = schema.get("type")
+    type_map = {
+        "string": "str",
+        "integer": "int",
+        "boolean": "bool",
+        "number": "float",
+        "object": "dict",
+    }
```
tpoliaw: Move this dict to a `_JSON_TYPE_MAP` (or similar) global/constant

src/blueapi/client/client.py
```diff
@@ -56,6 +55,35 @@
 
 log = logging.getLogger(__name__)
 
+_REPR_MAX_LENGTH = 100
+_REPR_MAX_ARGS_INLINE = 3
+
+
+def _pretty_type(schema: dict[str, Any]) -> str:
```
tpoliaw: This could be moved to the bottom of the module.

src/blueapi/client/client.py
```diff
@@ -56,6 +55,35 @@
...
+    if "$ref" in schema:
+        return schema["$ref"].split("/")[-1]
```
tpoliaw:
```suggestion
    if ref := schema.get("$ref"):
        return ref.split("/")[-1]
```

src/blueapi/client/client.py
```diff
@@ -56,6 +55,35 @@
...
+
+    if schema.get("type") == "array":
+        item_schema = schema.get("items", {})
+        inner = _pretty_type(item_schema)
+        return f"list[{inner}]"
```
tpoliaw: I'm not sure if `items` will ever be missing or empty but it'd be good to avoid the invalid `list[]` if it is


```suggestion
    if schema.get("type") == "array":
        items = schema.get("items")
        return f"list[{_pretty_type(items)}]" if items else "list"
```

src/blueapi/client/client.py
```diff
@@ -56,6 +55,35 @@
...
+
+    if "anyOf" in schema:
+        return " | ".join(_pretty_type(s) for s in schema["anyOf"])
```
tpoliaw:
```suggestion
    if anyof := schema.get("anyOf"):
        return " | ".join(_pretty_type(s) for s in anyof)
```

src/blueapi/client/client.py
tpoliaw:
```suggestion
        items = schema.get("items", {})
        return f"list[{_pretty_type(items)}]"
```

src/blueapi/client/client.py
```diff
@@ -201,10 +229,39 @@ def _build_args(self, *args, **kwargs):
...
         return params
 
-    def __repr__(self):
-        opts = [p for p in self.properties if p not in self.required]
-        params = ", ".join(chain(self.required, (f"{opt}=None" for opt in opts)))
-        return f"{self.name}({params})"
+    def __repr__(self) -> str:
+        def _format_arg(name: str, info: dict[str, Any], required: set[str]) -> str:
+            typ = _pretty_type(info)
+
```
tpoliaw, line 215: I think this could be simpler. `required` will be the same for every call so you could create it once before the function is defined and use it directly

```python
required = set(self.required)
def _format_arg(name: str, info: dict[str, Any]) -> str:
    ...
```

`is_required` and `has_default` are only used in one place so can be inlined or merged with the `default` use

```python
if name in required:
    return f"{name}: {typ}"
if default := info.get("default"):
    return f"{name}: {typ} = {default!r}"
return f"{name}: {typ} | None = None"
```

src/blueapi/client/client.py
```diff
@@ -201,10 +229,39 @@ def _build_args(self, *args, **kwargs):
...
+                    return f"{name}: {typ} | None = None"
+                return f"{name}: {typ} = {repr(default)}"
+
+            # optional with no default
+            return f"{name}: {typ} | None = None"
+
+        props = self.model.parameter_schema.get("properties", {})
+        args = [
+            _format_arg(name, info, set(self.required)) for name, info in props.items()
+        ]
```
tpoliaw: If `required` is defined above and `properties` becomes a `dict`, this becomes 
```suggestion
        args = [_format_arg(name, info) for name, info in self.properties.items()]
```

src/blueapi/client/client.py
```diff
@@ -164,7 +192,7 @@ def help_text(self) -> str:
         return self.model.description or f"Plan {self!r}"
 
     @property
-    def properties(self) -> set[str]:
+    def properties(self) -> KeysView[str]:
```
tpoliaw: I'm not sure what you can do with just the keys that you can't do with the whole dict so returning the full properties would be useful

tests/unit_tests/client/test_client.py
```diff
@@ -806,7 +806,72 @@ def test_plan_fallback_help_text(client):
...
+        ),
+        client,
+    )
+    assert (
+        plan.help_text == "Plan foo(\n"
+        "    one: Any,\n"
+        "    two: list[Any] | bool,\n"
+        "    three: Any = 3,\n"
+        "    four: Any | None = None\n"
+        ")"
```
tpoliaw:
```suggestion
        plan.help_text == dedent("""\
            Plan foo(
                one: Any,
                two: list[Any] | bool,
                three: Any = 3,
                four: Any | None = None
            )""")

```

src/blueapi/core/context.py
oliwenmandiamond: I tried this suggestion and it doesn't work, please see below:

```python
def test_optional_arg_generated_schema(
    empty_context: BlueskyContext,
):

    def demo_plan(foo: int | None = None) -> MsgGenerator:
        yield from ()
    ...
```
Error message
```
AssertionError: assert {'foo': {'title': 'Foo', 'type': 'integer'}} == {'foo': {'title': 'Foo', 'type': 'integer', 'default': None}}
  
  Differing items:
  {'foo': {'title': 'Foo', 'type': 'integer'}} != {'foo': {'title': 'Foo', 'type': 'integer', 'default': None}}
  
  Full diff:
    {
        'foo': {
           'title': 'Foo',
           'type': 'integer',
           'default': None,
        },
    }
```
The plan has a default of None but with this change the default gets dropped which is incorrect

src/blueapi/client/client.py
oliwenmandiamond: Done

src/blueapi/client/client.py
oliwenmandiamond: done

src/blueapi/client/client.py
oliwenmandiamond: Done

src/blueapi/client/client.py
oliwenmandiamond: Done

src/blueapi/client/client.py
oliwenmandiamond: Done

src/blueapi/core/context.py
oliwenmandiamond: done!

src/blueapi/client/client.py
```diff
@@ -201,10 +210,30 @@ def _build_args(self, *args, **kwargs):
...
+            default = info.get("default")
+
+            if name in required:
+                return f"{name}: {typ}"
+            if default := info.get("default"):
+                return f"{name}: {typ} = {default!r}"
+            return f"{name}: {typ} | None = None"
+
+        props: dict = self.model.parameter_schema.get("properties", {})
+        args = [_format_arg(name, info) for name, info in props.items()]
```
tpoliaw:
```suggestion
        args = [_format_arg(name, info) for name, info in self.properties.items()]
```

tpoliaw (approved):
Looks good now. Any other issues can be addressed in later PRs

PR #1473, opened 2026-04-07
refactor: Rewrite server middleware with websocket support

tpoliaw:
The builtin fastapi support for middleware only supports http/rest
requests. To enable the same middleware for websockets, a new
implementation of the starlette middleware is required and it makes
sense to use the same implementation for both protocols.

For rest requests, there should not be any change in behaviour from this
change.

### DIFF
--- a/src/blueapi/service/middleware.py
+++ b/src/blueapi/service/middleware.py
@@ -1,0 +1,58 @@
+import logging
+
+from opentelemetry.context import attach
+from opentelemetry.propagate import get_global_textmap
+from starlette.types import ASGIApp, Message, Receive, Scope, Send
+
+from blueapi import __version__
+from blueapi.config import ApplicationConfig
+
+OBS_LOGGER = logging.getLogger("blueapi.service.middleware.observability")
+
+CONTEXT_HEADER = ApplicationConfig.CONTEXT_HEADER.encode()
+VENDOR_CONTEXT_HEADER = ApplicationConfig.VENDOR_CONTEXT_HEADER.encode()
+
+API_VERSION = (b"x-api-version", ApplicationConfig.REST_API_VERSION.encode("utf-8"))
+VERSION = (b"x-blueapi-version", __version__.encode("utf-8"))
+
+
+class VersionHeaders:
+    def __init__(self, app: ASGIApp):
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope.get("type") not in ("websocket", "http"):
+            return await self.app(scope, receive, send)
+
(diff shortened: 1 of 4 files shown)


src/blueapi/service/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+        if scope["type"] not in ("http", "websocket"):
+            return await self.app(scope, receive, send)
+
+        ctx = None
+        v_ctx = None
+        for key, val in scope.get("headers", ()):
+            if key == CONTEXT_HEADER:
+                ctx = val.decode()
+            elif key == VENDOR_CONTEXT_HEADER:
+                v_ctx = val.decode()
```
NeilSmithDLS, line 50: I assume that `CONTEXT_HEADER` and `VENDOR_CONTEXT_HEADER` will only appear on `headers` once?

src/blueapi/service/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+
+        await self.app(scope, receive, local_send)
+
+
+class ObservabilityContextPropagator:
+    def __init__(self, app: ASGIApp):
+        self.app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
+        if scope["type"] not in ("http", "websocket"):
```
NeilSmithDLS: Line 24 reads:
    `if scope.get("type") not in ("websocket", "http"):`
whilst line 41 reads:
    `if scope["type"] not in ("http", "websocket"):`
why the difference in how `type` is retrieved from `scope`?

tpoliaw: No idea. The key should always be there but have changed it to `get` everywhere for consistency.

src/blueapi/service/middleware.py
tpoliaw: I hope so. And if not I think the 'last entry wins' approach is probably fine

src/blueapi/service/middleware.py
```diff
@@ -0,0 +1,58 @@
...
+        async def local_send(message: Message):
+            if message["type"] in ("websocket.accept", "http.response.start"):
+                message["headers"].append(VERSION)
+                message["headers"].append(API_VERSION)
+            await send(message)
+
+        return await self.app(scope, receive, local_send)
+
+
+class ObservabilityContextPropagator:
```
tpoliaw, line 36: Need to figure out how this relates to websockets. We might not want to be running the full method on every read and write.

tpoliaw: This only runs when the connection is first made

abbiemery (approved):

PR #1477, opened 2026-04-09
fix: improve error handling when request receives non-JSON response

NeilSmithDLS:
improve error handling when request receives non-JSON response.

Closes 1472

### DIFF
--- a/src/blueapi/client/rest.py
+++ b/src/blueapi/client/rest.py
@@ -106,13 +106,14 @@ class UnknownPlanError(Exception):
 
 
 def _exception(response: requests.Response) -> Exception | None:
-    try:
-        response.json()
-    except json.decoder.JSONDecodeError:
-        return BlueskyRemoteControlError(
-            "Response does not contain a valid JSON object"
-        )
     code = response.status_code
+    if response.content:
+        try:
+            response.json()
+        except json.decoder.JSONDecodeError:
+            return BlueskyRemoteControlError(
+                "Response does not contain a valid JSON object"
+            )
     if code < 400:
         return None
     elif code == 404:
--- a/src/blueapi/client/rest.py
+++ b/src/blueapi/client/rest.py
@@ -2,6 +2,7 @@ import logging
 from collections.abc import Callable, Mapping
 from typing import Any, Literal, TypeVar
 
+import json
 import requests
 from fastapi import status
 from observability_utils.tracing import (
@@ -107,8 +108,9 @@ class UnknownPlanError(Exception):
 def _exception(response: requests.Response) -> Exception | None:
     try:
         response.json()
-    except:
-        return BlueskyRemoteControlError("Response does not contain a valid JSON object")
+    except json.decoder.JSONDecodeError as je:
+        return BlueskyRemoteControlError(
+            "Response does not contain a valid JSON object")
     code = response.status_code
     if code < 400:
         return None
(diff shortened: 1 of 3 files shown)


src/blueapi/client/rest.py
```diff
@@ -106,6 +107,13 @@ class UnknownPlanError(Exception):
...
+    if response.content:
+        try:
+            response.json()
+        except json.decoder.JSONDecodeError:
+            return BlueskyRemoteControlError(
+                "Response does not contain a valid JSON object"
+            )
     if code < 400:
         return None
     elif code == 404:
```
Alexj9837, line 116: What would happen if you got a 500 code, that held html?

tpoliaw (changes requested):
I'm not sure the `_exception` method is the best place for this check to be, for instance running a plan uses the alternative `_create_task_exceptions` method and bypasses this check

```
$ # with config pointing to oauth-proxy
$ uv run blueapi -c path/to/config.yaml controller run --bg -i cm12345-3 sleep '{"time": 3}'
Error: task could not run: Expecting value: line 1 column 1 (char 0)
```

We're also parsing the json multiple times on the happy path when everything is working which isn't ideal.

Would it work to add a helper method to wrap wherever we're calling `.json()`?

src/blueapi/client/rest.py
```diff
@@ -138,6 +139,15 @@ def _create_task_exceptions(response: requests.Response) -> Exception | None:
         return BlueskyRequestError(code, response.text)
 
 
+def _response_json(response) -> Any:
+    try:
+        return response.json()
+    except json.decoder.JSONDecodeError:
+        return BlueskyRemoteControlError(
```
Alexj9837: I think this should be a raise? else you'll be returning an exception which might cause an issue when getting passed into `deserialized = TypeAdapter(target_type).validate_python(response.json())`

src/blueapi/client/rest.py
```diff
@@ -138,6 +139,15 @@ def _create_task_exceptions(response: requests.Response) -> Exception | None:
         return BlueskyRequestError(code, response.text)
 
 
+def _response_json(response) -> Any:
+    try:
+        return response.json()
+    except json.decoder.JSONDecodeError as exc:
+        raise BlueskyRemoteControlError(
```
tpoliaw: I wonder if this is worth its own exception type? If this raised `NonJsonResponseError` or similar, we could handle it in the `check_connection` wrapper and not have to think about it elsewhere.

With this change we're currently getting `server error with this message: Response does not contain a valid JSON object` which misleadingly suggests the error was received from the server.

src/blueapi/client/rest.py
```diff
@@ -138,6 +139,15 @@ def _create_task_exceptions(response: requests.Response) -> Exception | None:
         return BlueskyRequestError(code, response.text)
 
 
+def _response_json(response) -> Any:
+    try:
+        return response.json()
+    except json.decoder.JSONDecodeError as exc:
```
tpoliaw, line 149: Might be worth logging the text that was received here at debug level? Even with the full stack trace it's not clear what the content was or where it came from. An `Invalid json response from <request url>: <content>` message would be useful when debugging.

src/blueapi/client/rest.py
```diff
@@ -138,6 +139,15 @@ def _create_task_exceptions(response: requests.Response) -> Exception | None:
         return BlueskyRequestError(code, response.text)
 
 
+def _response_json(response) -> Any:
```
tpoliaw:
```suggestion
def _response_json(response: requests.Response) -> Any:
```

tpoliaw (approved):
This is much better, thanks. A couple of more minor comments if you want to look at them but this is a big improvement over the stacktrace it gave before.

PR #1481, opened 2026-04-09
feat: Add cronjob for time-stamper in ConfigMap 

Alexj9837:
**Context**
DAQ core plans to implement auto deletion of unused PVC, for this we need a way of telling how long something hasn't been mounted.

**Description**
By making use of a [Cronjob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) this runs a command every 5 minutes to run a bash script [(time-stamper.sh)](https://github.com/DiamondLightSource/blueapi/blob/6bbaf6dc94ad831d67ccdda523e8d2c768ef447a/helm/blueapi/files/scripts/time-stamper.sh). the bash script loops over the PVC and annotates them, it does this by gathering all PVC's in the blueapi,then all the mounted ones, loops through these annotating the ones that either are mounted or don't have a "last-used" annotation. This is linked via the [config map](https://github.com/DiamondLightSource/blueapi/blob/6bbaf6dc94ad831d67ccdda523e8d2c768ef447a/helm/blueapi/templates/configmap.yaml).


dan-fernandes:
Great work so far.

The script iterates over all PVCs and annotates any PVC that is mounted as well as any PVC that has no annotation. I'm leaning towards leaving PVCs without annotations alone (last use stamp would technically be incorrect / someone is going to inevitably complain about their old PVCs being messed up by the new system, especially if on by default.)

Instead could we just iterate over and mark all mounted PVCs?

Alexj9837:
Deferred label add as waiting for other parts of this feature to be developed.

PR #1482, opened 2026-04-10
feat: Checkout tags as well as branches when cloning into scratch

tpoliaw:
The branch field in the config file can now be set to the value of a
tag. While it could be confusing to use 'branch' for both, it mirrors
the behaviour of the `--branch` flag when cloning repositories with the
git CLI.

In a slight change in behaviour, the branch field is now only used when
first cloning a repository - if it already exists in the scratch area it
will be left in the same state. This is to avoid the ambiguity of how
tags should behave when the repo has been modified since being cloned.

### DIFF
--- a/helm/blueapi/config_schema.json
+++ b/helm/blueapi/config_schema.json
@@ -431,7 +431,7 @@
                     "type": "string"
                 },
                 "branch": {
-                    "description": "Branch of repo to check out - defaults to remote's default when cloning and the existing branch when the repo already exists",
+                    "description": "Branch (or tag) to check out when cloning - defaults to remote's HEAD. If a tag is used, the repo will be left in a 'detached head' state.",
                     "title": "Branch",
                     "type": "string"
                 }
--- a/helm/blueapi/config_schema.json
+++ b/helm/blueapi/config_schema.json
@@ -430,9 +430,9 @@
                     "title": "Remote Url",
                     "type": "string"
                 },
-                "branch": {
-                    "description": "Branch (or tag) to check out when cloning - defaults to remote's HEAD. If a tag is used, the repo will be left in a 'detached head' state.",
-                    "title": "Branch",
+                "target_revision": {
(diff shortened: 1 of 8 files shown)


helm/blueapi/config_schema.json
```diff
@@ -431,7 +431,7 @@
                     "type": "string"
                 },
                 "branch": {
-                    "description": "Branch of repo to check out - defaults to remote's default when cloning and the existing branch when the repo already exists",
+                    "description": "Branch (or tag) to check out when cloning - defaults to remote's HEAD. If a tag is used, the repo will be left in a 'detached head' state.",
```
ZohebShaikh: `targetRevision` Looks more appropriate instead of branch

src/blueapi/cli/scratch.py
```diff
@@ -69,29 +69,17 @@ def ensure_repo(
...
         LOGGER.info(f"Found {local_directory}")
     else:
         raise KeyError(
-            f"Unable to open {local_directory} as a git repository because it is a file"
+            f"Unable to open {local_directory} as a git repository because it is not a "
+            "directory"
         )
 
-    if branch:
-        if not (local := getattr(repo.heads, branch, None)):
```
ZohebShaikh, line 83: if local_directory exists 
can we check where the repo is pointing to the `targetRevision` mentioned in the config and if it is not 
then give a warning

ZohebShaikh (commented):
Looks good only few suggestions

helm/blueapi/config_schema.json
tpoliaw: have changed it to target revision for now but aliased `branch` and `tag` to keep existing configs working

src/blueapi/cli/scratch.py
tpoliaw: Now logs the revision the repo is on and then warns if it is not the same as the target

ZohebShaikh (approved):

PR #1484, opened 2026-04-13
fix: Remove tiled writer if submitted task doesn't exist

### LINKED ISSUE #1480: PUT to `/worker/task` stuck on 409, after a 404
If a non-existent task is PUT to `/worker/task` it correctly returns a 404.

However, any valid existing plans PUT to `/worker/task` following this return a 409, with a stack trace.

## Steps To Reproduce
1. Set up system as in system tests as per the docs: https://diamondlightsource.github.io/blueapi/main/tutorials/local-auth.html
2. PUT to `/worker/task` via the Swagger UI, eg:
```
{
  "task_id": "string"
}
```
4. Observe 404 response
5. PUT a valid task to `/tasks` eg:
```
{
  "name": "count",
  "params": {
    "detectors": [
      "sample_stage"
    ]
  },
  "instrument_session": "cm12345-1"
}
```
7. PUT the outputted task_id to `/worker/task`
8. Obsersve 409

## Acceptance Criteria
- Blueapi no longer 409's valid requests

tpoliaw:
Whenever a task is submitted, a TiledWriter is added to the run engine
and is only removed when the task is complete. If the task never starts
due to not being found, the writer would never be removed and subsequent
tasks would have duplicate writers causing write errors to tiled.

Catching the KeyError and removing the subscribers immediately is a
workaround that works but should be removed when the internal pending
tasks queue is removed.

Fixes #1480

### DIFF
--- a/src/blueapi/service/interface.py
+++ b/src/blueapi/service/interface.py
@@ -221,7 +221,12 @@ def begin_task(
         )
 
     if task.task_id is not None:
-        active_worker.begin_task(task.task_id)
+        try:
+            active_worker.begin_task(task.task_id)
+        except KeyError:
+            active_worker.worker_events.unsubscribe(remove_callback)
+            active_context.run_engine.unsubscribe(tiled_writer_token)
+            raise
     return task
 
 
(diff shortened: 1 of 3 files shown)


dan-fernandes:
Nit: could use a more descriptive name than `ch`

dan-fernandes (approved):

PR #1495, opened 2026-04-17
refactor: Use Depends injection to extract user name

tpoliaw:
And split the auth check into two so that other methods can access the
raw bearer token if required.

### DIFF
--- a/src/blueapi/service/authentication.py
+++ b/src/blueapi/service/authentication.py
@@ -329,3 +329,12 @@ def build_access_token_check(config: OIDCConfig):
 def access_token(request: Request) -> Mapping[str, Any] | None:
     """Get the decoded and verified access token of the user making the request"""
     return getattr(request.state, "decoded_access_token", None)
+
+
+def fedid(
+    access_token: Annotated[Mapping[str, Any] | None, Depends(access_token)],
+) -> str | None:
+    return access_token.get("fedid") if access_token else None
+
+
+Fedid = Annotated[str | None, Depends(fedid)]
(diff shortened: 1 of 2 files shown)


src/blueapi/service/authentication.py
```diff
@@ -325,3 +325,12 @@ def validate_bearer_token(request: Request, token: UncheckedBearerToken):
 def access_token(request: Request) -> Mapping[str, Any] | None:
     """Get the decoded and verified access token of the user making the request"""
     return getattr(request.state, "decoded_access_token", None)
+
+
+def fedid(
+    access_token: Annotated[Mapping[str, Any] | None, Depends(access_token)],
+) -> str | None:
+    return access_token.get("fedid") if access_token else None
```
ZohebShaikh, line 337: We can use a config variable in blueapi which defaults to `fedid`  called userIDClaim.

Below copied from oauth2-proxy [docs](https://oauth2-proxy.github.io/oauth2-proxy/configuration/alpha-config?_highlight=useridclaim#oidcoptions)
```
userIDClaim	string	UserIDClaim indicates which claim contains the user ID
default set to 'email'
```
This is a very nit comment. Something like the pot calling the kettle black. I don't use it in the Oauth2-proxy config that we have but I should start to use it.

ZohebShaikh (approved):

src/blueapi/service/authentication.py
tpoliaw: Agree that it would be good to fix it but as the existing code hardcoded `fedid` I think it can stay for now and be updated in a later PR.

PR #1498, opened 2026-04-20
chore: Add justfile for common tasks

tpoliaw:
Simplifies the various common tasks that should be run during
development.

Can be added to over time but for now mainly wraps running compose
services and calling blueapi so config files don't have to be specified
for every command.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,4 +1,4 @@
-set dotenv-filename := "tests/system_tests/.env"
+SESSION := "cm12345-1"
 
 compose +ARGS="up -d":
     docker compose -f tests/system_tests/compose.yaml {{ARGS}}
@@ -11,8 +11,12 @@ configure-adsim: (compose "exec" "numtracker" "/app/numtracker" "client" "config
 
 services: compose configure-adsim
 
-blueapi *ARGS="serve":
-    uv run blueapi -c tests/system_tests/config.yaml {{ARGS}}
+serve *OPTS: services
+    source tests/system_tests/.env
+    uv run blueapi -c tests/system_tests/config.yaml {{OPTS}} serve
+
+run PLAN PARAMS:
+    uv run blueapi -c tests/system_tests/config.yaml controller run -i {{ SESSION }} {{ PLAN }} '{{ PARAMS }}'
 
 lint:
     uv run ruff check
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -0,0 +1,25 @@
+set dotenv-filename := "tests/system_tests/.env"
+
+compose +ARGS="up -d":
```
Alexj9837, line 3: could this be "pcompose" as the [docs](https://diamondlightsource.github.io/blueapi/main/tutorials/local-auth.html) use docker and maybe have "compose" just for docker?

tpoliaw: Sure. I think both `docker compose` and `podman compose` will redirect to the same binary anyway so we can probably stick with one.

justfile
```diff
@@ -0,0 +1,25 @@
...
+configure-adsim: (compose "exec" "numtracker" "/app/numtracker" "client" "configure" "adsim"
+        "--directory" '/tmp/'
+        "--scan" '{instrument}-{scan_number}'
+        "--detector" '{instrument}-{scan_number}-{detector}'
+        "--number" "43")
+
+services: compose configure-adsim
+
+blueapi *ARGS="serve":
+    uv run blueapi -c tests/system_tests/config.yaml {{ARGS}}
```
tpoliaw: This doesn't work well when running plans.

    just blueapi controller run sleep '{"time": 3}'

loses the quotes around the JSON parameters and becomes

    uv run blueapi controller run sleep {"time": 3}

which fails with `Parameters are not valid JSON`.

tpoliaw: Have split into serve and run for now

Alexj9837:
This is a cool feat, could we also add an doc update to reference this ?

tpoliaw:
I imagine the tasks are going to need adjustments when they meet every day use so I'm tempted to leave it as an undocumented development thing for now and then write it up when they've settled down a bit.

justfile
```diff
@@ -0,0 +1,29 @@
...
+
+serve *OPTS: services
+    source tests/system_tests/.env
+    uv run blueapi -c tests/system_tests/config.yaml {{OPTS}} serve
+
+run PLAN PARAMS:
+    uv run blueapi -c tests/system_tests/config.yaml controller run -i {{ SESSION }} {{ PLAN }} '{{ PARAMS }}'
+
+lint:
+    uv run ruff check
```
Alexj9837:
```suggestion
    uv run ruff check
    uv run ruff format 
```
CI runs ruff check and format, might be useful to add it also 
Image

tpoliaw: Might be worth running the full pre-commit check

Alexj9837: makes sense

Alexj9837 (approved):

PR #1499, opened 2026-04-22
feat(scratch): Make scratch init faster

### LINKED ISSUE #1417: Improve time required to clone repositories in initContainer
Currently the repos are cloned in sync and it clones everything(default)

The solution that I propose to this is 

1. Using Asycn IO to do the clone ( There is no native support for this but it is easy to implement more details [here](https://github.com/gitpython-developers/GitPython/discussions/1330))

2. Do git clone with the following option `--filter=blob:none` more details [here](https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/) 

There is `--filter=tree:0` which is also a good option but in my personal opinion going with blob:none is better because the thing that we want to avoid cloning is the [gh-pages](https://github.com/DiamondLightSource/blueapi/tree/gh-pages) branch but clone most of the source code which the blob:none is good at.

Here is the size difference for the following options for blueapi in terms of size

```
311M    /tmp/blueapi/
11M     /tmp/blueapi-blob-none/ # --filter=blob:none
6.9M    /tmp/blueapi-tree-0/ # --filter=tree:0
```

And for dodal much bigger improvement

```
585M    /tmp/dodal/
33M     /tmp/dodal-blob-none/ #--filter=blob:none
9.8M    /tmp/dodal-tree-0/ # --filter=tree:0
```

To clone dodal without any options takes 1m35s 

This is a better approach and will definetly improve the time required for the initContainer to start

## Acceptance Criteria
- InitContainer start time with large git repositories significantly faster than before

ZohebShaikh:
closes #1417

### DIFF
--- a/src/blueapi/cli/scratch.py
+++ b/src/blueapi/cli/scratch.py
@@ -87,7 +87,7 @@ def ensure_repo(
 
     if not local_directory.exists():
         LOGGER.info(f"Cloning {remote_url}")
-        repo = Repo.clone_from(
+        Repo.clone_from(
             remote_url,
             local_directory,
             branch=target_revision,
@@ -119,8 +119,8 @@ def ensure_repo(
 def scratch_install(*paths: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) -> None:
     """
     Install scratch packages. Make blueapi aware of repositories checked out in
-    the scratch area. Make it automatically follow code changes to that repositories
-    (pending a restart). Do not install any of the package's dependencies as they
+    the scratch area. Make it automatically follow code changes to those repositories
+    (pending a restart). Do not install any of the packages' dependencies as they
     may conflict with each other.
 
     Args:
--- a/src/blueapi/cli/scratch.py
+++ b/src/blueapi/cli/scratch.py
@@ -118,7 +118,7 @@ def ensure_repo(
 
 def scratch_install(*paths: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) -> None:
     """
-    Install all scratch packages. Make blueapi aware of repositories checked out in
+    Install scratch packages. Make blueapi aware of repositories checked out in
     the scratch area. Make it automatically follow code changes to that repositories
     (pending a restart). Do not install any of the package's dependencies as they
     may conflict with each other.
--- a/src/blueapi/cli/scratch.py
+++ b/src/blueapi/cli/scratch.py
@@ -65,7 +65,7 @@ def setup_scratch(
                 raise RuntimeError("Failed to clone repositories") from exc
 
     scratch_install(
-        [config.root / repo.name for repo in config.repositories],
+        *(config.root / repo.name for repo in config.repositories),
         timeout=install_timeout,
     )
 
@@ -91,7 +91,7 @@ def ensure_repo(
             remote_url,
             local_directory,
             branch=target_revision,
-            multi_options=["--filter=blob:none"],
+            filter="blob:none",
         )
         LOGGER.info(f"Cloned {remote_url} -> {local_directory}")
     elif local_directory.is_dir():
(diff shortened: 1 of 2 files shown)


src/blueapi/cli/scratch.py
```diff
@@ -69,58 +83,63 @@ def ensure_repo(
...
+        paths: List of Path to the checked out repositories
         timeout: Time to wait for installation subprocess
     """
-
-    _validate_directory(path)
-
-    LOGGER.info(f"Installing {path}")
-    process = Popen(
-        [
+    if paths:
```
tpoliaw: nit: `if not paths: return` would stop the whole function being indented

src/blueapi/cli/scratch.py
```diff
@@ -93,34 +111,37 @@ def ensure_repo(
         )
 
 
-def scratch_install(path: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) -> None:
+def scratch_install(
+    paths: list[Path], timeout: float = _DEFAULT_INSTALL_TIMEOUT
```
tpoliaw:
```suggestion
    *paths: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT
```

Would prevent the API changing for existing calls

src/blueapi/cli/scratch.py
```diff
@@ -93,34 +111,37 @@ def ensure_repo(
...
+) -> None:
     """
-    Install a scratch package. Make blueapi aware of a repository checked out in
+    Install all scratch package. Make blueapi aware of a repository checked out in
```
tpoliaw:
```suggestion
    Install scratch packages. Make blueapi aware of repositories checked out in
```

(and a couple more plural consistency changes in the doc string)

src/blueapi/cli/scratch.py
```diff
@@ -69,7 +87,12 @@ def ensure_repo(
 
     if not local_directory.exists():
         LOGGER.info(f"Cloning {remote_url}")
-        Repo.clone_from(remote_url, local_directory, branch=target_revision)
+        repo = Repo.clone_from(
+            remote_url,
+            local_directory,
+            branch=target_revision,
+            multi_options=["--filter=blob:none"],
```
tpoliaw: I think you can pass this as a keyword
```suggestion
            filter="blob:none",
```

src/blueapi/cli/scratch.py
```diff
@@ -69,7 +87,12 @@ def ensure_repo(
 
     if not local_directory.exists():
         LOGGER.info(f"Cloning {remote_url}")
-        Repo.clone_from(remote_url, local_directory, branch=target_revision)
+        repo = Repo.clone_from(
```
tpoliaw:
```suggestion
       Repo.clone_from(
```

src/blueapi/cli/scratch.py
```diff
@@ -93,34 +116,36 @@ def ensure_repo(
...
 
 
-def scratch_install(path: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) -> None:
+def scratch_install(*paths: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) -> None:
     """
-    Install a scratch package. Make blueapi aware of a repository checked out in
-    the scratch area. Make it automatically follow code changes to that repository
+    Install scratch packages. Make blueapi aware of repositories checked out in
+    the scratch area. Make it automatically follow code changes to that repositories
     (pending a restart). Do not install any of the package's dependencies as they
```
tpoliaw:
```suggestion
    the scratch area. Make it automatically follow code changes to those repositories
   (pending a restart). Do not install any of the packages' dependencies as they
```

tpoliaw (approved):
Looks good, thanks

PR #1501, opened 2026-04-22
feat: Add helper print commands to easily print avaliable devices and plans to user

oliwenmandiamond:
Add small feature to quickly show available devices and plans to user via printing their name from client.

```Python
>>> bc.print_devices()
psk1
psj1
dcm
psi2
jgap
fsj1
...

>>> bc.print_plans()
count(detectors, num=None, delay=None, metadata=None)
spec_scan(detectors, spec, metadata=None)
move(moves, group=None)
move_relative(moves, group=None)
set_absolute(movable, value, group=None, wait=None)
set_relative(movable, value, group=None, wait=None)
sleep(time)
wait(group=None, timeout=None)
...
```

### DIFF
--- a/src/blueapi/client/client.py
+++ b/src/blueapi/client/client.py
@@ -295,8 +295,8 @@ class BlueapiClient:
         """
         return self._rest.get_plan(name)
 
-    def show_plans(self) -> None:
-        """Command to print all avaliable plans."""
+    def print_plans(self) -> None:
+        """Print all avaliable plans."""
         for name in self.plans:
             print(name)
 
@@ -339,8 +339,8 @@ class BlueapiClient:
 
         return self._rest.get_device(name)
 
-    def show_devices(self) -> None:
-        """Command to print all avaliable devices."""
+    def print_devices(self) -> None:
+        """Print all avaliable devices."""
         for name in self.devices:
             print(name)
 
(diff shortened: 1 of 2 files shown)


src/blueapi/client/client.py
```diff
@@ -299,6 +299,11 @@ def get_plan(self, name: str) -> PlanModel:
         """
         return self._rest.get_plan(name)
 
+    def show_plans(self) -> None:
+        """Command to print all avaliable plans."""
```
tpoliaw:
```suggestion
        """Print all available plans."""
```

src/blueapi/client/client.py
```diff
@@ -338,6 +343,11 @@ def get_device(self, name: str) -> DeviceModel:
 
         return self._rest.get_device(name)
 
+    def show_devices(self) -> None:
+        """Command to print all avaliable devices."""
```
tpoliaw:
```suggestion
        """Print all available devices."""
```

tpoliaw (commented):
nit: I think `print_plans`/`print_devices` would be more descriptive for what the functions are doing.

src/blueapi/client/client.py
oliwenmandiamond: Done

src/blueapi/client/client.py
oliwenmandiamond: Done

tpoliaw (approved):

PR #1514, opened 2026-04-24
feat: Pvc auto deletion logic

Alexj9837:
Adds a new autodeletion cronjob that automattically deletes stale PVSs based on the last-used annotation, written by the time-stamper cronjob.

**how it works:**

- Deletion bash script iterates over all the PVCs 
- If a PVC has a last-used older than 3 months it gets deleted
- Unless it has the annotation of protected = true


helm/blueapi/templates/configmap.yaml
```diff
@@ -31,6 +31,6 @@ data:
   init_config.yaml: |-
     scratch:
       {{- toYaml .Values.worker.scratch | nindent 6 }}
-{{- end }}
 
----
+---    
+{{- end }}
```
ZohebShaikh: This change should be reverted

helm/blueapi/templates/cronjob.yaml
```diff
@@ -0,0 +1,169 @@
...
+kind: CronJob
+metadata:
+  name: {{ include "blueapi.fullname" . }}-last-used-stamper
+  namespace: {{ .Release.Namespace }}
+spec:
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 1
+  failedJobsHistoryLimit: 1
+  schedule: "*/5 * * * *"
+
```
ZohebShaikh, line 43: Running every 5 minutes looks like a lot to me. I think it should be once a day at 1200 pm So if the job fails the alert comes during working time

helm/blueapi/templates/cronjob.yaml
```diff
@@ -0,0 +1,169 @@
...
+          - name: pvc-auto-deletion
+            env:
+              - name: RELEASE_NAME
+                value: {{ .Release.Name }}
+              - name: RELEASE_NAMESPACE
+                value: {{ .Release.Namespace }}
+            volumeMounts:
+              - name: {{include "blueapi.fullname" . }}-pvc-auto-deletion-script
+                mountPath: /scripts
+            image: bitnami/kubectl:latest
```
ZohebShaikh: More info [here](https://github.com/bitnami/containers/issues/83267) about why we should not use bitnami 

```suggestion
                mountPath: /scripts
            image: rancher/kubectl@sha256:05d2b313e2f397e0ade252136aed47abd72d56ead11d1b027ac70f66362c8495 # v1.36.0
```

helm/blueapi/values.yaml
ZohebShaikh, line 39: This change should be reverted

helm/blueapi/values.yaml
```diff
@@ -47,7 +47,7 @@ securityContext:
   #   drop:
   #   - ALL
 
-# This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/
+  # This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/
```
ZohebShaikh: This change should be reverted

helm/blueapi/values.yaml
```diff
@@ -75,13 +75,13 @@ ingress:
...
   #      - chart-example.local
 
-# -- Sets the compute resources available to the pod.
-# These defaults are appropriate when using debug mode or an internal PVC and therefore
-# running VS Code server in the pod.
-# In the Diamond cluster, requests must be >= 0.1*limits
-# When not using either of the above, the limits may be lowered.
-# When idle but connected, blueapi consumes ~400MB of memory and 1% cpu
-# and may struggle when allocated less.
+  # -- Sets the compute resources available to the pod.
```
ZohebShaikh: Same here
This is leading to the docs getting removed

helm/blueapi/values.yaml
```diff
@@ -209,7 +204,7 @@ worker:
     repositories: []
     # - name: "dodal"
     #   remote_url: https://github.com/DiamondLightSource/dodal.git
-  # -- Configures logging. Port 12231 is the `dodal` input on graylog which will be renamed `blueapi`
+    # -- Configures logging. Port 12231 is the `dodal` input on graylog which will be renamed `blueapi`
```
ZohebShaikh: Same here

helm/blueapi/values.yaml
```diff
@@ -228,6 +223,12 @@ initContainer:
     # -- Size of persistent volume
     size: "1Gi"
 
+timeStampCron:
+  enabled: true
+
+pvcAutoDeletion:
+  enabled: true
```
ZohebShaikh, line 238: Can the docs be added for the protected label to prevent it from deletion ?
and In general docs for both of the options

helm/blueapi/files/scripts/time-stamper.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# Get all PVCs currently mounted by running pods
+MOUNTED_PVCS=$(kubectl get pods -n  $RELEASE_NAMESPACE  \
+  -o=jsonpath='{.items[*].spec.volumes[*].persistentVolumeClaim.claimName}' | tr ' ' '\n' | sort -u)
+BLUEAPI_PVCS=$( echo $MOUNTED_PVCS | tr ' ' '\n' | grep blueapi-scratch)
```
ZohebShaikh: This can come as `{{ include "blueapi.fullname" . }}-scratch-` as argument to the script 
and The regex should be  `^p46-blueapi-scratch-` i.e. starts with ^

helm/blueapi/files/scripts/pvc-deletion.sh
```diff
@@ -0,0 +1,26 @@
...
+        if [ -n "$LAST_USED"  ]; then
+            #check if last_used is older than 3 months
+            if [ $(($NOW - LAST_USED)) -gt 7884000 ]; then
+                #checking if the pvc is protected, if it is protected skip deletion
+                if  [ "$(kubectl get pvc "$pvc" -n  $RELEASE_NAMESPACE  -o=jsonpath='{.metadata.annotations.protected}')" = "true" ]; then
+                    echo " PVC $pvc is protected, skipping deletion"
+                    continue
+                fi
+                #PVC has not been used for more than three months, delete it
+                kubectl delete pvc "$pvc" -n  $RELEASE_NAMESPACE
```
ZohebShaikh: I have not explored this but is there a timeout that can be set in the script
Because if the PVC is being used it will be not deleted and this job will left in a hanging state.

helm/blueapi/files/scripts/pvc-deletion.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+# Get all PVCs by running pods
+ALL_PVCS=$(kubectl get pvc -n  $RELEASE_NAMESPACE  -o=jsonpath='{.items[*].metadata.name}' | tr ' ' '\n' | sort -u)
+BLUEAPI_PVCS=$( echo $ALL_PVCS | tr ' ' '\n' | grep blueapi-scratch)
```
ZohebShaikh: Same as below for the `blueapi-scratch`

helm/blueapi/files/scripts/pvc-deletion.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
ZohebShaikh: The docs for setting flags in script [here](https://expeditor.chef.io/docs/patterns/bash-scripts/#troubleshooting)

ZohebShaikh (changes requested):
Looks good...
Just few things to tidy up
I think the other [PR#1481](https://github.com/DiamondLightSource/blueapi/pull/1481) should be closed as it merged into this

helm/blueapi/files/scripts/pvc-deletion.sh
Alexj9837: done

helm/blueapi/files/scripts/pvc-deletion.sh
Alexj9837: done

Alexj9837:
Work is being paused, till next spint

PR #1524, opened 2026-04-28
feat: Add tag for deprecated endpoints

ZohebShaikh:


tpoliaw:
What is the benefit of adding this as a tag? As far as I can see, with the `Deprecated` tag, the endpoints lose their grouping on the docs page and get bundled into one group at the end. Without the tag (but with the existing `deprecated=True` flag), the endpoints are in their original groups but greyed out. In the openapi schema, the old endpoints already have a `deprecated=true` field so the tag doesn't really add anything there.

ZohebShaikh:
I personally feel that it looks neater to see the docs page like that.

I'm not fussed about it. Will close the PR

PR #1527, opened 2026-04-29
fix: move worker-config from values.yaml volumeMount to statefulset.yaml volumeMount

### LINKED ISSUE #1451: BlueAPI Config Mount Easily Overwriteable
The BlueAPI helm deployment requires a configuration file mounted at `/config/config.yaml` in order to run. This is badly implemented, being easily overwritten when other other custom volumes are mounted.

## Current Implementation
The config volume is defined in `statefulset.yaml` [here](https://github.com/DiamondLightSource/blueapi/blob/18e04caaa36aeb2900f927bd8fa89b25ad024e30/helm/blueapi/templates/statefulset.yaml#L43), while the volume mount is defined in `.Values.volumeMounts` [here](https://github.com/DiamondLightSource/blueapi/blob/18e04caaa36aeb2900f927bd8fa89b25ad024e30/helm/blueapi/values.yaml#L137) and propagated into the `StatefulSet` [here](https://github.com/DiamondLightSource/blueapi/blob/18e04caaa36aeb2900f927bd8fa89b25ad024e30/helm/blueapi/templates/statefulset.yaml#L146).

This functions, but can be broken by anyone defining their own `.Values.volumeMounts` in a custom values file, which would remove the config mount. This has caused issues for a support engineer on I19.

I think the intention behind this was to allow mounting user-written config files at `/config/config.yaml`. However, configuration can already be customised through user-written `Values.yaml` files, and I prefer this method.

## Acceptance Criteria
- The configuration file mounted at `/config/config.yaml` is not overridden when custom values are set in `.Values.volumeMounts`

shree-iyengar-dls:
Fixes #1451 

Tested on p48:
- created a new PVC which was referenced in the p48-blueapi-0 pod under volumes: 
```
 name: my-pv-claim
    persistentVolumeClaim:
      claimName: my-pv-claim
```
- The PV was also bound:
 ``` 
pvc-905772f8-0fe1-4d8d-8186-4838473eb289      1Gi        RWX            Delete           Bound       p48-beamline/my-pv-claim
```

### CODE REVIEW

File: helm/blueapi/values.yaml [hunk 1]
```diff
@@ -48,7 +47,7 @@ securityContext:
   #   drop:
   #   - ALL
 
-# This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/
+  # This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/
```
dan-fernandes: The indentation here and on lines 79-85 and 208 have removed some useful documentation from the values.schema.json. Probably just you IDE autoformatting files.

Unindent to fix

helm/blueapi/values.yaml
```diff
@@ -134,10 +133,7 @@ volumes: []
 
 # -- Additional volumeMounts on the output StatefulSet definition.
 # Define how volumes are mounted to the container referenced by using the same name.
-volumeMounts:
-  - name: worker-config
-    mountPath: "/config"
-    readOnly: true
+volumeMounts: []
```
dan-fernandes, line 140: Could add a commented-out example of a volumeMount, as per extraEnvVars in the same file. That way people can see what to expect.

dan-fernandes (changes requested):
Looks really good! Very minor corrections

helm/blueapi/values.yaml
```diff
@@ -205,7 +209,7 @@ worker:
     repositories: []
     # - name: "dodal"
     #   remote_url: https://github.com/DiamondLightSource/dodal.git
-  # -- Configures logging. Port 12231 is the `dodal` input on graylog which will be renamed `blueapi`
```
dan-fernandes, line 208: This line is now too unindented

dan-fernandes (changes requested):

dan-fernandes (approved):

PR #1528, opened 2026-04-29
fix: Install debugpy in container instead of adding it

### LINKED ISSUE #1529: Incorrect version reported by blueapi running from container
The version reported by blueapi running from a container does not match the container being run. When running a server, this appears as the client failing its version check (`Version mismatch: Blueapi server version is X but client version is Y`).

The simplest demonstration is to run `blueapi --version` using the image.

```
$ podman run --rm ghcr.io/diamondlightsource/blueapi:1.14.0 --version
blueapi, version 1.14.1.dev0+gaf0695875.d20260424
```
## Acceptance Criteria
- The version reported by blueapi in the container matches the version of the container

tpoliaw:
Running `uv add debugpy` in the container modifies pyproject.toml and
uv.lock so `git describe` reports local changes in the repo and appends
a `dev0+g<commit>.d<date>` suffix to the version used for the blueapi in
the container

Using `uv pip install debugpy` makes it available in the environment
without modifying any tracked files.

This has to be run after the `uv sync` to stop it being removed again.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get update -y && apt-get install -y --no-install-recommends \
     graphviz \
     && apt-get dist-clean
 
-# Install helm for the dev container. This is the recommended 
+# Install helm for the dev container. This is the recommended
 # approach per the docs: https://helm.sh/docs/intro/install
 RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3; \
     chmod 700 get_helm.sh; \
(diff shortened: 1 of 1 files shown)


ZohebShaikh (commented):
Can a check be added [here](https://github.com/DiamondLightSource/blueapi/blob/d3364b343f07161832ca778ac779a83a82eb2255/.github/workflows/_container.yml#L56) 
If publish == true ..version == tagged_version && 
if publish is false version is semver compliant 
So we can have some regression test for it

abbiemery (approved):

tpoliaw:
I don't understand why the version number is being mangled by the setuptools-scm. `1.14.1-dev.1` is becoming `1.14.1.dev1` and I haven't figured out why yet.

ZohebShaikh (approved):
For the test we can create another issue, lets get this merged as is

tpoliaw:
Have removed the test commit from this PR and moved it to #1561

PR #1532, opened 2026-04-30
ci: Do not run PR title check in merge queue

ZohebShaikh:


tpoliaw:
Does this still mark the job as passing?

ZohebShaikh:
I think what it does is...Just not run the PR lint check when the PR is inside a merge queue 

Can I get a approve to just test it, I'll remove it from the queue if it doesn't work

tpoliaw (approved):

PR #1538, opened 2026-05-08
refactor: improve numtracker error messages

### LINKED ISSUE #1507: Give better error message on non-existant instrument session
If I accidentally give a non-existent instrument session when submitting a plan to BlueAPI I get an error like:

```
RuntimeError: Numtracker error: [{'message': 'Authentication failed', 'locations': [{'line': 3, 'column': 5}], 'path': ['scan']}]
```

this doesn't really give me any hints to why the authentication has failed.

## Acceptance Criteria
- A failure on the above gives a more useful response like "Visit XXX not found"

shree-iyengar-dls:
Fixes #1507 

Summary of changes:
- Added error extensions in numtracker.rs (see https://github.com/DiamondLightSource/numtracker/pull/275) 
- Made changes to NumtrackerClient in blueapi

### DIFF
--- a/src/blueapi/utils/numtracker.py
+++ b/src/blueapi/utils/numtracker.py
@@ -117,8 +117,7 @@ class NumtrackerClient:
                         raise RuntimeError("Numtracker authentication missing")
                     case "AUTH_SERVER_ERROR":
                         raise RuntimeError("Numtracker server authentication error")
-                    case _:
-                        raise RuntimeError(f"Numtracker error: {json['errors']}")
+            raise RuntimeError(f"Numtracker error: {json['errors']}")
 
         new_collection = NumtrackerScanMutationResponse.model_validate(json["data"])
         LOGGER.debug("New NumtrackerNewScan: %s", new_collection)
--- a/src/blueapi/utils/numtracker.py
+++ b/src/blueapi/utils/numtracker.py
@@ -116,7 +116,7 @@ class NumtrackerClient:
                     case "AUTH_MISSING":
                         raise RuntimeError("Numtracker authentication missing")
                     case "AUTH_SERVER_ERROR":
-                        raise RuntimeError("Server authentication error")
+                        raise RuntimeError("Numtracker server authentication error")
                     case _:
                         raise RuntimeError(f"Numtracker error: {json['errors']}")
 
--- a/src/blueapi/utils/numtracker.py
+++ b/src/blueapi/utils/numtracker.py
@@ -105,6 +105,14 @@ class NumtrackerClient:
         json = response.json()
 
         if json.get("errors") is not None:
+            for error in json["errors"]:
+                code = error.get("extensions").get("code")
+
+                if code == "AUTH_FAILED":
+                    raise RuntimeError(
+                        f"Not authorised to create a scan number for "
+                        f"{instrument} and {instrument_session}."
+                    )
             raise RuntimeError(f"Numtracker error: {json['errors']}")
 
         new_collection = NumtrackerScanMutationResponse.model_validate(json["data"])
--- a/src/blueapi/utils/numtracker.py
+++ b/src/blueapi/utils/numtracker.py
@@ -106,14 +106,19 @@ class NumtrackerClient:
 
         if json.get("errors") is not None:
             for error in json["errors"]:
-                code = (error.get("extensions") or {}).get("code")
-
-                if code == "AUTH_FAILED":
-                    raise RuntimeError(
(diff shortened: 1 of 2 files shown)


src/blueapi/utils/numtracker.py
```diff
@@ -105,6 +105,14 @@ async def create_scan(
         json = response.json()
 
         if json.get("errors") is not None:
+            for error in json["errors"]:
+                code = (error.get("extensions") or {}).get("code")
```
tpoliaw: You can pass a default value to `get` to avoid the ` or {}`.


```suggestion
                code = error.get("extensions", {}).get("code")
```

src/blueapi/utils/numtracker.py
```diff
@@ -105,6 +105,14 @@ async def create_scan(
...
+
+                if code == "AUTH_FAILED":
```
tpoliaw: Do we want to improve the error message for the other cases while we're here? Could be a good use for match - something like

```python
match code:
    case "AUTH_FAILED":
        raise ...
    case "AUTH_MISSING":
        raise ...
    case "AUTH_SERVER_ERROR":
        raise ...
```
and then the fall-through case could be the existing `Numtracker error: ...` we were previously raising.

tests/unit_tests/utils/test_numtracker.py
```diff
@@ -33,6 +33,18 @@ def numtracker() -> NumtrackerClient:
     ],
 }
 
+AUTH_ERROR = {
+    "data": None,
+    "errors": [
+        {
+            "message": "No configuration available for instrument p46",
```
tpoliaw: really minor but having the message match the error might prevent confusion later on

tpoliaw (approved):
Looks good. None of the comments are blockers

src/blueapi/utils/numtracker.py
shree-iyengar-dls: Yuss that's a good idea thanks

src/blueapi/utils/numtracker.py
```diff
@@ -105,7 +105,20 @@ async def create_scan(
...
+                        raise RuntimeError(
+                            f"Not authorised to create a scan number for "
+                            f"{instrument} and {instrument_session}"
+                        )
+                    case "AUTH_MISSING":
+                        raise RuntimeError("Numtracker authentication missing")
+                    case "AUTH_SERVER_ERROR":
+                        raise RuntimeError("Server authentication error")
+                    case _:
+                        raise RuntimeError(f"Numtracker error: {json['errors']}")
```
tpoliaw: With the fallback being in the match statement, the for loop will never get past the first iteration. It probably doesn't matter given we're unlikely to get multiple errors but moving out to the level of the for loop would mean an auth error in a later error would be raised with a useful error.

```suggestion
            raise RuntimeError(f"Numtracker error: {json['errors']}")
```

src/blueapi/utils/numtracker.py
```diff
@@ -105,7 +105,20 @@ async def create_scan(
...
+                match code:
+                    case "AUTH_FAILED":
+                        raise RuntimeError(
+                            f"Not authorised to create a scan number for "
+                            f"{instrument} and {instrument_session}"
+                        )
+                    case "AUTH_MISSING":
+                        raise RuntimeError("Numtracker authentication missing")
+                    case "AUTH_SERVER_ERROR":
+                        raise RuntimeError("Server authentication error")
```
tpoliaw: This should probably specify that _numtracker_ had the server error

tests/unit_tests/utils/test_numtracker.py
```diff
@@ -109,3 +109,41 @@ async def test_create_scan_raises_runtime_error_on_graphql_error(
...
         await numtracker.create_scan("ab123", "p46")
+
+
+@pytest.mark.parametrize(
+    ("auth_error_code", "expected_message"),
+    [
+        ("AUTH_FAILED", "Not authorised to create a scan number for p46 and ab123"),
+        ("AUTH_MISSING", "Numtracker authentication missing"),
+        ("AUTH_SERVER_ERROR", "Server authentication error"),
+        ("UNKWNOWN_ELSE", "Numtracker error:"),
```
tpoliaw:
```suggestion
        ("UNKNOWN_ELSE", "Numtracker error:"),
```

tpoliaw (approved):

tpoliaw:
This looks good now - thanks

PR #1539, opened 2026-05-15
feat: add authz checks to API

shree-iyengar-dls:
Require authz checks as part of endpoints that modify state of the blueapi deployment either by running plans or changing existing tasks.


src/blueapi/service/main.py
```diff
@@ -278,6 +279,16 @@ def get_device_by_name(
 )
 
 
+def submission_check(
+    opa: Annotated[OPAClient, Depends(get_opa_client)],
```
ZohebShaikh: I think there will be multiple of these depends 
1. submit_task
2. get tasks
3. delete_tasks
....
Basically most of the endpoint in the tasks group

and as we will have multiple of these we can move them closed to where OPAClient will be

src/blueapi/service/main.py
```diff
@@ -146,6 +152,33 @@ def get_app(config: ApplicationConfig):
...
 
+def access_task_permission(
+    opa: Annotated[OpaUserClient | None, Depends(opa)],
+    task_id: str,
+    fedid: Fedid,
+    runner: Annotated[WorkerDispatcher, Depends(_runner)],
+):
+    task = runner.run(interface.get_task_by_id, task_id)
+
+    if opa and not opa.admin() and (task and fedid != task.task.metadata.get("user")):
```
tpoliaw: This only checks task access if OPA is configured. Do we want to restrict access based on user name when authn is enabled but authz is not?

src/blueapi/service/authorization.py
```diff
@@ -1,15 +1,20 @@
...
+from fastapi import Depends, HTTPException, Request
+from starlette.status import HTTP_401_UNAUTHORIZED, HTTP_403_FORBIDDEN
 
 from blueapi.config import OIDCConfig, OpaConfig, ServiceAccount
-from blueapi.service.authentication import TiledAuth
+from blueapi.service.authentication import TiledAuth, unchecked_bearer_token
+from blueapi.service.model import TaskRequest
 
 LOGGER = logging.getLogger(__name__)
+INSTRUMENT_SESSION_RE = re.compile(r"^[a-z]{2}(?P<proposal>\d+)-(?P<visit>\d+)$")
```
ZohebShaikh: nit: Can we use the same value [here](https://github.com/DiamondLightSource/blueapi/blob/ab2ebc635684fc107bf2992339015ed1f18d1bb1/src/blueapi/utils/serialization.py#L32) ? Maybe move it to a constants file and import it from there

I had a look at the ISypB database they have all lower-case first 2 characters. So the above link regex should drop A-Z

We never make any check on the code(cm,sm..) of the visit, so why bother checking if it is 2 character and a-z

because it will pass even if you write cm12345-1 or ab12345-1

src/blueapi/service/main.py
```diff
@@ -496,6 +539,16 @@ def set_state(
...
     ):
+        active = runner.run(interface.get_active_task)
+
+        if (
+            opa
+            and not opa.admin()
+            and active
+            and active.task.metadata.get("user") != fedid
+        ):
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN)
```
ZohebShaikh: In general for all HTTPException Can we have content="Not authorised to set worker state" etc

src/blueapi/config.py
```diff
@@ -299,6 +299,8 @@ class Tag(StrEnum):
 class OpaConfig(BlueapiBaseModel):
     root: HttpUrl = HttpUrl("http://localhost:8181")
     tiled_service_account_check: str
+    submit_task_check: str
+    admin_check: str
```
ZohebShaikh, line 317: I think I have written this in some other comments as well but just to be safe
Can we have the default values to the policy so it is slightly smaller change to enable authZ

src/blueapi/service/main.py
```diff
@@ -470,6 +510,9 @@ def get_state(runner: Annotated[WorkerDispatcher, Depends(_runner)]) -> WorkerSt
 def set_state(
     state_change_request: StateChangeRequest,
     response: Response,
+    fedid: Fedid,
+    opa: Annotated[OpaUserClient, Depends(opa)],
+    # _: Annotated[None, Depends(access_task_permission)],
```
ZohebShaikh:
```suggestion
    _: Annotated[None, Depends(access_task_permission)],
```

tests/unit_tests/service/test_authorization.py
```diff
@@ -105,6 +112,105 @@ async def test_opa_adds_input_fields(session: MagicMock, opa_config: OpaConfig):
...
+    )
+
+
+@patch_client_session
+async def test_opa_require_submit_task_invalid_session(
+    session: MagicMock, opa_config: OpaConfig
+):
+    client = OpaClient(instrument="p45", config=opa_config)
+
+    with pytest.raises(ValueError):
```
ZohebShaikh: match = "" ?

ZohebShaikh (commented):
Looks Good, few comments

src/blueapi/service/authorization.py
tpoliaw: I've moved it to the root of utils for now until there are enough constants to be worth a module. 

I think it makes sense to keep the prefix check to ensure that `123-456` isn't accepted as a valid visit. We might be able to confirm the correct valid in future.

src/blueapi/service/main.py
tpoliaw: I think this should be removed completely. There isn't a task in the input to be accessed.

src/blueapi/config.py
tpoliaw: Noted. Conversation can be kept to the other comment.

src/blueapi/service/authorization.py
shree-iyengar-dls, line 18: Nit: might be helpful to have a brief comment explaining the difference between OpaClient and OpaUserClient

src/blueapi/service/main.py
```diff
@@ -146,6 +152,37 @@ def get_app(config: ApplicationConfig):
...
+    runner: Annotated[WorkerDispatcher, Depends(_runner)],
+):
+    task = runner.run(interface.get_task_by_id, task_id)
+
+    if (
+        opa
+        and not await opa.admin()
+        and (task and fedid != task.task.metadata.get("user"))
+    ):
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
```
shree-iyengar-dls, line 169: How come 404 not found rather than 401 unauthorised?

src/blueapi/service/main.py
```diff
@@ -346,6 +386,10 @@ def get_tasks(
         tasks = runner.run(interface.get_tasks_by_status, desired_status)
     else:
         tasks = runner.run(interface.get_tasks)
+
+    if opa and not await opa.admin():
```
shree-iyengar-dls, line 372: Should this be `if opa and await opa.admin()`? As in admin can also get tasks?

tpoliaw: This check is gating the task filtering. If the user is an admin they should be able to see all tasks, not just their own.

src/blueapi/service/main.py
tpoliaw: To avoid leaking information about tasks to people without authorization. If you get a different response depending on whether the task exists (403) vs not (404) you can check for available tasks without having authorization.

src/blueapi/service/authorization.py
ZohebShaikh: Can you add this comment ?

ZohebShaikh (approved):
I have checked that the policy is working as expected. There is just one docs update which can be added at a later point

PR #1540, opened 2026-05-15
chore: Include all ports in system compose file

tpoliaw:
The ports are not required when they're not being forwarded but
including them means there is one place to look to figure out which
service is running where.

### DIFF
--- a/tests/system_tests/compose.yaml
+++ b/tests/system_tests/compose.yaml
@@ -47,9 +47,7 @@ services:
 
   tiled:
     image: ghcr.io/bluesky/tiled:0.2.4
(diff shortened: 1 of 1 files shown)


shree-iyengar-dls (approved):
This is helpful thanks thanks

tpoliaw:
@ZohebShaikh This doesn't affect anything if the port is being mapped to itself, right?

ZohebShaikh:
This shouldn't hurt but they are not required because they are on network host.

Another option is to just add them as comments because I think you are just trying to document them ?

ZohebShaikh (approved):

PR #1541, opened 2026-05-15
feat: Add OpaClient to wrap auth checks

tpoliaw:
Proof of concept opa client with dependency injection and example check

### DIFF
--- a/helm/blueapi/values.schema.json
+++ b/helm/blueapi/values.schema.json
@@ -756,6 +756,11 @@
             "title": "OpaConfig",
             "type": "object",
             "properties": {
+                "audience": {
+                    "title": "Audience",
+                    "default": "account",
+                    "type": "string"
+                },
                 "root": {
                     "title": "Root",
                     "default": "http://localhost:8181/",
--- a/helm/blueapi/values.schema.json
+++ b/helm/blueapi/values.schema.json
@@ -751,6 +751,22 @@
             },
             "additionalProperties": false
         },
+        "OpaConfig": {
+            "$id": "OpaConfig",
+            "title": "OpaConfig",
+            "type": "object",
+            "properties": {
+                "root": {
+                    "title": "Root",
+                    "default": "http://localhost:8181/",
+                    "type": "string",
+                    "format": "uri",
+                    "maxLength": 2083,
+                    "minLength": 1
+                }
+            },
+            "additionalProperties": false
+        },
         "PlanSource": {
             "$id": "PlanSource",
             "title": "PlanSource",
@@ -1005,6 +1021,16 @@
                     "anyOf": [
                         {
                             "$ref": "OIDCConfig"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ]
+                },
+                "opa": {
+                    "anyOf": [
+                        {
+                            "$ref": "OpaConfig"
                         },
                         {
                             "type": "null"
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -298,6 +298,7 @@ class Tag(StrEnum):
 
 class OpaConfig(BlueapiBaseModel):
     root: HttpUrl = HttpUrl("http://localhost:8181")
+    audience: str = "account"
 
 
 class ApplicationConfig(BlueapiBaseModel):
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -296,6 +296,10 @@ class Tag(StrEnum):
     META = "Meta"
 
 
+class OpaConfig(BlueapiBaseModel):
+    root: HttpUrl = HttpUrl("http://localhost:8181")
+
+
 class ApplicationConfig(BlueapiBaseModel):
     """
     Config for the worker application as a whole. Root of
@@ -335,6 +339,7 @@ class ApplicationConfig(BlueapiBaseModel):
     oidc: OIDCConfig | None = None
     auth_token_path: Path | None = None
     numtracker: NumtrackerConfig | None = None
+    opa: OpaConfig | None = None
 
     def __eq__(self, other: object) -> bool:
         if isinstance(other, ApplicationConfig):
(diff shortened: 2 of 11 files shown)


src/blueapi/config.py
```diff
@@ -296,6 +296,10 @@ class Tag(StrEnum):
     META = "Meta"
 
 
+class OpaConfig(BlueapiBaseModel):
+    root: HttpUrl = HttpUrl("http://localhost:8181")
```
fajinyuan, line 300: why hard coded this URL in ? will this always the case on beamline?

helm/blueapi/values.schema.json
```diff
@@ -751,6 +751,22 @@
...
+                    "title": "Root",
+                    "default": "http://localhost:8181/",
+                    "type": "string",
+                    "format": "uri",
+                    "maxLength": 2083,
+                    "minLength": 1
+                }
+            },
+            "additionalProperties": false
+        },
```
fajinyuan, line 774: I am new to cloud programming, but would like to understand why there are the same json section here as in config_schema.json above?

fajinyuan (commented):
please see my questions.

src/blueapi/config.py
tpoliaw: The hardcoded value is a default (the opa server used in system tests) similar to the ones for numtracker and OIDC.  On beamlines, this will be overwritten by the config (in the beamline's `values.yaml`)

helm/blueapi/values.schema.json
tpoliaw: When deploying blueapi via helm, the config is included in the `values.yaml` file. When the schema of the config changes, the embedded config in the values file changes in the same way.

src/blueapi/config.py
fajinyuan: Thanks for the explaination as from the codes I did see how this URL being replaced by other on beamline, I must have missed some codes that does the overwrite.

src/blueapi/service/authorization.py
```diff
@@ -0,0 +1,49 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+class OpaClient:
+    client: aiohttp.ClientSession
+
+    def __init__(self, instrument: str, config: OpaConfig):
+        LOGGER.info("Creating OpaClient for %s with config %s", instrument, config)
+        self._instrument = instrument
+        self._conf = config
```
ZohebShaikh: nit: it is a private variable so you can leave it as it is 
```suggestion
        self._config = config
```

src/blueapi/service/authorization.py
```diff
@@ -0,0 +1,49 @@
...
+        LOGGER.info("Closing OPA session")
+        await self._session.close()
+
+    async def _call_opa(self, endpoint, data: Mapping[str, Any]) -> bool:
+        resp = await self._session.post(
+            endpoint,
+            json={
+                "input": {
+                    "beamline": self._instrument,
+                    "audience": "account",
```
ZohebShaikh: I have been going back and forth on this, but I think this should be configurable in the config rather than hardcoded. we can default it to "account"
```suggestion
                    "audience": self._audience,
```

src/blueapi/service/authorization.py
```diff
@@ -0,0 +1,49 @@
...
+        LOGGER.info("Creating OpaClient for %s with config %s", instrument, config)
+        self._instrument = instrument
+        self._conf = config
+        self._session = ClientSession(base_url=config.root.encoded_string())
+
+    async def aclose(self):
+        LOGGER.info("Closing OPA session")
+        await self._session.close()
+
+    async def _call_opa(self, endpoint, data: Mapping[str, Any]) -> bool:
```
ZohebShaikh:
```suggestion
    async def _call_opa(self, endpoint: str, data: Mapping[str, Any]) -> bool:
```

src/blueapi/service/authorization.py
```diff
@@ -0,0 +1,49 @@
...
+            endpoint,
+            json={
+                "input": {
+                    "beamline": self._instrument,
+                    "audience": "account",
+                    **data,
+                }
+            },
+        )
+        return (await resp.json())["result"]
```
ZohebShaikh, line 36: nit: Will be good to TypeAdapter the result to bool  because of python

Something like this 
```
        from pydantic import TypeAdapter
        return TypeAdapter(bool).validate_python((await resp.json())["result"])
```

I think we should as put this in 
```
try:
  	TypeAdapter(bool).validate_python((await resp.json())["result"])
except KeyError : # on result unlikely
	...
except Timeout to OPA as e:
	raise e as Timeouterror
```
or just Exception

src/blueapi/config.py
```diff
@@ -343,6 +348,7 @@ def __eq__(self, other: object) -> bool:
                 & (self.env == other.env)
                 & (self.logging == other.logging)
                 & (self.api == other.api)
+                & (self.opa == other.opa)
```
ZohebShaikh, line 352: This should be done is some other PR

```
& (self.tiled == other.tiled)
& (self.oidc == other.oidc)
& (self.numtracker == other.numtracker)
```

ZohebShaikh (commented):
looks good

src/blueapi/service/authorization.py
tpoliaw: What does the TypeAdapter protect against? OPA returning `"False"` instead of `False`? 

For the exception handling, does wrapping the exception here add much beyond letting the original exception be raised?

src/blueapi/config.py
tpoliaw: [It is handled by pydantic in the `ConfigLoader` in `config.py`](https://github.com/DiamondLightSource/blueapi/blob/main/src/blueapi/config.py#L413)

helm/blueapi/values.schema.json
fajinyuan: sounds like these schema are generated, not hand crafted, is that correct? if so why they are required to be PR'ed? 
My guess is these need to be committed into repo for deployment, is that correct?

ZohebShaikh (approved):

PR #1546, opened 2026-05-27
chore(deps): Update starlette in uv lock file

abbiemery:
Update starlette to include a patch resolving the vulnerability CVE-2026-48710. We depend on starlette via fastaspi which has it pinned on a lower version.


tpoliaw (approved):
Looks ok. Could you link/reference the CVE in the message

PR #1547, opened 2026-05-28
refactor: Move auth extractors into authentication module

tpoliaw:
Splitting up the auth extractor method allows other extractors to access the raw token as well as easy access to the validated and decoded access token.

### DIFF
--- a/src/blueapi/service/authentication.py
+++ b/src/blueapi/service/authentication.py
@@ -6,16 +6,20 @@ import threading
 import time
 import webbrowser
 from abc import ABC, abstractmethod
+from collections.abc import Mapping
 from functools import cached_property
 from http import HTTPStatus
 from pathlib import Path
-from typing import Any, cast
+from typing import Annotated, Any, cast
 
 import httpx
 import jwt
 import requests
+from fastapi import Depends, HTTPException, Request
+from fastapi.security.utils import get_authorization_scheme_param
 from pydantic import TypeAdapter
 from requests.auth import AuthBase
+from starlette.status import HTTP_401_UNAUTHORIZED
 
 from blueapi.config import OIDCConfig, ServiceAccount
 from blueapi.service.model import Cache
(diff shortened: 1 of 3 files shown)


src/blueapi/service/authentication.py
```diff
@@ -272,3 +276,52 @@ def get_access_token(self):
...
         yield request
+
+
+def unchecked_bearer_token(req: Request) -> str | None:
+    """Get bearer token value from authorization header"""
+    auth = req.headers.get("Authorization")
+    scheme, param = get_authorization_scheme_param(auth)
+    if scheme.casefold() != "bearer":
+        return None
+    return param.strip()
```
ZohebShaikh, line 291: This is a copy of this 
```
    authorization = request.headers.get("Authorization")
    scheme, param = get_authorization_scheme_param(authorization)
    if not authorization or scheme.lower() != "bearer":
        if self.auto_error:
            raise self.make_not_authenticated_error()
        else:
            return None  # pragma: nocover
    return param
```
Why not use the function already in fastAPI ?

After this PR you only use it at 1 more place.
A better thing to do would be to create oauth_scheme in main
```
oauth_scheme = OAuth2AuthorizationCodeBearer(
    authorizationUrl="",
    tokenUrl="",
    refreshUrl=None,
)
```
and do depend(oauth_scheme) to get the unchecked_access_token


To a certain extend I get that you want authN related stuff to be in the authentication.py file

ZohebShaikh (changes requested):
Generally I think why write extra code when it is already been written by someone else ?

src/blueapi/service/authentication.py
tpoliaw: > Why not use the function already in fastAPI ?

Because passing in null/placeholder values that aren't needed makes the code more confusing. It also inserts auth details into the API schema because this dependency is always on the endpoints even though is optional.

> After this PR you only use it at 1 more place

For now yes. This was paving the way for the authz changes that need access to the unvalidated access_token

> A better thing to do would be to create oauth_scheme in main

I think if it's in main we end up with circular imports. main imports authz imports main. Having all access token handling in main makes it end up doing everything. As you said, having authn stuff in the authn module makes it clearer.

ZohebShaikh: This looks like a preference in doing things a certain way so will not block the merge, but having a link to the fastapi where it is being copied from and with some explanation will be a good thing

ZohebShaikh (approved):

PR #1548, opened 2026-05-28
feat: Validate tiled ServiceAccount config at startup

tpoliaw:
Ensure that the tiled auth configuration is valid and fail fast instead of when the first data is written.

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -299,6 +299,7 @@ class Tag(StrEnum):
 class OpaConfig(BlueapiBaseModel):
     root: HttpUrl = HttpUrl("http://localhost:8181")
     audience: str = "account"
+    tiled_service_account_check: str
 
 
 class ApplicationConfig(BlueapiBaseModel):
--- a/tests/unit_tests/service/test_authorization.py
+++ b/tests/unit_tests/service/test_authorization.py
@@ -1,11 +1,13 @@
-from unittest.mock import AsyncMock, MagicMock, patch
+from contextlib import AbstractContextManager, nullcontext
+from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
 from pydantic import HttpUrl
 
-from blueapi.config import OpaConfig
+from blueapi.config import OIDCConfig, OpaConfig, ServiceAccount
 from blueapi.service.authorization import (
     OpaClient,
+    validate_tiled_config,
 )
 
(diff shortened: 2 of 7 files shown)


src/blueapi/config.py
```diff
@@ -298,6 +298,7 @@ class Tag(StrEnum):
 
 class OpaConfig(BlueapiBaseModel):
     root: HttpUrl = HttpUrl("http://localhost:8181")
+    tiled_service_account_check: str
```
ZohebShaikh, line 302: nit: This changes only when the policy changes so we can have a default value for it here

tests/unit_tests/service/test_authorization.py
```diff
@@ -20,9 +22,50 @@
...
+        (True, nullcontext()),
+    ],
+)
+async def test_tiled_service_account(
+    session: MagicMock,
+    opa_config: OpaConfig,
+    result: bool,
+    context: AbstractContextManager,
+):
+    session.return_value.post = AsyncMock(
```
ZohebShaikh, line 43: I'm writing this comment only once,But it applies to all the tests.

I personally think that [pytest-aiohttp](https://docs.aiohttp.org/en/stable/testing.html#unittest) provide a better way of do this type of test. Creating so many mocks looks unnecessary/ clutters the code.

tpoliaw: I'm not sure how pytest-aiohttp would be used here? It appears to be targetted at testing webservers not mocking out responses.

src/blueapi/config.py
tpoliaw: It depends if we want to include diamond specific configuration as defaults? As far as I can see so far, there is nothing included so far. Would the services copier template be a better place to put the default values?

tests/unit_tests/service/test_authorization.py
ZohebShaikh: I briefly tried rewriting the tests with aiohttp, Can put it latter as a PR to continue this conversation

ZohebShaikh (approved):

src/blueapi/config.py
ZohebShaikh: In some places we do write defaults which are diamond specific some places we don't as there is no consistency. Happy for this to go into services template default

ZohebShaikh (approved):

PR #1549, opened 2026-05-28
feat: Add submit task permission checks

tpoliaw:
Adds methods for calling OPA service to check if a user can submit a task


tpoliaw:
Merging this into the PR where the checks are being used.

Issue #1381, opened 2026-02-06
Refactor CLI login process


tpoliaw:
The `login` subcommand and the supporting methods are spread across three modules (cli, client and rest). The configuration for them is split between the server (for the oidc config) and the config file (which we want the client to avoid needing (#1327)).

* To login, a `SessionManager` is created using a path from the config and queried for a token
* If it does not find a valid token, a client is created from the config (creating a new session manager from the same path)
* The client is used to get the oidc_config (via the rest client)
* The oidc config is used to create a third `SessionManager`
* The final `SessionManager` is used to start the device flow and cache the token

Without having fully worked through the repercussions, it would be good to:
* Remove the token file path from the config - we don't want to need a config file for the client
  * Make the token path a CLI option if it still needs to be configurable
* Move the login method into the rest client - if we need the rest client to access the oidc config anyway, it makes sense to use it to log in.
  * `interactive_login()` - to prompt the user to login
  * `check_login()` - to fail early for scripts

PR #1553, opened 2026-06-10
refactor: allow any token retriever in the rest client

jacob-williamson:
See https://confluence.diamond.ac.uk/pages/viewpage.action?pageId=398853592&spaceKey=DCT&title=04-06-2026
Currently the rest client requires a session manager to handle retrieving auth tokens. I've made it more flexible so that it now accepts anything with a `get_valid_access_token` method, so a UDC client can use the same rest client.


DominicOram (changes requested):
Thanks, I think we want `SessionManager` to explicitly implement `TokenRetriever` though, right?

I also think we should add the `UDCTokenRetriever` (from https://github.com/DiamondLightSource/daq-queuing-service/pull/40/changes) into BlueAPI proper, it seems like a reasonable thing to have in here. Though maybe @tpoliaw disagrees?

src/blueapi/service/authentication.py
```diff
@@ -238,7 +242,7 @@ def start_device_flow(self):
 
 
 class JWTAuth(AuthBase):
-    def __init__(self, session_manager: SessionManager | None):
+    def __init__(self, session_manager: TokenRetriever | None):
```
tpoliaw:
```suggestion
    def __init__(self, token_source: TokenRetriever | None):
```

src/blueapi/client/rest.py
```diff
@@ -157,16 +157,16 @@ def _response_json(response: requests.Response) -> Any:
 
 class BlueapiRestClient:
     _config: RestConfig
-    _session_manager: SessionManager | None
+    _token_retreiver: TokenRetriever | None
```
tpoliaw:
```suggestion
    _token_retriever: TokenRetriever | None
```

tpoliaw (commented):
Thanks - this makes things simpler and should help when we come to refactor the login stuff in #1381.

Possibly a bit picky but can the protocol be called `TokenSource` or `TokenProvider` or similar. It's not really retrieving it if it's generating a new one and blueapi shouldn't care where it comes from.

>I think we want SessionManager to explicitly implement TokenRetriever though, right?

It can do but I don't think it needs to.

>I also think we should add the UDCTokenRetriever into BlueAPI proper

I'm not against the UDC ServiceAccount provider being in blueapi core but it doesn't need to be in this PR.

jacob-williamson:
I've renamed to `TokenSource` and made `SessionManager` explicitly inherit from it. Can move the UDC token source over in a later PR

PR #1561, opened 2026-06-24
test: Add test for app version in container

tpoliaw:
Ensure that the version number of the app in the container matches the version we expect.

### DIFF
--- a/.github/workflows/_container.yml
+++ b/.github/workflows/_container.yml
@@ -52,9 +52,6 @@ jobs:
           load: true
           tags: tag_for_testing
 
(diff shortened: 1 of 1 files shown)


tpoliaw:
From previous PR https://github.com/DiamondLightSource/blueapi/pull/1528#issuecomment-4555560232
> I don't understand why the version number is being mangled by the setuptools-scm. 1.14.1-dev.1 is becoming 1.14.1.dev1 and I haven't figured out why yet.

abbiemery (dismissed):

PR #1562, opened 2026-06-24
chore: Use Ubuntu 26.04 (resolute) for devcontainer

tpoliaw:
Also includes additional tools:
* gh
* glab
* just
* lazygit
* nodejs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # The devcontainer should use the developer target and run as root with podman
 # or docker with user namespaces.
-FROM ghcr.io/diamondlightsource/ubuntu-devcontainer:noble AS developer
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # The devcontainer should use the developer target and run as root with podman
 # or docker with user namespaces.
-FROM ghcr.io/diamondlightsource/ubuntu-devcontainer:noble AS developer
+FROM ghcr.io/diamondlightsource/ubuntu-devcontainer:resolute AS developer
```
ZohebShaikh, line 4: This should come from copier template

tpoliaw: Sure, I'll open a PR there instead

PR #1565, opened 2026-06-25
fix: add missing values ApplicationConfig equality check

ZohebShaikh:
Currently the following configs are not checked for equality

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -338,6 +338,9 @@ class ApplicationConfig(BlueapiBaseModel):
                 & (self.logging == other.logging)
                 & (self.api == other.api)
                 & (self.opa == other.opa)
+                & (self.tiled == other.tiled)
+                & (self.oidc == other.oidc)
+                & (self.numtracker == other.numtracker)
             )
         return False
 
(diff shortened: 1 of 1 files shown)


src/blueapi/config.py
```diff
@@ -338,6 +338,9 @@ def __eq__(self, other: object) -> bool:
                 & (self.logging == other.logging)
                 & (self.api == other.api)
                 & (self.opa == other.opa)
+                & (self.tiled == other.tiled)
```
tpoliaw: Any chance these could be added in the same order as the fields are defined? Are `scratch` and `auth_token_path` intentionally ignored?

src/blueapi/config.py
tpoliaw, line 334: While we're updating this method, this should probably be `type(other) == type(self)` so that equality doesn't break for subclasses.

ZohebShaikh: I can add this but ApplicationConfig has no subclass

tpoliaw: True, probably not worth it. But by the same argument, are we ever checking for equality?

tpoliaw (approved):

PR #1566, opened 2026-06-25
feat: run unit tests in parallel

ZohebShaikh:
As we have a 4 core GitHub action run we can utilise it more efficiently by running all our unit test in parallel.

Currently it take the following amount of time to run unit test

-  python3.11: 2m       vs now  59s
- python3.12: 2m 1s   vs now  1m 28s
- python3.13: 1m 44s vs now  1m 12s

It is consistently less in the future it will be a nice addition to run system test in parallel as well. It will require some more work so will do it in another PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ filterwarnings = ["error", "ignore::DeprecationWarning"]
 # Doctest python code in docs, python code in src docstrings, test functions in tests
 testpaths = "docs src tests"
 asyncio_mode = "auto"
-timeout = 30
+timeout = 3
 
 [tool.coverage.run]
 patch = ["subprocess"]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -102,7 +103,7 @@ filterwarnings = ["error", "ignore::DeprecationWarning"]
 # Doctest python code in docs, python code in src docstrings, test functions in tests
 testpaths = "docs src tests"
 asyncio_mode = "auto"
-timeout = 3
+timeout = 10
```
ZohebShaikh, line 106: I think when tests are running in a different core sometimes it takes longer for the same test to complete which needs this extra timeout.
Might need to look if this can be optimized any further

tpoliaw (approved):
Tests now take <10s on my machine :exploding_head:. It is slightly quicker to use `-n 8` instead of `logical` as the time to set up the workers starts taking longer than the tests but as a default it makes sense (and is still faster than running single threaded).

> in the future it will be a nice addition to run system test in parallel as well

I think that will be difficult as many of the calls to the server can't be run if another task is running or if the state is not known.

Issue #1567, opened 2026-06-25
Calling `login` on a BlueapiClient doesn't allow that client to be used


tpoliaw:
Run `blueapi logout` to clear any cached tokens.

In a REPL create a new client and login

```python
>>> from blueapi.client import BlueapiClient
>>> bc = BlueapiClient.from_config_file("tests/system_tests/config.yaml")
>>> bc.login()
Logging in
Please login from this URL:- http://localhost:8081/realms/master/device?user_code=ABCD-EFGH
Logged in and cached new token
>>> bc.plans
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/qan22331/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/scratch/athena/blueapi/.venv/lib/python3.11/site-packages/observability_utils/tracing/decorators.py", line 151, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/athena/blueapi/src/blueapi/client/client.py", line 260, in plans
    return PlanCache(self, self._rest.get_plans().plans)
                           ^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/athena/blueapi/src/blueapi/client/rest.py", line 202, in get_plans
    return self._request_and_deserialize("/plans", PlanResponse)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/athena/blueapi/.venv/lib/python3.11/site-packages/observability_utils/tracing/decorators.py", line 151, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/athena/blueapi/src/blueapi/client/rest.py", line 324, in _request_and_deserialize
    raise exception
blueapi.client.rest.UnauthorisedAccessError: (401, '{"detail":"Not authenticated"}')
```
Creating a new client now works as expected so the login did work to some degree
```python
>>> bc = BlueapiClient.from_config_file("tests/system_tests/config.yaml")
>>> bc.plans
PlanCache(8 plans)
```

## Acceptance Criteria
- `bc.login()` logs in and lets the client be used

PR #1568, opened 2026-06-25
chore: Add just task to launch repl with blueapi client

tpoliaw:
$ just repl

will now launch a repl (using ptpython) with an initialized
BlueapiClient named `bc` using the config from the system tests (the
same used by `just serve` to run the server).

### DIFF
--- a/justfile
+++ b/justfile
@@ -34,3 +34,11 @@ system *OPTS:
 coverage:
     uv run pytest tests/unit_tests --cov --cov-report html
     xdg-open htmlcov/index.html
+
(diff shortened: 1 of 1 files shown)


abbiemery (commented):
Is there a way to include the login step with this?

tpoliaw:
Yes but it currently wouldn't work: https://github.com/DiamondLightSource/blueapi/issues/1567

abbiemery (approved):
It's useful enough as is, can add login in the future if we want to

PR #1573, opened 2026-06-26
fix: Remove dodal from default application config

abbiemery:
The default application still included dodal plans as a default. With the change to remove this dependency from the default application this would cause the server to error when trying to fetch plans. This replaces those defaults with an empty list.

### DIFF
--- a/src/blueapi/config.py
+++ b/src/blueapi/config.py
@@ -140,10 +140,7 @@ class EnvironmentConfig(BlueapiBaseModel):
             PlanSource | DeviceManagerSource,
             Field(discriminator="kind"),
         ]
-    ] = [
-        PlanSource(module="dodal.plans"),
-        PlanSource(module="dodal.plan_stubs.wrapped"),
-    ]
+    ] = Field(default=[])
     events: WorkerEventConfig = Field(default_factory=WorkerEventConfig)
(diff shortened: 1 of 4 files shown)


src/blueapi/config.py
```diff
@@ -140,10 +140,7 @@ class EnvironmentConfig(BlueapiBaseModel):
             PlanSource | DeviceManagerSource,
             Field(discriminator="kind"),
         ]
-    ] = [
-        PlanSource(module="dodal.plans"),
-        PlanSource(module="dodal.plan_stubs.wrapped"),
-    ]
+    ] = Field(default=[])
```
tpoliaw, line 143: I think you can use `[]` on its own if you don't mind the [pydantic magic](https://pydantic.dev/docs/validation/latest/concepts/fields/#mutable-default-values)

tpoliaw (approved):

src/blueapi/config.py
abbiemery: I know, I just thought i'd be consistent with ~the rest of the~ the adjacent config

abbiemery: I don't have an opinion either way so will change it if you prefer

tpoliaw: nope, keeping it consistent makes sense

PR #1576, opened 2026-06-26
feat: Reduce logging noise

### LINKED ISSUE #1560: Reduce logging of http requests
When running, BlueAPI logs for every time it gets a HTTP request, this is probably fine for `POST` requests because we want to know when it's been triggered and with what but it also logs `GET` e.g. `10.42.0.1:39058 GET /worker/state 200`. This makes the logs very noisy when you have a UI open that's polling for information. I'm also not sure it generally gives much value.

## Acceptance Criteria
- Logging of `GET` requests is removed/reduced

dan-fernandes:
Now only logs GET at debug level, and all else at INFO.

For GET requests, only logs after processing request. All else log both before and after processing request.

### DIFF
--- a/src/blueapi/service/main.py
+++ b/src/blueapi/service/main.py
@@ -594,9 +594,15 @@ def start(config: ApplicationConfig):
 async def log_request_details(
     request: Request, call_next: Callable[[Request], Awaitable[StreamingResponse]]
 ) -> Response:
-    """Middleware to log all request's host, method, path, status and request and
-    body"""
-    log = LOGGER.debug if request.url.path == "/healthz" else LOGGER.info
+    """Middleware to log request's host, method, path, status and request and
+    body
+
+    Logs POST and PUT as info, all else as debug."""
+    log = (
+        LOGGER.info
+        if request.method == "POST" or request.method == "PUT"
+        else LOGGER.debug
+    )
     request_body = await request.body()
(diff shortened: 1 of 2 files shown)


src/blueapi/service/main.py
```diff
@@ -594,16 +594,23 @@ def start(config: ApplicationConfig):
...
-    """Middleware to log all request's host, method, path, status and request and
-    body"""
-    log = LOGGER.debug if request.url.path == "/healthz" else LOGGER.info
+    """Middleware to log request's host, method, path, status and request and
+    body
+
+    Logs POST and PUT as info, all else as debug."""
+    log = (
+        LOGGER.info
+        if request.method == "POST" or request.method == "PUT"
```
tpoliaw: would it be simpler to special case `GET` instead of both `POST` and `PUT`? `DELETE` at info could potentially be useful

tpoliaw (approved):

src/blueapi/service/main.py
dan-fernandes: Yes - good point

tpoliaw (approved):

src/blueapi/service/main.py
tpoliaw: PR summary needs updating before merging - the queue uses it for the commit message

PR #1564, opened 2026-06-25
docs: restructure how-to and tutorial pages

shree-iyengar-dls:
Fixes #1468 and #1416 and #1402

Summary of changes made:
Tutorials section - walk through for new users to run blueapi server, add plans and devices to the config file and run a plan
How-to section - more details that can be added to config file and table of dev command tools

PR #1577, opened 2026-06-26
docs: Add simple devices and plans for use in tutorials

abbiemery:
With the removal of dodal as a dependency, we need to provide some simple devices and plans in order to follow through the tutorials. These are getting overhauled in https://github.com/DiamondLightSource/blueapi/pull/1564.

This adds ophyd-async sim devices and dodal plans to allow users to run simple plans and make a hdf5 file.


src/blueapi/tutorial/plans.py
abbiemery, line 1: This is just a direct copy of the wrapped count in dodal. I don't know if it is worth while paring this down to something more simple, or to add a move plan or something.

src/blueapi/core/protocols.py
abbiemery, line 1: Is there a better place for this to live?

tpoliaw: utils would make the most sense if there's a way of getting it working without import order issues

src/blueapi/tutorial/plans.py
tpoliaw: I don't think we should have a direct copy - paring it down to the basic would make it clearer, eg the validation checks are not useful for tutorial purposes.

src/blueapi/core/protocols.py
abbiemery: utils but not include it in the `__all__` ?

abbiemery:
After a lot of rabbit holes, I went back to using optional dependencies for the demo. I don't think the config should live in that module though, I will put it just in the docs.

fajinyuan (approved):
Looks good tome.

tpoliaw (approved):

PR #1578, opened 2026-06-30
fix: get tasks endpoint to a enum

ZohebShaikh:
Currently get_tasks takes a string to filter on task_status, but it should be a enum.

### DIFF
--- a/src/blueapi/service/interface.py
+++ b/src/blueapi/service/interface.py
@@ -233,11 +233,6 @@ def begin_task(
     return task
 
 
-def get_tasks_by_status(status: TaskStatusEnum) -> list[TrackableTask]:
-    """Retrieve a list of tasks based on their status."""
-    return worker().get_tasks_by_status(status)
-
-
 def get_active_task() -> TrackableTask | None:
     """Task the worker is currently running"""
     return worker().get_active_task()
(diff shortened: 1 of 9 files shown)


src/blueapi/service/main.py
tpoliaw: I think if we accept an enum, fastapi will do this validation for us so the block can be removed

ZohebShaikh: This is true, While I was going through the code I thought I will remove the get_tasks function as well and simplify it a little bit.

src/blueapi/service/interface.py
```diff
@@ -264,10 +259,10 @@ def cancel_active_task(failure: bool, reason: str | None) -> str:
     return worker().cancel_active_task(failure, reason)
 
 
-def get_tasks() -> list[TrackableTask]:
-    """Return a list of all tasks on the worker,
-    any one of which can be triggered with begin_task"""
-    return worker().get_tasks()
+def get_tasks_by_status(status: TaskStatusEnum | None = None) -> list[TrackableTask]:
```
tpoliaw: I think I'd be inclined to leave this called `get_tasks`. 

`get_tasks(COMPLETE)` with a filter makes more sense without context than `get_tasks_by_status()` with no status.

tpoliaw (approved):
Will leave name decision to you. I think this is an improvement already

PR #1579, opened 2026-06-30
chore(deps): update uv lock and python version

ZohebShaikh:
Currently there are [40](https://github.com/DiamondLightSource/blueapi/security/dependabot) dependency security alert.

[python-copier#issue-339](https://github.com/DiamondLightSource/python-copier-template/issues/339) should be prioritised to enable renovate.

Changes in the PR:
1. Change python-version to `3.14`
2. Update all `uv.lock` file
3. Add `httpx2` to dev dependences for [Deprecation Warning](https://github.com/Kludex/starlette/blob/main/starlette/testclient.py#L46)
4. Upgrade to latest ophyd-async Which has the changed to use [ADWriterFactory](https://blueskyproject.io/ophyd-async/main/explanations/decisions/0016-ad-writer-factory.html)
5. Change HTTP_422_UNPROCESSABLE_ENTITY to [HTTP_422_UNPROCESSABLE_CONTENT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422)  as the previous is deprecated


Alexj9837 (approved):
Looks good, spoke earlier about the need to keep HTTPX in due to tiled and other services use it.

PR #1581, opened 2026-06-30
feat: Log error details when task parameters are invalid

### LINKED ISSUE #1580: Better logging on 422 errors
When running plans on i15-1 we're occasionally getting 422 errors in blueAPI:

```
2026-06-30 09:18:16,779     INFO bluesky.RE.state Change state on <bluesky.run_engine.RunEngine object at 0x7fe173e31990> from 'running' -> 'idle'
2026-06-30 09:18:16,780     INFO bluesky Cleaned up from plan <generator object sleep at 0x7fe0c2271700>
2026-06-30 09:18:16,780     INFO blueapi.worker.task_worker Task ran successfully - returned: None
2026-06-30 09:18:16,781     INFO blueapi.worker.task_worker Awaiting task
2026-06-30 09:18:17,885     INFO blueapi.service.main ::1:48356 GET /worker/state 200
2026-06-30 09:18:19,063     INFO blueapi.service.main ::1:48362 GET /worker/state 200
2026-06-30 09:18:20,245     INFO blueapi.service.main ::1:48370 GET /worker/state 200
2026-06-30 09:18:20,554     INFO blueapi.service.main ::1:48382 POST /tasks 422
2026-06-30 09:18:21,426     INFO blueapi.service.main ::1:48384 GET /worker/state 200
2026-06-30 09:18:22,611     INFO blueapi.service.main ::1:48390 GET /worker/state 200
2026-06-30 09:18:23,794     INFO blueapi.service.main ::1:48392 GET /worker/state 200
```
To diagnose these it would be good to have some logging as to what was POSTed and why it failed.

This can be reproduced locally by running the UI against a local BlueAPI and sending a selection of different plans.

## Acceptance Criteria
- On 422 errors there is better logging

tpoliaw:


src/blueapi/service/main.py
```diff
@@ -296,6 +296,7 @@ def submit_task(
             for err in e.errors()
         ]
 
+        LOGGER.info("Error submitting task: %s - %s", task_request, e)
```
DominicOram, line 299: Should: `info` seems low. I would say at least warn, maybe even error.

DominicOram (approved):
Awesome, thank you for the quick fix! The message that is logged is good, just disagree about the level. Will approve assuming that you increase the level.

src/blueapi/service/main.py
tpoliaw: It's not really an error from blueapi's perspective if the user provided invalid input. Happy to see what others think though

ZohebShaikh: It is a client error but a server info so I agree with Peter that it should only be logged as info on server side.

This is similar to how fastapi handles a request which has correct syntax but wrong instruction, and we are sending the correct response for it [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/422) same as fastapi

DominicOram: Sure, I think it will make support slightly easier if it's `error` but I'm not that fussed as long as the information is there

abbiemery: merge it for now, but if it proves to be frustrating on the user side we can change it

PR #1583, opened 2026-06-30
chore: update pre-commit hooks

ZohebShaikh:
1. Decouple `pyright` version from `uv.lock` file
2. Update asyncapi check file to a more maintained project
3. Update all pre-commit hooks
4. Use docker images instead of raw builds to make it easier to work on dls workstations

### DIFF
--- a/.github/workflows/asyncapi.yml
+++ b/.github/workflows/asyncapi.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - name: check asyncapi.yaml file
         uses: asyncapi/cli@fb73ff3a429c1bc6a14c1b89f12601288677b9fa # v6.0.2
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/asyncapi.yml
```diff
@@ -19,6 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```
shree-iyengar-dls: Nit: could update actions/checkout to v7? https://github.com/actions/checkout/commit/9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0

shree-iyengar-dls (commented):
Could update actions/checkout version

prek.toml
```diff
@@ -13,32 +13,32 @@ hooks = [
...
+rev = "v0.15.20"
 hooks = [
   { id = "ruff-check", args = ["--fix"] },
   { id = "ruff-format" },
 ]
 
 [[repos]]
 repo = "https://github.com/astral-sh/uv-pre-commit"
-rev = "0.10.2"
+rev = "0.11.26"
```
shree-iyengar-dls: Nit: also maybe 0.11.27? https://github.com/astral-sh/uv-pre-commit

shree-iyengar-dls (commented):
One more potential version update

.github/workflows/asyncapi.yml
ZohebShaikh: There are lots of other checkout versions in our CI but they come from copier template so it should be updated there

shree-iyengar-dls (approved):

PR #1587, opened 2026-07-08
chore: revert to build pre-commit

ZohebShaikh:
Currently docker builds work only outside of the dev container in dls workstation. So reverting this change

```
error: Failed to run hook `helm-docs-container`
  caused by: Failed to run `docker run --rm --tty --user 0:0 --init --volume /workspaces/blueapi:/src:rw,Z --workdir /src jnorwood/helm-docs:latest --chart-search-root=helm`
  caused by: No such file or directory (os error 2)
error: Recipe `lint` failed on line 25 with exit code 2
```

### DIFF
--- a/.github/workflows/_system_test.yml
+++ b/.github/workflows/_system_test.yml
@@ -22,7 +22,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
 
       - name: Compose services
-        uses: hoverkraft-tech/compose-action@11beaa1c2dae4e8ed7b1665aa074723b6cecb0e4 # v3.0.0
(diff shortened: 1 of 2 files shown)


Alexj9837 (approved):
looks good. seems to be rolling back changes you made in https://github.com/DiamondLightSource/blueapi/pull/1583/changes#diff-72359463e1c50ddcb8dfabe900898a8f1d92416bbe6754f2e497f62d96232021

ZohebShaikh:
I had a look at [docker-in-docker](https://github.com/devcontainers/features/tree/main/src/docker-in-docker) which can be useful for this case but it looks fragile because we use docker and it might not work in the dls-workstations

PR #1594, opened 2026-07-14
chore: remove super-state-machine dependency

### LINKED ISSUE #555: Remove super-state-machine as direct dependency
This was added in #554 for #553.

Once https://github.com/bluesky/bluesky/issues/1760 has been resolved it should be removed.

ZohebShaikh:
closes #555 

Using _vendor because we are only using it for typing and errors


Alexj9837 (approved):
looks good, one for one change for TransitionError.

PR #1596, opened 2026-07-15
test: fix pytest-xdist and pytest-timeout working together

ZohebShaikh:
When I had done the PR for pyxdist I increased the timeout to 10 seconds as they where not playing nice with eachother.

I think PR I have gone through the docs/issues and found out that we can use the commands as in PR to make them work nicely with each other

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,11 +103,11 @@ addopts = '--tb=native -vv --doctest-modules --doctest-glob="*.rst"'
 # https://iscinumpy.gitlab.io/post/bound-version-constraints/#watch-for-warnings
 filterwarnings = ["error", "ignore::DeprecationWarning"]
 # Doctest python code in docs, python code in src docstrings, test functions in tests
-timeout_method = "thread"
-timeout_func_only = true
 testpaths = "docs src tests"
 asyncio_mode = "auto"
 timeout = 1.5
+timeout_method = "thread"
+timeout_func_only = true
 [tool.coverage.run]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -103,10 +103,11 @@ addopts = '--tb=native -vv --doctest-modules --doctest-glob="*.rst"'
 # https://iscinumpy.gitlab.io/post/bound-version-constraints/#watch-for-warnings
 filterwarnings = ["error", "ignore::DeprecationWarning"]
 # Doctest python code in docs, python code in src docstrings, test functions in tests
```
tpoliaw, line 107: This comment belongs to the line below the ones you've added. Might be useful to add a comment for why these lines are needed as well. It's not obvious what they do.

pyproject.toml
```diff
@@ -105,8 +105,11 @@ filterwarnings = ["error", "ignore::DeprecationWarning"]
...
 testpaths = "docs src tests"
 asyncio_mode = "auto"
+# Timeout settings for pytest-timeout 
 timeout = 10
```
ZohebShaikh, line 111: I had a look at ophyd-async they are keeping timeout at 60 [seconds](https://github.com/bluesky/ophyd-async/blob/6e7ac2166eae34c6d8944d3ce5d3b7e571d25204/pyproject.toml#L113)
So I think I kept this timeout can stay at 10 seconds

tpoliaw (approved):

PR #1569, opened 2026-06-25
fix: Update rest session manager when logging in

tpoliaw:
If the user is not logged in when the client is created, the rest client
does not get a session manager. When the user calls login, a new session
manager is created to cache the new token but if the rest client is not
updated, any subsequent calls remain unauthenticated.

Setting the rest client's session manager to the new instance when a
login is successful means the tokens can be used without recreating the
client.

Now that the login method works as expected, the client created in the
`just repl` task can now automatically refresh its login if required.

Also adds a `logout` method to the client so users can log out from the
repl as well.

PR #1611, opened 2026-08-05
Add a logout method to blueapi client

noemifrisina:
For python applications using the BlueapiClient directly, currently there is a way to login but not one to logout available.

### DIFF
--- a/src/blueapi/client/client.py
+++ b/src/blueapi/client/client.py
@@ -774,6 +774,17 @@ class BlueapiClient:
             else:
                 print("Server is not configured to use authentication!")
 
+    def logout(self, token_path: Path | None = None):
+        try:
+            auth: SessionManager = SessionManager.from_cache(token_path)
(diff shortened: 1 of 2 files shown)


ZohebShaikh:
@tpoliaw I think this would be useful, Was there a reason it was left out during the initial go ?

tpoliaw:
I agree that having a `logout` method makes sense but it's a bit odd that it accepts a token path that could potentially logout an unrelated session. Could it call logout on the currently active `SessionManager` instead? #1569 made the session manager accessible from outside the rest client so that could be used here as well. Something like

```python
if sm := self._rest.session_manager:
    sm.logout()
```

ZohebShaikh:
Created [PR](https://github.com/DiamondLightSource/blueapi/pull/1630) for logout

tpoliaw:
Merged into #1569

PR #1604, opened 2026-07-29
ci: Change blueapi github pages site publishing

shree-iyengar-dls:
Using the [myst-version-switcher-plugin](https://diamondlightsource.github.io/myst-version-switcher-plugin/main/how-to/use-with-sphinx/), the changes in this PR aim to address the issues surrounding the GitHub pages site being built through the `gh-pages` branch (see [this issue](https://github.com/DiamondLightSource/python-copier-template/issues/344) in the python copier template). Through this change, there is no `gh-pages` branch, instead the github action [deploy-pages](https://github.com/actions/deploy-pages) publishes one artifact as the entire site.

PR #1614, opened 2026-08-06
ci: publish docs only on tags

ZohebShaikh:
Currently we are publishing GitHub page on every branch. 
We should be only building docs on each branch and publish only on tags

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
   publish:
     needs: [docs]
     # don't publish a pages site in the fork's org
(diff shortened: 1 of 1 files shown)


shree-iyengar-dls:
Looks fine, would you be able to add the validation you mentioned from my [PR](https://github.com/DiamondLightSource/blueapi/pull/1604)? (I merged it since it was approved but didn't add the check)

ZohebShaikh:
I think it could be added, but it should be done in a separate PR.

tpoliaw (approved):

PR #1620, opened 2026-08-11
ci: change fetch depth for system_tests

ZohebShaikh:
It is not required for system test to fetch the entire history


tpoliaw:
> Need this to get version number from last tag

Is this no longer true?

tpoliaw (approved):
Tests pass so it can't be too broken

ZohebShaikh:
>Is this no longer true?

It was never actually required in the first place.

PR #1621, opened 2026-08-11
refactor: Add defaults for authZ policy

ZohebShaikh:
It feels slightly awkward to have to specify the policy values explicitly. I think having sensible defaults would provide a better UX.

They could also be included in Copier, but having them as defaults would mean one less thing to worry about when running `copier update`.

### DIFF
--- a/helm/blueapi/config_schema.json
+++ b/helm/blueapi/config_schema.json
@@ -301,23 +301,21 @@
                     "type": "string"
                 },
                 "tiled_service_account_check": {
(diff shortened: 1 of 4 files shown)


ZohebShaikh:
@shree-iyengar-dls If this change gets merged in Then the only docs that should be added is that please point it to the authorization provider which is opa.diamond.ac.uk

ZohebShaikh:
@DiamondLightSource/developers-daq-core Can anyone in our team have a look at this PR?

tpoliaw (approved):

PR #1625, opened 2026-08-13
fix: well_known_url field deprecation warning

jacob-williamson:
I've been using the `OIDCConfig` class in the queue service, and ran into a bug where my tests were failing because of a deprecation warning.

```python
def _well_known_url(self) -> str:
        if self.issuer:
            if self.well_known_url:
                LOGGER.warning(
                    DeprecationWarning(
                        "well_known_url and issuer are both set. "
                        "Defaulting to issuer URL"
                    ),
                )
            return self.issuer + "/.well-known/openid-configuration"
        return cast(str, self.well_known_url)
```
The issue is that even if you don't use the deprecated field (`well_known_url`), the model itself accesses it when it does `if self.well_known_url:`. This causes pydantic to emit a warning which causes tests to fail. I've replaced this with a check that doesn't access the actual field which fixes the issue.

Since pydantic also emits its own warning, we could probably get rid of the check all together.


src/blueapi/config.py
tpoliaw: I think you're right - if pydantic intercepts the field access to raise a deprecation warning that is probably enough and we can remove this entire check

src/blueapi/config.py
```diff
@@ -238,7 +238,7 @@ class OIDCConfig(BlueapiBaseModel):
     def check_urls(self) -> Self:
         if self.issuer is None and self.well_known_url is None:
             raise ValueError("Please provide 'OIDCConfig.issuer'")
-        if self.well_known_url:
+        if "well_known_url" in self.model_fields_set:
             LOGGER.warning(
```
tpoliaw, line 242: I think this should either raise a warning or log a message. Logging a warning is not really doing either how they're intended.

src/blueapi/config.py
jacob-williamson: From looking at the stdio output in the tests, I now think when using the cli you don't see the pydantic warning, so may still be good to have the check

src/blueapi/config.py
jacob-williamson: Makes sense, I've made it a log message

ZohebShaikh (approved):

PR #1628, opened 2026-08-17
feat: add blueapi_task_id to metadata

### LINKED ISSUE #1554: Add task_id into run metadata
There is currently no way to correlate the task that was run to the runs that were recorded in tiled. The task_id in blueapi is probably the easiest way of doing this.

ZohebShaikh:
closes #1554

### DIFF
--- a/src/blueapi/worker/task_worker.py
+++ b/src/blueapi/worker/task_worker.py
@@ -273,6 +273,7 @@ class TaskWorker:
         """
         task.prepare_params(self._ctx)  # Will raise if parameters are invalid
         task_id: str = str(uuid.uuid4())
+        task.metadata["blueapi_task_id"] = task_id
         add_span_attributes({"TaskId": task_id})
         request_id = get_baggage("correlation_id")
         # If request id is not a string, we do not pass it into a TrackableTask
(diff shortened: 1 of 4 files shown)


src/blueapi/worker/task_worker.py
```diff
@@ -271,8 +271,12 @@ def submit_task(self, task: Task) -> str:
         Returns:
             str: A unique ID to refer to this task
         """
-        task.prepare_params(self._ctx)  # Will raise if parameters are invalid
         task_id: str = str(uuid.uuid4())
+        if self._ctx.tiled_conf is not None:
```
ZohebShaikh: There 2 questions that came to my mind

1. I don't know if it is going to be useful to have blueapi_task_id regardless if you are using tiled.
2. The code can be just
```
task.metadata["blueapi_task_id"] = task_id #type: ignore
```
I have done the dance only to not put the type ignore

tpoliaw: I think we should keep the metadata the same regardless of whether tiled is being used.

I wouldn't be against making the field `dict`. There's no real benefit to keeping it `Mapping` when it's only our code using it and we're always using dicts.

Was the prepare_params intentionally moved below the task id handling? Might be better to keep failing fast if the parameters are wrong.

Alexj9837: I'm in agreement that one extra bit of meta won't be an issue and it keeps all the data align in a way.

tpoliaw (approved):

PR #1636, opened 2026-08-20
test: Use pytest's tmp_path fixture instead of our own

tpoliaw:


tpoliaw:
This was initially done as part of a hunt for what was leaving hundreds of empty temp directories (turned out to be https://github.com/bluesky/ophyd/pull/1288) but figured it was worth leaving in.

Alexj9837 (approved):
lgtm

PR #1640, opened 2026-08-21
ci: Configure keycloak using python script

ZohebShaikh:
This has lead to a speed up by 37 seconds from (old)3m 22s - (new)2m 45s and made the script more maintainable.


Alexj9837 (approved):
Looks good , my only question was about the ignore in pyproject but you explain it!

PR #1641, opened 2026-08-24
ci: Configure numtracker as part of compose

ZohebShaikh:
Configure numtracker as part of compose services

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,14 +4,6 @@ RUNNER := "docker"
 compose +ARGS="up -d":
     {{ RUNNER }} compose -f tests/system_tests/compose.yaml {{ARGS}}
 
-configure-adsim: (compose "exec" "numtracker" "/app/numtracker" "client" "configure" "adsim"
-        "--directory" '/tmp/'
-        "--scan" '{instrument}-{scan_number}'
-        "--detector" '{instrument}-{scan_number}-{detector}'
-        "--number" "43")
-
-services: compose configure-adsim
-
 serve *OPTS:
     #!/usr/bin/env bash
     source tests/system_tests/.env
--- a/tests/system_tests/services/startup.py
+++ b/tests/system_tests/services/startup.py
@@ -19,6 +19,7 @@ SERVER = os.environ.get("KEYCLOAK_SERVER", "http://localhost:8081")
 REALM = os.environ.get("KEYCLOAK_REALM", "master")
 ADMIN_USERNAME = os.environ.get("KC_BOOTSTRAP_ADMIN_USERNAME", "admin")
 ADMIN_PASSWORD = os.environ.get("KC_BOOTSTRAP_ADMIN_PASSWORD", "admin")
+NT_URL = os.environ.get("NT_URL", "http://localhost:8406/graphql")
 
 USERS = {"alice": "alice", "bob": "bob"}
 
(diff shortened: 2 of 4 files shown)


justfile
```diff
@@ -4,14 +4,6 @@ RUNNER := "docker"
 compose +ARGS="up -d":
     {{ RUNNER }} compose -f tests/system_tests/compose.yaml {{ARGS}}
 
-configure-adsim: (compose "exec" "numtracker" "/app/numtracker" "client" "configure" "adsim"
-        "--directory" '/tmp/'
-        "--scan" '{instrument}-{scan_number}'
-        "--detector" '{instrument}-{scan_number}-{detector}'
-        "--number" "43")
-
-services: compose configure-adsim
```
tpoliaw, line 13: This is going to play havoc with muscle memory :/. Are there any docs that mention `just services` that will need updating?

tests/system_tests/services/startup.py
```diff
@@ -254,7 +260,44 @@ def create_clients() -> None:
...
+    nt_url = os.environ.get("NT_URL", "http://localhost:8406/graphql")
+    response = requests.post(
+        str(nt_url),
+        json={
+            "query": """mutation {
+              configure(instrument: "adsim",
+                        config: {directory: "/tmp/",
+                            scan: "{instrument}-{scan_number}",
+                            detector: "{instrument}-{scan_number}-{detector}",
+                            scanNumber: 43}) {
```
tpoliaw:
```suggestion
                        config: {directory: "/tmp/",
                        scan: "{instrument}-{scan_number}",
                        detector: "{instrument}-{scan_number}-{detector}",
                        scanNumber: 43}) {
```

tests/system_tests/services/startup.py
```diff
@@ -254,7 +260,44 @@ def create_clients() -> None:
         )
 
 
+def configure_numtracker():
+    def _():
```
tpoliaw: Why wrap this in a function? Isn't it only being used once?

tests/system_tests/services/startup.py
```diff
@@ -254,7 +260,44 @@ def create_clients() -> None:
...
+            data={
+                "client_id": "system-test-blueapi-admin",
+                "client_secret": "secret",
+                "grant_type": "client_credentials",
+            },
+        )
+        response.raise_for_status()
+        return response.json().get("access_token")
+
+    nt_url = os.environ.get("NT_URL", "http://localhost:8406/graphql")
```
tpoliaw: Could this go with the other constants at the top of the module?

tests/system_tests/services/startup.py
```diff
@@ -254,7 +260,44 @@ def create_clients() -> None:
...
+    response = requests.post(
+        str(nt_url),
```
tpoliaw: Isn't the url already a string?

Alexj9837 (approved):
Looks good, as some fall backs and takes whats in the just file and adds it into configure_numtracker

tpoliaw (dismissed):
Comments are mostly minor but can this PR be merged after the other convert-to-python one so that they stay separate in the history

justfile
ZohebShaikh: I'll let @shree-iyengar-dls look at it as she is already re writing the docs. But I did a quick search suggests that it is not mentioned in the docs

PR #1646, opened 2026-08-27
feat: Use websockets if stomp config not present

### LINKED ISSUE #1638: Support websocket connection in scripting interface
The `BlueapiClient` support for running plans via websockets is limited to the new `run_blocking` method. The scripting interface (eg `bc.plans.count(...)`) still uses the existing stomp connection.

While the websocket connection is still experimental, there should be some way of opting in to use it for the scripting as well. Potentially an option on the client? on the plan cache?

```python
# option on the client
bc.use_websockets = True

# option on the plan cache
bc.plans._websockets = True
```

Is there a simpler/more convenient approach?

tpoliaw:
Update the client's run_task method to use stomp if the configuration is present
and fall back to the new websocket connection if not.

Allows the scripting interface (bc.plans.count(...)) to run without
configuration using websockets.

The CLI also defaults to websockets if no configuration is present. To force
stomp monitoring, a new --stomp flag is added to be paired with the --ws flag.

The previous run_task method on BlueapiClient has been renamed to run_stomp and
run_task will delegate to the correct method depending on the configuration
available. This should not be a breaking change as there is no change in
behaviour if configuration is present.


src/blueapi/client/client.py
```diff
@@ -283,6 +285,14 @@ def from_config(cls, config: ApplicationConfig) -> Self:
         else:
             return cls(rest)
 
+    @classmethod
+    def for_host(cls, host: str | None = None) -> Self:
```
tpoliaw, line 289: Not sure what this should be called. `for_host` doesn't make sense when called as `for_host()` to use the default.

PR #1647, opened 2026-08-28
fix: prevent RUNNER detection failure

ZohebShaikh:
Append `|| true` so the justfile can still be parsed/used on systems without a container runtime like docker or podman, instead of erroring out at command substitution time.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,6 +1,13 @@
 SESSION := "cm12345-1"
 RUNNER := `command -v docker || command -v podman || true`
 
+_check-runner:
+    #!/usr/bin/env bash
+    if [[ "{{ RUNNER }}" == "" ]]; then
+        echo "{{ style('error') }}error{{ NORMAL }}: No container runtime available - either podman or docker is required"
+        exit 1
+    fi
+
 default: compose serve
 
 init-example-services:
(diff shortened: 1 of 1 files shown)


tpoliaw:
Doesn't this just make it fail later with an arguably more cryptic `compose: command not found`?

It would be good to catch this though, maybe something like

```justfile
RUNNER = `command -v docker || command -v podman || true`

_check-runner: 
    #!/usr/bin/env bash
    if [[ "{{ RUNNER }}" == "" ]]; then
        echo "No container runtime available - either podman or docker is required"
        exit 1
    fi

compose <... existing args ...>: _check-runner
    ...
```

This then fails with 
```
$ just
No container runtime available - either podman or docker is required
error: recipe `_check-runner` failed with exit code 1
```

ZohebShaikh:
I was thinking of adding a check if git or uv is not available to make it more fool proof but I thought it might be a bit excessive.

tpoliaw:
We could have a `require x` recipe that checked if a command was available and gave a useful error? Could then be used as something like

```just
serve: (_require "uv")
    ...

init-example: (_require "git")
    ...
```

ZohebShaikh:
just changed uv to uva to give a example output
```
/tmp/just-J4sNoQ/serve: line 19: uva: command not found
error: Recipe `serve` failed with exit code 127
```
Is this difficult for a software engineer to understand this ?

tpoliaw:
> Is this difficult for a software engineer to understand this?

Probably not but I've been surprised before

tpoliaw (approved):

PR #1648, opened 2026-09-01
feat: add docs for justfile

ZohebShaikh:
```
just --list
Available recipes:
    compose +ARGS="up -d --no-recreate" # Bring up the example services with docker/podman compose
    coverage                            # Run unit tests with coverage and open the HTML report
    default                             # Start the example services via compose and run the blueapi server
    init-example-services               # Clone the example-services submodule if needed but leave it alone otherwise
    lint                                # Regenerate schemas, run pre-commit checks and type checking
    repl                                # Start an interactive REPL logged in to a blueapi client
    run PLAN PARAMS                     # Run a plan against the blueapi controller
    serve *OPTS                         # Run the blueapi server against the system test config
    system *OPTS                        # Run the system test suite
    unit *OPTS="-n logical"             # Run the unit test suite
```

Shows docs. I have example a vscode [plugin](https://just.systems/man/en/visual-studio-code.html) which has made some spacing changes.

I have also added just to completion in dev-container for ease of use

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -45,8 +45,7 @@
                 "charliermarsh.ruff",
                 "ms-pyright.pyright",
                 "ms-azuretools.vscode-docker",
-                "ms-kubernetes-tools.vscode-kubernetes-tools",
-                "kokakiwi.vscode-just"
+                "ms-kubernetes-tools.vscode-kubernetes-tools"
             ]
         }
     },
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -45,7 +45,8 @@
                 "charliermarsh.ruff",
                 "ms-pyright.pyright",
                 "ms-azuretools.vscode-docker",
-                "ms-kubernetes-tools.vscode-kubernetes-tools"
+                "ms-kubernetes-tools.vscode-kubernetes-tools",
+                "kokakiwi.vscode-just"
             ]
         }
     },
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,9 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/s
     rm get_helm.sh
 RUN helm plugin install https://github.com/losisin/helm-values-schema-json.git --version 2.3.1
 
+# Enable bash completion for just commands
+RUN echo 'source <(just --completions bash)' >> /root/terminal-config/bashrc-default
+
 # The build stage installs the context into the venv
 FROM developer AS build
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,9 +15,6 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/s
     rm get_helm.sh
 RUN helm plugin install https://github.com/losisin/helm-values-schema-json.git --version 2.3.1
 
-# Enable bash completion for just commands
-RUN echo 'source <(just --completions bash)' >> /root/terminal-config/bashrc-default
-
 # The build stage installs the context into the venv
 FROM developer AS build
 
--- a/justfile
+++ b/justfile
@@ -14,7 +14,6 @@ default: compose serve
 [doc('Clone the example-services submodule if needed but leave it alone otherwise')]
 init-example-services:
     #!/usr/bin/env bash
-    # Clone the example-services submodule if needed but leave it alone otherwise
     if [[ $(git submodule status example-services) =~ ^- ]]; then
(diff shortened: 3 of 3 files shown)


shree-iyengar-dls (approved):

Dockerfile
```diff
@@ -15,6 +15,9 @@ RUN curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/s
     rm get_helm.sh
 RUN helm plugin install https://github.com/losisin/helm-values-schema-json.git --version 2.3.1
 
+# Enable bash completion for just commands
+RUN echo 'source <(just --completions bash)' >> /root/terminal-config/bashrc-default
```
tpoliaw: shouldn't this be in the ubuntu-devcontainer image?

justfile
```diff
@@ -8,42 +8,52 @@ _check-runner:
         exit 1
     fi
 
+[doc('Start the example services via compose and run the blueapi server')]
 default: compose serve
 
+[doc('Clone the example-services submodule if needed but leave it alone otherwise')]
 init-example-services:
     #!/usr/bin/env bash
     # Clone the example-services submodule if needed but leave it alone otherwise
```
tpoliaw: This isn't needed in both places

```suggestion
```

justfile
```diff
@@ -8,42 +8,52 @@ _check-runner:
...
     if [[ $(git submodule status example-services) =~ ^- ]]; then
         git submodule update --init example-services
     fi
 
+[doc('Bring up the example services with docker/podman compose')]
 compose +ARGS="up -d --no-recreate": init-example-services _check-runner
-    {{ RUNNER }} compose -f tests/system_tests/compose.yaml {{ARGS}}
+    {{ RUNNER }} compose -f tests/system_tests/compose.yaml {{ ARGS }}
 
+[doc('Run the blueapi server against the system test config')]
```
tpoliaw:
```suggestion
[doc('Run the blueapi server with the system test config')]
```

justfile
```diff
@@ -8,42 +8,52 @@ _check-runner:
...
 serve *OPTS:
     #!/usr/bin/env bash
     source tests/system_tests/.env
-    uv run blueapi -c tests/system_tests/config.yaml {{OPTS}} serve
+    uv run blueapi -c tests/system_tests/config.yaml {{ OPTS }} serve
 
+[doc('Run a plan against the blueapi controller')]
```
tpoliaw:
```suggestion
[doc('Run a plan using the blueapi CLI')]
```

justfile
```diff
@@ -8,42 +8,52 @@ _check-runner:
...
+[doc('Run the system test suite')]
 system *OPTS:
     uv run pytest tests/system_tests {{ OPTS }}
 
+[doc('Run unit tests with coverage and open the HTML report')]
 coverage:
     uv run pytest tests/unit_tests --cov --cov-report html
     xdg-open htmlcov/index.html
 
+[doc('Start an interactive REPL logged in to a blueapi client')]
```
tpoliaw:
```suggestion
[doc('Start an interactive REPL with an authenticated blueapi client')]
```

.devcontainer/devcontainer.json
```diff
@@ -45,7 +45,8 @@
                 "charliermarsh.ruff",
                 "ms-pyright.pyright",
                 "ms-azuretools.vscode-docker",
-                "ms-kubernetes-tools.vscode-kubernetes-tools"
+                "ms-kubernetes-tools.vscode-kubernetes-tools",
+                "kokakiwi.vscode-just"
```
tpoliaw: Could this go in the copier template?

tpoliaw (commented):
The docs are good, thanks for adding them, mainly insure what should be here vs in the upstream projects.

tpoliaw:
Was there a reason for preferring the `[doc(...)]` format over the plain comment docstring approach?

```just
[doc('comment here')]
recipe:
   ...
```
vs
```just
# comment here
recipe:
    ...
```

Dockerfile
ZohebShaikh: just is getting installed by default in the [image](https://github.com/DiamondLightSource/ubuntu-devcontainer/blob/3a574c959db44119db28bdd43ffd5b706cd563ae/Dockerfile#L16) will raise a PR there

ZohebShaikh:
>Was there a reason for preferring the [doc(...)] format over the plain comment docstring approach?

Yes I personally like to make it explicit it is for documentation. Can use the docstring if you prefer it.

When I was going through the docs There where a lot of, if you do it this way the docstring will come, if you do it this way it wont and doc approach was the one which had the least cognitive load in my mind

tpoliaw (approved):
No strong preference either way

PR #1650, opened 2026-09-01
refactor: update coverage command

ZohebShaikh:
changed coverage to open webrowser using uv rather than depending on Linux specific command


ZohebShaikh:
@tpoliaw If you don't like it, I will close this

tpoliaw (approved):

PR #1652, opened 2026-09-03
fix: reorder just recipes to make default default again

tpoliaw:
The first recipe is the one that is treated as default, the name is not
relevant. Renaming 'default' to 'dev-server' better describes what it is
doing but as it is intended as the default recipe, it is unlikely that
the name will ever be used.


ZohebShaikh (approved):
I was seeing this issue as well, you got here first

shree-iyengar-dls (approved):
Makes sense

PR #1654, opened 2026-09-07
fix: install scratch dependencies one by one

### LINKED ISSUE #1590: Cannot install a scratch version of a dependency that has a pin
In my values.yaml I have the following:
```yaml
    scratch:
      repositories:
        - name: dodal
          remote_url: https://github.com/DiamondLightSource/dodal.git
        - name: crystallography-bluesky
          remote_url: https://github.com/DiamondLightSource/crystallography-bluesky.git
        - name: daq-config-server
          remote_url: https://github.com/diamondLightSource/daq-config-server.git
        - name: ophyd-async
          remote_url: https://github.com/bluesky/ophyd-async.git
        - name: heliotrapi
          remote_url: https://github.com/diamondlightsource/heliotrapi.git
```

where `crystallography-bluesky` has a dependency on `dodal` like:
```
dependencies = [
    "ispyb",
    ...
    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
]
```

When trying to install this I get an error of:
```
INFO blueapi.cli.scratch Installing packages
Using Python 3.11.15 environment at: /app/.venv
   Updating https://github.com/DiamondLightSource/dodal.git (main)
    Updated https://github.com/DiamondLightSource/dodal.git (e9a073c8a2344071ab5482283134a9461f79d5c3)
  × Failed to resolve dependencies for `crystallography-bluesky`
  │ (v0.1.dev160+g5f0e93a59)
  ╰─▶ Requirements contain conflicting URLs for package `dls-dodal`:
      - file:///workspace/dodal (editable)
      - git+https://github.com/DiamondLightSource/dodal.git@main
```

I would expect the editable version to override the one specified in the `pyproject.toml`. I _can_ get round this by not specifying a specific version in the `pyproject.toml`, which I should arguably be doing anyway as I should be making it a proper release but this is annoying.

## Acceptance Criteria
- Specified scratch repos will install over ones in the dependency tree

ZohebShaikh:
closes #1590

I have gone with the suggestion given by Jacob in installing the dependencies.

It take slightly longer but if it is the expected behaviour we should support that.

### DIFF
--- a/src/blueapi/cli/scratch.py
+++ b/src/blueapi/cli/scratch.py
@@ -129,22 +129,16 @@ def scratch_install(*paths: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) -> 
     """
     if not paths:
         return
-    args = [
-        "uv",
-        "pip",
-        "install",
-    ]
+    LOGGER.info("Installing packages")
     for path in paths:
         _validate_directory(path)
-        args.extend(["-e", str(path)])
-
-    LOGGER.info("Installing packages")
-    process = Popen(args)
-    process.wait(timeout=timeout)
-    if process.returncode != 0:
-        raise RuntimeError(
-            f"Failed to install packages: Exit Code: {process.returncode}"
-        )
+        args = ["uv", "pip", "install", "-e", str(path)]
+        process = Popen(args)
+        process.wait(timeout=timeout)
+        if process.returncode != 0:
+            raise RuntimeError(
+                f"Failed to install packages: Exit Code: {process.returncode}"
+            )
 
 
 def _validate_root_directory(root_path: Path, required_gid: int | None) -> None:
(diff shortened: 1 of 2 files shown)


src/blueapi/cli/scratch.py
```diff
@@ -129,22 +129,18 @@ def scratch_install(*paths: Path, timeout: float = _DEFAULT_INSTALL_TIMEOUT) ->
...
-        args.extend(["-e", str(path)])
 
     LOGGER.info("Installing packages")
-    process = Popen(args)
-    process.wait(timeout=timeout)
-    if process.returncode != 0:
-        raise RuntimeError(
-            f"Failed to install packages: Exit Code: {process.returncode}"
-        )
+    for path in paths:
```
tpoliaw: Could this all be done in the same loop instead of validating the directories first?

tpoliaw (approved):

PR #1655, opened 2026-09-07
ci: always use latest v4 for codeql

ZohebShaikh:
In each codeql version the bundle is getting updated, and we are receiving it later because we merge it manually to the latest.
The commit has was required for un-trusted actions and codeql is a trusted action by github


Alexj9837 (approved):
lgtm

PR #1664, opened 2026-09-10
refactor: use directory name instead of pypi name

ZohebShaikh:
More than a PR this is to explore should we have 

 `dodal` as appose to [`dls-dodal`](https://github.com/DiamondLightSource/dodal/blob/3c6eab0c1079178192a23c4c6437a9d4d9287f73/pyproject.toml#L6)


tpoliaw:
Where is this name being used? In reporting the environment? If the main goal is to get rid of the tomlkit dependency, now we're on 3.10+ there is tomllib in the standard library.

PR #1665, opened 2026-09-10
refactor: make tiled client smaller

ZohebShaikh:
Dependencies reduced by 731M to 522M.

The main thing removed is llvmlite 171.2 MB


pyproject.toml
```diff
@@ -14,7 +14,10 @@ classifiers = [
 ]
 description = "Lightweight bluesky-as-a-service wrapper application. Also usable as a library."
 dependencies = [
-    "tiled[client]>=0.2.4",
+    "tiled[minimal-client]>=0.2.4",
+    "pyarrow",                                       # Needed for TiledWriter
+    "pandas",                                        # Needed for TiledWriter
+    "ndindex",                                       # Needed for TiledWriter
     "bluesky[plotting]>=1.14.0",                     # plotting includes matplotlib, required for BestEffortCallback in run plans
```
ZohebShaikh, line 21: I'm going backwards and forward on whether I should put this in tiled or [bluesky-tiled-plugin](https://github.com/bluesky/bluesky-tiled-plugins/blob/main/pyproject.toml) or just leave it on blueapi level

tpoliaw: If these dependencies are only needed by the TiledWriter, they should be added to wherever TiledWriter is defined.

tpoliaw (changes requested):
Reducing the tiled dependencies is great but not if we have to start managing transitive dependencies.

Issue #1414, opened 2026-02-25
Add optional dependency groups for server and client components


keithralphs:
Installing blueapi into a new project shouldn't include dependencies required for the server and client. It should only include API types (`TypeRequest`, `DataEvent` etc).

Adding `blueapi[client]` should bring in the `BlueapiClient` and dependencies without dragging in fastapi etc.

PR #1666, opened 2026-09-10
refactor: make client almost client again

ZohebShaikh:
I would like to go further in the split by doing the following but would like feedback on this before

1. Move setup-scratch to be server only
2. Have a click interface for only the serve which can only run commands like serve, setup-scratch
3. Move the ```bluesky, ophyd-async,event-model,opentelemetry-distro,observability-utils``` to server only

Related to but does not solves https://github.com/DiamondLightSource/blueapi/issues/1414


Alexj9837 (approved):
LGTM

src/blueapi/core/context.py
```diff
@@ -139,6 +150,8 @@ def __post_init__(self, configuration: ApplicationConfig | None):
 
         if (nt_conf := configuration.numtracker) is not None:
             if configuration.env.metadata is not None:
+                from blueapi.service.numtracker import NumtrackerClient
```
tpoliaw, line 153: ~~If we need to import it anyway, is there any real benefit to wrapping the initial import in the TYPE_CHECKING check?~~

I see the circular import error when importing the client eagerly. Were there other issues with numtracker being in the utils package instead of service? Lazy imports to avoid circular references feels like we're working around doing something wrong.

pyproject.toml
```diff
@@ -14,29 +14,25 @@ classifiers = [
...
     "ophyd-async>=0.13.5",
-    "aioca",
     "pydantic>=2.0",
     "pydantic-settings",
     "stomp-py",
     "PyYAML>=6.0.2",
-    "click>=8.2.0",
-    "fastapi>=0.112.0",
-    "uvicorn>=0.52.1",
+    "click>=8.2.0",                                  # need for blueapi serve and blueapi setup-scratch
```
tpoliaw, line 25: click is needed for the whole CLI isn't it?

```suggestion
    "click>=8.2.0",
```

PR #1668, opened 2026-09-11
ci: update tiled version to 0.2.18

ZohebShaikh:


tpoliaw (approved):
Again, not really sure what is going on but looks believable

PR #1669, opened 2026-09-11
feat: update tiled node insertion structure

ZohebShaikh:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


tests/system_tests/services/opa_config/config.yaml
```diff
@@ -5,7 +5,7 @@ services:
 bundles:
   diamond-policies:
     service: ghcr
-    resource: ghcr.io/diamondlightsource/authz-policy:0.0.24
+    resource: ghcr.io/zohebshaikh/authz-policy:0.0.25-alpha
```
ZohebShaikh, line 8: PR [here](https://github.com/DiamondLightSource/authz/pull/345)

src/blueapi/service/interface.py
```diff
@@ -205,7 +209,43 @@ def begin_task(
                 api_key=tiled_config.authentication,
                 headers=pass_through_headers,
             )
-
+        if task.task_id is not None:
```
tpoliaw, line 212: I think if we get this far and there is no task_id, nothing is going to work. It would be better to remove this check and the one on line 270 and replace them with 

```python
if not task.task_id:
    raise ValueError("No task_id provided")
```

and then we wouldn't have to worry about it for the rest of the function.

src/blueapi/utils/__init__.py
```diff
@@ -32,6 +32,8 @@
 Return = TypeVar("Return")
 
 INSTRUMENT_SESSION_RE = re.compile(r"^[a-z]{2}(?P<proposal>\d+)-(?P<visit>\d+)$")
+# Full proposal code (e.g. "cm12345" from "cm12345-1") for building tiled node paths.
+TILED_PROPOSAL_RE = re.compile(r"^(?P<proposal>[a-z]{2}\d+)-\d+$")
```
tpoliaw, line 36: Can we combine this with the other regex somehow instead of having two that do subtly different things?

Something like `^(?P<proposal>(?P<code>[a-z]{2})(?P<proposal_number>\d+))-(?P<visit>\d+)$` would give us `code`, `proposal`, `proposal_number` and `visit` so it could be used for both.

src/blueapi/service/interface.py
```diff
@@ -205,7 +209,43 @@ def begin_task(
...
+                proposal = match["proposal"]
+                # Each level's access blob is the prefix of the full
+                # (beamline, proposal, visit) one that access_blob() builds,
+                # matching the beamline/proposal/session tiers the tiled
+                # access policy expects a container to be tagged with.
+                session_blob = json.loads(access_blob(instrument_session, instrument))
+                level_access_tags = [
+                    [json.dumps({"beamline": instrument})],
+                    [json.dumps({"beamline": instrument, "proposal": proposal})],
+                    [json.dumps(session_blob)],
```
tpoliaw, line 229: I don't think having `access_blob` as its own function makes sense any more. It should either be updated to return all three of these, or the `proposal/visit/beamline` line should be inlined here. 

It's also going from dict -> json -> dict -> json where it probably shouldn't

src/blueapi/service/interface.py
```diff
@@ -205,7 +209,43 @@ def begin_task(
...
+                if not (match := TILED_PROPOSAL_RE.match(instrument_session)):
+                    raise ValueError("Invalid instrument session")
+                proposal = match["proposal"]
+                # Each level's access blob is the prefix of the full
+                # (beamline, proposal, visit) one that access_blob() builds,
+                # matching the beamline/proposal/session tiers the tiled
+                # access policy expects a container to be tagged with.
+                session_blob = json.loads(access_blob(instrument_session, instrument))
+                level_access_tags = [
+                    [json.dumps({"beamline": instrument})],
```
tpoliaw, line 227: might be a bit late at this stage but why are we using `beamline` instead of `instrument`?

src/blueapi/service/interface.py
```diff
@@ -205,7 +209,43 @@ def begin_task(
...
+                            tiled_client.create_container(
+                                key=key, access_tags=access_tags
+                            )
+                        except ClientError as e:
+                            if (
+                                e.response.status_code == status.HTTP_409_CONFLICT
+                            ):  # already exists
+                                ...
+                            else:
+                                raise
```
tpoliaw, line 247:
```suggestion
                            if response.status_code != status.HTTP_409_CONFLICT:
                                raise
                            # else: node already exists, continue
```

src/blueapi/service/interface.py
```diff
@@ -205,7 +209,43 @@ def begin_task(
...
+                # (beamline, proposal, visit) one that access_blob() builds,
+                # matching the beamline/proposal/session tiers the tiled
+                # access policy expects a container to be tagged with.
+                session_blob = json.loads(access_blob(instrument_session, instrument))
+                level_access_tags = [
+                    [json.dumps({"beamline": instrument})],
+                    [json.dumps({"beamline": instrument, "proposal": proposal})],
+                    [json.dumps(session_blob)],
+                ]
+                for key, access_tags in zip(
```
tpoliaw, line 231: Could `level_access_tags` be a dict instead of having to immediately zip the list with something else?

tests/system_tests/test_blueapi_system.py
```diff
@@ -612,7 +613,12 @@ def on_event(event: AnyEvent) -> None:
     assert stream_resource["run_start"] == start_doc["uid"]
     assert stream_resource["uri"] == f"file://localhost/tmp/adsim-{scan_id}-det.h5"
 
-    tiled_url = f"http://localhost:8407/api/v1/metadata/{start_doc['uid']}"
+    proposal = TILED_PROPOSAL_RE.match(start_doc["instrument_session"])["proposal"]  # type: ignore
```
tpoliaw, line 616: is this type ignore still needed?

tpoliaw (commented):
I don't really know how the authorization stuff for tiled works but it looks plausible.

The amount of stuff going on in the `begin_task` is starting to look a bit suspect though. Can be a later PR but is it possible offload some of it to the tiled writer or client?

PR #1673, opened 2026-09-15
fix: Remove trailing slash from numtracker auth

tpoliaw:


shree-iyengar-dls:
Switched to your `nt_auth` branch but system tests still failing :( especially a lot of the authentication ones

shree-iyengar-dls (approved):

PR #1679, opened 2026-09-18
feat(scratch): Add support for installing scratch repositories with `uv.lock`

### LINKED ISSUE #1678: Ensure reproducible beamline deployments by pinning dependencies from uv.lock
Currently, deploying bluesky module in scratch on a beamline resolves dependencies directly from pyproject.toml. Because dependency can pull in newer updates,, this can lead to unexpected environment between release tags.

To ensure production releases remain stable and fully reproducible, an option should be available to install locked dependency versions derived from uv.lock rather than resolving dependencies dynamically at deploy time.

## Acceptance Criteria
- A release deployment can be reproduced identically at any point in the future.

Relm-Arrowny:
fix #1678 

## Changes
- Added `use_uv_lock` in `ScratchRepository`
- Modified `scratch_install` to make use of `use_uv_lock` to allow `uv sync`
- Updated Helm schema and tests to reflect these changes


ZohebShaikh (changes requested):
`use_uv_lock` as designed doesn't really deliver on its own premise. Because blueapi installs scratch repos into one shared environment, no single repo's uv.lock tells you what's actually installed in blueapi — it only reflects that repo's own resolution. If repo A opts into use_uv_lock and repo B doesn't (or B has its own lockfile with different pinned versions), B's install can silently override dependencies A's lock was relying on. Per-repo opt-in doesn't prevent this, and worse, it doesn't even reliably work per-repo: if two use_uv_lock repos pin different versions of the same transitive dependency, the outcome depends on install order — it'll work by accident in some configs and break in others.

Proposed fix: move use_uv_lock up to ScratchConfig (not per-repo), so it's an explicit, all-or-nothing decision about the whole scratch environment's install strategy, plus a configurable path to the lockfile to use (defaulting to uv.lock when there's only one scratch repo). For multiple repos, there's no sound way to combine independently-generated lockfiles at install time — a consolidated lockfile covering all scratch repos together needs to be generated in CI at release time instead.

Relm-Arrowny:
You are 100% right, I am not sure there is any advanatage of doing uv.lock, I think I will just close the issue, as `target version` already covered my usecase in the issue.

